PR #22, opened 2024-01-24
Corgi errdq

JuergenSchreiber:
Adding ERR and DQ attributes (extensions) to the Image. They are now kept as optional as possible, but they could also be added as zero arrays in any case if wished. The first pipeline step a DQ might be needed is the bad pixel mask step. So please comment.


maxwellmb:
A couple thoughts/points of discussion after looking over changes to `data.py`: 
- When reading in different HDUs, I've been thinking it might be more robust to read in based on the 'EXTNAME' keyword rather than it's index in the hdulist. What do you all think? (e.g. around lines 162)
- In the `update_after_processing_step` when we have to pass in the err and dq arrays, we're not currently passing around their headers or allowing us to touch them explicitly. Can we imagine scenarios where we want to use that function to adjust not just the HISTORY keyword in the ext_hdr, but also the headers of the err and dq arrays? I suppose that's a more general question: how much do we want to use the err and dq headers at all? My first thought is that they're good places to keep track of edits to their data arrays, but maybe then we're keeping track or too many headers. what do people think? 
- I'm not overly familiar with masked arrays in numpy, and I'm a tad worried that it limits functionality, but at the moment it's fairly unobtrusive, so I'm not opposed to keeping that function for the time being.

semaphoreP (commented):
Hi @JuergenSchreiber, thanks for implementing this! I made some comments through the code I think could use some small changes. I think a bigger picture behavioral philosophy we can embrace is that all data we read in will be given Error and DQ maps. That way, there's no need to check so much for what data in our pipeline has them. If we do this, we need to write a pipeline function (later), that computes the photon noise contribution from the data to be added to other noise terms.

semaphoreP:
I'm not sure whether folks want to use masked arrays or not, but since it's an optional function I'm fine with it. As long as the `data` field is just a regular array.

JuergenSchreiber:
Thanks all for the really valuable comments. I tried to update accordingly. So please check again. If we have a file read with ERR/DQ extensions and an err/dq array in the Image parameters, the err/dq arrays now take precedence, since I think the user wants to overwrite the arrays from the file then. What do you think?

JuergenSchreiber:
Thanks @semaphoreP for reviewing and updating. That is of course fine with me.

Issue #13, opened 2024-01-08
Port L2a -> L2b data analysis code from JPL II&T


maxwellmb:
Many folks at JPL have already done a lot of work on this. Let's not double efforts.

We will wait for JPL to approve open-sourcing this before really getting to it.

Included capabilities functions are:

**L2a to L2b:** 
This will implement the following steps:
         - Convert from DN to electrons
         - Divide by EM gain
         - Subtract bias-subtracted, gain-divided master dark in electrons
         - Divide by corresponding flat field
         - Compute per-frame bad-pixel map from fixed bad pixel map (master bad
           pixel map constructed from calibration data such as dark and flat as
           well as other sources of knowledge where present) and additional
           flagged pixels at the frame level if any.

semaphoreP:
@mygouf, yes please! feel free to also recruit others to help if it is convenient.

maxwellmb:
For this one and the other one, it might make sense to split these tasks up
into different issues for the different sub-steps. I anticipate we might
have some new people at the meeting next week who might want to get
involved and having smaller chunks might help get others involved.

maxwellmb:
I wanted to break this task up more into functions to split up the work for better parallelization and give some porting tasks over to more Junior folks.  At the moment we have the following functions set up in the L2a_to_L2b.py file that could use some attention: `convert_to_electrons`, `cti_correction`, `flat_division`, `correct_bad_pixels`. @mygouf, have you started on any of these specifically? For the ones that haven't been started, I'll create separate issues. There are some missing steps (e.g. "divide by EM gain"), that will also have to be added.

maxwellmb:
This issue is now superseded by Issues #48, #49, #50, #51, #52 and #53.

Issue #12, opened 2024-01-08
Port L1 -> L2a data analysis code from JPL II&T 


maxwellmb:
Many folks at JPL have already done a lot of work on this. Let's not double efforts. 

We will wait for JPL to approve open-sourcing this before really getting to it. 

Included capabilities functions are: 

**L1_to_L2a:**
This will implement the following steps:
         - Separate image from prescan
         - Compute bias from prescan, including offset subtraction, and
           subtract from image
         - Build per-frame cosmic ray masks and remove cosmic rays
         - Correct for nonlinearity

semaphoreP:
Should crop down to 1024/1024 at the end of this process

mygouf:
Should I go ahead and get started on this?

semaphoreP:
I think @m-samland also had some interest in helping with this. @mygouf perhaps you can divide up some of the work to @m-samland so he can help you with this? You could decide to share a branch where you push code for this.

mygouf:
Ok, let me start working on it and think about how to divide and conquer. Once we have a good plan for L1 to L2a we can use a similar template for higher levels port and implementation. We can chat more in person next week.

maxwellmb:
a natural way to divide and conquer might be to break up all the functions in the ttr5_analog.py into separate issues. What do you think?

mygouf:
Yes this is what I had in mind. But wanted to wait for your review before going ahead and starting creating new issues.

mygouf:
> a natural way to divide and conquer might be to break up all the functions in the ttr5_analog.py into separate issues. What do you think?

Yes this is what I had in mind. But wanted to wait for your review before going ahead and starting creating new issues.

m-samland:
Yes, I am interested in helping out with porting these to the new pipeline to get used to the code base. It would be good to discuss with @mygouf so that we can coordinate!

maxwellmb:
Closing this issue since we have split it up into sub-issues.

PR #26, opened 2024-02-03
Port L1 to L2 data analysis code from iit 

mygouf:
Ported L1 to L2 data analysis code from iit.
Updated the example script ttr5_analog.py accordingly to match the analysis code architecture. 
Provided comments and suggestions directly into the ttr5_analog.py script. 
Added new classes for each processing level with a complete description of those data levels in the documentation.
This pull request is relevant to issues #12 and #13 but don't address them completely. More work is needed to do to fully port the code. Once this is reviewed and merged I'll create relevant issues to facilitate and coordinate the implementation.

### DIFF
--- a/corgidrp/L1_to_L2_proc.py
+++ b/corgidrp/L1_to_L2_proc.py
@@ -1,0 +1,618 @@
+# -*- coding: utf-8 -*-
+"""Process EMCCD images from L1 to L2."""
+
+import numpy as np
+import os
+from pathlib import Path
+
+from . import check
+from .read_metadata import Metadata
+from .gsw_nonlin import get_relgains
+from .gsw_emccd_frame import EMCCDFrame
+
+here = Path(os.path.dirname(os.path.abspath(__file__)))
+meta_path_default = Path(here, 'metadata.yaml')
+
+class L1_to_L2_proc(object):
+    """Process EMCCD frames from L1 to L2.
+
+    L1 to L2a
+    cleaned, single frame, with applicable calibrations applied
+    
+    --
+    Use cases:
+        - TTR5 Analog Non-PSF-subtracted on-sky observation data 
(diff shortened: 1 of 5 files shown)


mygouf:
> Hi Marie, I think we shouldn't merge this code into `main` until we spend more time porting the code over. In particular, we aren't using the same class architecture as the II&T pipeline.
> 
> Instead, I suggest we port functionality one at a time over, rather than copying all the code and modifying it later.

Sure, I wasn't sure how you wanted to proceed. I'm good with waiting for a more complete implementation before merging.

mygouf:
> I think the changes to ttr5_analog.py should wait as well as. This was meant to be more of a guiding script for new people doing corgidrp development to give everyone an idea about general architecture philosophy (function-based rather than too class-heavy) rather than actually be how things would be run.

I had initially started a separate script with those new edits but I was not comfortable about having separate versions of it once we move forward with the porting of L1 to L2 data analysis.

> The "Options" you suggested are excellent and I think are further ahead than this example was meant to be. How exactly we package things to run L1-L2a together is still something to be discussed, but we will definitely do that at some point both for convenience and for testing.

I'd be in favor to discuss this early on as this will facilitate future development and testing. 

> When we break out functions into issues, we should make sure to make note on the github issues on the very useful pointers you've made in tt5_analog.py to similar functions in the II&T pipeline!

Yes, this is planned! I'll go ahead and create the issues then as soon as I have some time.

semaphoreP:
I just sent an email to have a meeting to discuss how the broader data reduction framework will look like. Regardless, I believe compartmentalizing the steps into the functional framework outlined for corgidrp should work. The question is just want code will live on top of it that strings together the pipeline steps.

maxwellmb:
Closing this old pull request back from the early days.

PR #35, opened 2024-02-08
Expanding the README FAQ to include high-level pipeline design guidance

ell-bogat:
This PR is to help define the overarching pipeline design, so things like where to store metadata, global pipeline variables, etc. can be consistent. If you have ideas please discuss in the PR comments!

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,9 +112,16 @@ To run all the tests in the test suite, go to the base corgidrp folder in a term
 ### Create a pull request to merge your changes
 Use the Github pull request feature to request that your changes get merged into the `main` branch. Assign Jason/Max to be your reviewers. Your changes will be reviewed, and possibly some edits will be requested. You can simply make additional pushes to your branch to update the pull request with those changes (you don't need to delete the PR and make a new one). When the branch is satisfactory, we will pull your changes in. 
 
-### Overarching Design Guidance (Working Draft)
+### Overarching Design Guidance and FAQ
 
-* Hard-coding variables
+* What python version should I develop in?
+  * Python 3.11
+    
+* How should I treat hard-coded variables
   * If a variable value is extremely unlikely to change AND is only required by one module, it can be hard coded inside that module.
-  * If it is unlikely to change but will need to be referenced by multiple modules, it should be added to the central metadata file `corgidrp/metadata.yaml`. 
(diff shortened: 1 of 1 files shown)


ell-bogat:
Some initial ideas from the first DRP breakout:
  * Q: What python version should I develop in?
    * Automatic testing is currently implemented using Python 3.10, so that is the best way to ensure your tests pass both remotely and on your local machine. However versions between 3.9 and 3.11 should work well.
  * Q: Can I hard-code variables?
    * If a variable value is extremely unlikely to change AND is only required by one module, it can be hard coded inside that module.
    * If it is unlikely to change but will need to be referenced by multiple modules, it should be added to the central metadata file `corgidrp/metadata.yaml`. (filename is TBD)
    * If it is likely to change, it should be implemented by a user-defined argument.
  * Q: Where should I store computed variables so they can be referenced later in the pipeline?
    * If possible, in the header of the dataset being processed.
    * What types of data would make more sense to be included as a class attribute rather than a header keyword?

README.md
```diff
@@ -111,3 +111,10 @@ To run all the tests in the test suite, go to the base corgidrp folder in a term
 
 ### Create a pull request to merge your changes
 Use the Github pull request feature to request that your changes get merged into the `main` branch. Assign Jason/Max to be your reviewers. Your changes will be reviewed, and possibly some edits will be requested. You can simply make additional pushes to your branch to update the pull request with those changes (you don't need to delete the PR and make a new one). When the branch is satisfactory, we will pull your changes in. 
+
+### Overarching Design Guidance (Working Draft)
+
+* Hard-coding variables
+  * If a variable value is extremely unlikely to change AND is only required by one module, it can be hard coded inside that module.
+  * If it is unlikely to change but will need to be referenced by multiple modules, it should be added to the central metadata file `corgidrp/metadata.yaml`.
```
maxwellmb, line 119: Matthias and I were talking about potentially renaming the config file to be slightly more descriptive (although we didn't discuss the new name. Perhaps we can leave the name out for now?

dsavransky:
> My guess is that we'll probably want to freeze the python version number for testing purposes.

In that case, I advocate for 3.11 or 3.12.  3.10 will reach EOL around the time of launch.  3.11 will continue to be supported through 2027 and 3.12 through 2028.

maxwellmb (approved):
Added these comments. More to come.

PR #40, opened 2024-02-10
File structure reformatting and stubs

maxwellmb:
I set up a new file structure where pipeline steps will be functions in lX_to_lY.py files (heavily inspired by @mygouf's suggestions!). Hopefully this will help us stay more organized and limit the number of functions per module. Within those functions I created some very basic stubs for the current list of functions in the ttr5_analog example. These should be expanded, but we now have a start.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,11 @@ You will create a "feature branch" so you can develop your feature without impac
 
 ### Write your pipeline step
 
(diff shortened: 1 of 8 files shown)


maxwellmb:
Check failed. Looking into it.

semaphoreP:
Looks good! We should define a play for calibration function stubs as well in the future.

Issue #34, opened 2024-02-08
Port L1 -> L2a JPL II&T code Part 3: Correct for nonlinearity


maxwellmb:
This is an off-shoot of issue #12, L1_to_L2a.

I'm breaking it up since we this to be a single function.

This will implement the following steps:
- Correct for nonlinearity

maxwellmb:
I started a branch today [non-linearity-correction](https://github.com/roman-corgi/corgidrp/tree/non-linearity-correction). It's got the code to correct for the non-linearity (and a test for it), but not the code to generate the calibration file. That's next! 

One difference from the II&T pipeline here is a conversion of the non-linearity calibration file to a fits file to allow better tracking and keep calibration files in a similar format.

maxwellmb:
Closing this with the recent merge. Commit [4c73a83](https://github.com/roman-corgi/corgidrp/commit/4c73a839a311a0bed66745efb7cc9398659cb498)

PR #43, opened 2024-02-15
Implemented Non linearity correction port from II&T

maxwellmb:
I implemented a port over of the II&T non-linearity correction. This involved creating a new calibration data type, a "step function" in `L1_to_L2.py`, and a utility function in detector.py (based on an II&T function of the same name). 

I used a sample of the expected calibration data format in the II&T pipeline, using the file "nonlim_sample.csv". For corgidrp we want the calibration files to be fits files, so I converted that csv to a fits file. I implemented a test where I used that calibration file to generate a dataset (of 2 files) that ramps from 0 to 65536 along the x-direction. Then demonstrate that the pipeline function properly corrects for the non-linearity. Within the test file I've included a comparison with the II&T functions that correct for non-linearity and show that they result in the same correction. 

If acceptable, this may or may not fully complete Issue #34. It does not include the code to actually generate the calibration. Thoughts on whether we should make that a separate issue, or keep Issue #34 open until we complete that would be appreciated.


corgidrp/mocks.py
```diff
@@ -28,6 +30,45 @@ def create_dark_calib_files(filedir=None, numfiles=10):
...
+    Returns:
+        corgidrp.data.Dataset:
+            The simulated dataset
+    """
+    filepattern = "simcal_nonlin_{0:04d}.fits"
+    frames = []
+    for i in range(numfiles):
+        prihdr, exthdr = create_default_headers()
+        #Add the EMGAIN to the headers
+        exthdr['EMGAIN'] = em_gain
```
semaphoreP, line 51: What/who is responsible for this keyword to be populated?

corgidrp/mocks.py
```diff
@@ -88,7 +129,7 @@ def create_default_headers():
     exthdr['DPAM_H'] = 1.0
     exthdr['DPAM_V'] = 1.0
     exthdr['DATETIME'] = '2024-01-01T11:00:00.000Z'
-    exthdr['DATA_LEVEL'] = "L1"
+    exthdr['DATA_LVL'] = "L1"
```
semaphoreP, line 132: The current data specification is using DATA_LEVEL, unless you know this has changed.

tests/test_non_linearity_correction.py
```diff
@@ -0,0 +1,154 @@
...
+    # Check that curves (data in columns) contain or straddle 1.0
+    if (np.min(relgains, axis=0) > 1).any() or \
+       (np.max(relgains, axis=0) < 1).any():
+        raise NonlinException('Gain curves (array columns) must contain or '
+                              'straddle a relative gain of 1.0')
+
+    return gain_ax, count_ax, relgains
+
+
+def get_relgains(frame, em_gain, nonlin_path):
```
semaphoreP, line 60: It would be useful to include in the docstring that this is an exact copy of the II&T pipeline code.

tests/test_non_linearity_correction.py
```diff
@@ -0,0 +1,154 @@
...
+
+    #Relative correction
+    relative_correction = (collapsed_data-expected_ramp)/collapsed_data
+
+    #We are happy if the relative correction is less than 1% [TBC]
+    assert np.all(relative_correction < 1e-2)
+
+    
+    #Let's test that this returns the same thing as the II&T pipeline
+    linear_data_iit = nonlinear_dataset.all_data*get_relgains(nonlinear_dataset.all_data,emgain,os.path.dirname(os.path.abspath(__file__))+"/test_data/nonlin_sample.csv")
```
semaphoreP, line 147: best practice to use os.path.join() rather than adding strings for filenames: e.g., `os.path.join(os.path.dirname(os.path.abspath(__file__)), "test_data", "nonlin_sample.csv")`

tests/nonlin_sample.fits
semaphoreP, line 1: can this be saved in the same directory as the corresponding csv?

corgidrp/data.py
```diff
@@ -354,9 +353,82 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+        ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw 2D data is passed in)
+        input_dataset (corgidrp.data.Dataset): the Image files combined together to make this dark file (required only if raw 2D data is passed in)
+    """
+    def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=None):
+        # run the image class contructor
+        super().__init__(data_or_filepath, pri_hdr=pri_hdr, ext_hdr=ext_hdr)
+        # additional bookkeeping for Dark
+
+        # if this is a new dark, we need to bookkeep it in the header
+        # b/c of logic in the super.__init__, we just need to check this to see if it is a new dark
```
semaphoreP, line 404: comments, in the above 3 lines can probably change from the Dark version

corgidrp/data.py
```diff
@@ -354,9 +353,82 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
         if 'DATATYPE' not in self.ext_hdr or self.ext_hdr['DATATYPE'] != 'Dark':
             raise ValueError("File that was loaded was not a Dark file.")
 
+class NonLinearityCalibration(Image):
```
semaphoreP, line 356: Can we call this NonLinearityCorrection() instead?

corgidrp/data.py
```diff
@@ -354,9 +353,82 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+    """
+    Class for non-linearity calibration files. Although it's not stricly an image that you might look at, it is a 2D array of data
+
+    The required format for calibration data is as follows:
+     - CSV
```
semaphoreP, line 361: The required input is a CSV file?

corgidrp/data.py
```diff
@@ -354,9 +353,82 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+    if nonlin_raw.ndim < 2 or nonlin_raw.shape[0] < 2 or \
+       nonlin_raw.shape[1] < 2:
+        raise NonlinException('Nonlin array must be at least 2x2 (room for x '
+                              'and y axes and one data point)')
+    if not np.isnan(nonlin_raw[0, 0]):
+        raise NonlinException('First value of csv (upper left) must be set to '
+                              '"nan"')
+
+     Args:
+        data_or_filepath (str or np.array): either the filepath to the FITS file to read in OR the 2D image data
```
semaphoreP, line 393: It may be useful to clarify what data needs to be passed in here, given it's not just a regular image. I presume it is a np.array of that 2-D array specified above. How does the CSV come into play?

corgidrp/data.py
```diff
@@ -354,9 +353,82 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+        [2000, 0.950, 1.000, 1.010, 1.050],
+        [3000, 1.000, 1.001, 1.011, 1.060],
+    ],
+
+    where the row headers [1, 1000, 2000, 3000] are dn counts, the column
+    headers [1, 10, 100, 1000] are EM gains, and the first data column
+    [0.900, 0.910, 0.950, 1.000] is the first of the four relative gain curves.
+
+    Some checks that could be done on this data (ported from the IIT code): 
+    # File format checks
```
semaphoreP, line 383: is there a reason we don't just put this in the code below?

semaphoreP (commented):
Looks good overall. Some small comments overall to try to clean things up.

semaphoreP:
I think it makes sense to separate "Correcting for non-linearity" from computing the calibration file as separate issues.

corgidrp/mocks.py
maxwellmb: EM Gain is an interesting one with a few different cases: 
1) If the commanded emgain is 1 we probably don't need to do anything. I'm looking in to the header keyword we might expect. There's one that is CMDGAIN that is the commanded gain, but I don't know if that's the K-gain or the emgain at the moment (I've asked Alex). 
2) If the EM Gain is v high we can measure it from the pre-scan region. 
3) If the EM gain is lower then we may not be able to do that, and I believe that it's an outstanding task to figure out how to do this. 

I suggest we have a step function that we call no matter what. If CMDGAIN =1 (assuming this is the correct keyword) then it just sets EMGAIN=1, otherwise it does a calculation and sets the keyword. In all cases its that same function that is responsible for setting that keyword. Calculating EMGAIN is a required step that doesn't have a stub yet, only because I based the stubs off the ttr5_analog.py script which wasn't based on the EMGAIN workflow. It should be done. 

Thoughts?

corgidrp/mocks.py
maxwellmb: Right, I only changed this because having so many letters threw a warning that it makes it a HIERARCH-type keyword, which seemed unnecessary. Alex just told me though that this was requested specifically from the PS team, so I don't think we should change it for now. Ultimately I think the functionality and use will be the same as a normal keyword, but for visually inspecting headers it's a slight pain. We could consider changing it later. @mygouf are you aware of a reason that it was chosen to be this way?

corgidrp/mocks.py
maxwellmb: Ok, just heard from Alex, this is on the list of nebulous stuff that we should be able to get into the primary header. I've added a requested Header keyword to Data Formats page on Confluence, CMDEMGN, for ComManDed EM GaiN.

corgidrp/data.py
maxwellmb: We could, but why? I personally think of _correcting_ as an action and a _calibration_ file/data as the information (it's true there's also the action of performing a calibration).

corgidrp/mocks.py
semaphoreP: It looks like `CMDGAIN` is already a planned keyword in the extension header?

Ok understood. So, we'll need an additional pipeline step before this on that calculates the EMGAIN then.

corgidrp/data.py
semaphoreP: Fair. We can leave it as is.

corgidrp/mocks.py
maxwellmb: CMDGAIN is not the EMGAIN (confirmed by Alex)

But yes, we'll need that step.

tests/test_non_linearity_correction.py
maxwellmb: Done

tests/test_non_linearity_correction.py
maxwellmb: Done

tests/nonlin_sample.fits
maxwellmb: Done

corgidrp/data.py
maxwellmb: Done

corgidrp/data.py
maxwellmb: CSV was copied and pasted from previous docstring. Deleted now with more data format descriptions.

corgidrp/data.py
maxwellmb: Added more description

corgidrp/data.py
maxwellmb: I stopped being lazy and put it in.

maxwellmb:
@semaphoreP, I think it's good for you to look at again.

corgidrp/data.py
semaphoreP: one more "dark" in line 409 :P

semaphoreP:
Looks good! I only found 1 "typo". I also created an issue about the EMGAIN step. Otherwise, looks good to me!

semaphoreP (approved):

PR #47, opened 2024-02-20
Implemented Pre-scan processing and bias subtraction port from II&T

ell-bogat:
We ported over the cropping, prescan processing, and bias correction steps from the II&T code. We defined the function `prescan_biassub()` in `l1_to_l2a.py`, which does the following:
* crops the frame down to the science area (a switch exists to return the full frame if desired), using a utility function `slice_section()` in `detector.py` and a dictionary `detector_areas` which defines the different detector areas for both the science and engineering observation types. The `detector_areas` dictionary can be moved to a more central location for pipeline constants if preferred.
* measures the bias on a row-by-row basis, and subtracts a bias offset which is determined via a `float` input parameter. (Is this something we will get from the telescope data?)
* Subtracts the offset-corrected bias from the image.

We also added a unit test which ensures that the output dataset is the correct shape, and that the links between the `dataset.all_data` attribute and individual frame data are maintained (as well as those for the error and dq arrays), but we haven't done a comparison with the output from the II&T code to make sure that the behavior is precisely the same. Ideas on the best way to do this (and exactly what aspects of the behavior are critical to keep constant) are much appreciated!


semaphoreP:
A couple of suggestions to help with PEP8 issues:

1. You need a new line before `Returns` in the docstring, to get rid of those issues
2. For `obstype` you have a extra space after specifying the type of data and before the `:`. 

For the last syntax error issue: Instead of asserts, I would recommend using an if statement + throw exception on error. The syntax checking also is saying that `{i}` is undefined in that assert statement print-out. And you need to define i for it to now create an error.

This isn't a full code review, but just to help fix the errors from the automatic tests.

tests/test_prescan_sub.py
```diff
@@ -0,0 +1,78 @@
...
+    },
+    'ENG' : {
+        True: (2200,2200),
+        False : (1024,1024)
+    }
+}
+
+def test_prescan_sub():
+    """
+    Generate mock input data and pass into prescan processing function
```
semaphoreP, line 22: It would be useful to make the docstring more informative. What does it test specifically? E.g., generate both ENG & SCI data, makes sure the output shapes are the size we expect, and cropping does not mess up the pointers in the Image class.

semaphoreP (approved):
Looks pretty good! I don't have much to say with regards to adding the II&T tests (perhaps useful to get @mygouf's thoughts on that). But the code here looks good with only a few minor comments.

maxwellmb:
In terms of testing: I think the primary things that this function does is to crop the image and do the pre-scan bias subtraction. If possible, in order to compare it to II&T, can you find the code/function(s) that does it there and copy+paste it and give it similar inputs and make sure that the output is the same (similar strategy as what I did)?

maxwellmb:
@mygouf, @kjl0025 could one of you please assign yourselves to this pull request and take a look? Thanks.

corgidrp/mocks.py
```diff
@@ -30,10 +31,52 @@ def create_dark_calib_files(filedir=None, numfiles=10):
...
+        sim_data = np.random.poisson(lam=150, size=size)
+        frame = data.Image(sim_data, pri_hdr=prihdr, ext_hdr=exthdr)
+        if filedir is not None:
+            frame.save(filedir=filedir, filename=filepattern.format(i))
+        frames.append(frame)
+    dataset = data.Dataset(frames)
+    return dataset
+
+
+def create_default_headers(obstype="ENG"):
```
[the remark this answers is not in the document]
ell-bogat: Done!

corgidrp/mocks.py
```diff
@@ -71,10 +72,52 @@ def create_nonlinear_dataset(filedir=None, numfiles=2,em_gain=2000):
...
+    """
+    
+    if obstype == "SCI":
+        size = (1200, 2200)
+    elif obstype == "ENG":
+        size = (2200, 2200)
+    elif obstype == "CAL":
+        size = (2200,2200)
+    else:
+        raise ValueError(f'Obstype {obstype} not in ["SCI","ENG","CAL"]')
```
kjl0025, line 96: "CAL" should be taken out.

kjl0025: Oh, maybe the "CAL" case will be added later?

semaphoreP: I think we need to coordinate with @agreenbaum on exactly what all the different OBSTYPE categories will be for the real data.

semaphoreP: But maybe not something to figure out right now for this PR

kjl0025:
Just a few general comments, not specific to the changed code on this branch.  And I will probably make some more thorough comments in Confluence for testing standards:

1. `exthdr['DATA_LEVEL']` in `create_default_headers()` is called b/c 'DATA_LEVEL' is more than 8 characters, so a HIERARCH is instead made.  We could suppress that warning with the `warnings` module if we really want that name, or we could shorten to something like 'DATA_LVL'.  @ell-bogat Perhaps you could make that change in this branch?
2.  For the modules I delivered to Roman, we had to test everything, including inputs.  Every input to a function had to checked in the function, and that check was usually tested in a unit test to make sure the check worked.  For example, if an input was supposed to be an integer, the function would raise a `TypeError` if that input was not an integer, and there was a unit test that made sure this non-integer case was caught and that a `TypeError` was raised.  However, if we agree that we are okay with skipping input checks that seem trivial (using our own judgment there), then I suppose I'm okay with that. 
3.  I think it would be good to have a `test_filename.py` corresponding to each `fliename.py`.  That way, we can keep track of the tests that belong to each file.  
4.  I also think it's better to make 1 test function per thing tested (or per small group of closely related things tested) so that when `pytest` is run, we can easily see what all fails at once rather than just seeing the first `assert` line that didn't make it within a test function while not knowing that some `assert` lines further along in the function also fail.

maxwellmb:
Hi @kjl0025 , 
Thanks for the thoughts! Some quick responses: 
1. I think we should leave this for now. This is one of the default keywords coming from the SSC and @agreenbaum mentioned to be that the PS team specifically asked for this keyword for L2 scripts. I would like to eventually change, but I don't know if we understand the full implications (e.g. if the GITL software needs it). 
2. For the moment we're staying away from this philosophy. Jason can elaborate further. 
3. At the moment we're testing things on a "step_function" level with each test file. This would be a significant change, but would be interesting to discuss the pros and cons. I don't think this should hold up this pull request and is maybe best discussed at our meeting next week. 
4. I think we've been doing this inconsistently, but seems like we could implement this with either the current testing framework or a `test_filename.py` framework. @semaphoreP, I may add some comments about this to the testing instructions in the README.

tests/test_prescan_sub.py
```diff
@@ -0,0 +1,348 @@
...
+            # axes[1].set_title('II&T result')
+
+            # im2 = axes[2].imshow(corgidrp_result-iit_result,cmap='seismic')
+            # plt.colorbar(im2)
+            # axes[2].set_title('Difference')
+
+            # plt.tight_layout()
+            # plt.show()
+
+
```
ell-bogat, line 345: Note: tolerance for this test in II&T code was set to 1, but I'm not sure how it ever passed given the sigma on the normal distribution was also set to 1. tolerance of 6 seems to pass more regularly

ell-bogat: Update: they were returning the bias itself as measured by the bias subtraction function, whereas I was only returning the bias-subtracted data. Since each pipeline step should only return a dataset object, I added a header keyword `'MED_BIAS'` which is the median bias measured (including the bias offset) in the step, so this can be referenced in the future. The test works now with the original (~1%) tolerance!

tests/test_prescan_sub.py
```diff
@@ -342,7 +340,146 @@ def test_prescan_sub():
...
+        r0c0 = detector_areas[obstype]['prescan']['r0c0']
+        rows = detector_areas[obstype]['prescan']['rows']
+        cols = detector_areas[obstype]['prescan']['cols']
+        dataset_10.all_data[:,r0c0[0]:r0c0[0]+rows, r0c0[1]:r0c0[1]+cols] += bias_offset
+    
+        for return_full_frame in [True,False]:
+            output_dataset_0 = prescan_biassub(dataset_0, return_full_frame=return_full_frame,bias_offset=0)
+            output_dataset_10 = prescan_biassub(dataset_10, return_full_frame=return_full_frame,bias_offset=bias_offset)
+
+            if not np.nanmax(output_dataset_0.all_data-output_dataset_10.all_data) < tol:
```
kjl0025, line 475: Change to `np.nanmax(np.abs(`

tests/test_prescan_sub.py
```diff
@@ -342,7 +340,146 @@ def test_prescan_sub():
...
+        # Overwrite data with normal distribution
+        rng = np.random.default_rng(seed)
+        dataset.all_data[:,:,:] = rng.normal(bval, 1,
+                                             size=dataset.all_data.shape)
+
+        for return_full_frame in [True, False]:
+            
+            output_dataset = prescan_biassub(dataset, return_full_frame=return_full_frame)
+
+            if output_dataset[0].ext_hdr['MED_BIAS'] - bval > tol:
```
kjl0025, line 389: Change to np.abs() > tol. It works without this b/c the quantity turns out to be positive, but this change would make it clearer to someone seeing this test fresh that the subtraction is not negative, thus avoiding the tolerance check.

tests/test_prescan_sub.py
```diff
@@ -321,10 +318,11 @@ def test_prescan_sub():
             corgidrp_result = output_dataset[0].data
             iit_result = iit_frames[0] if return_full_frame else iit_images[0]
 
-            if not corgidrp_result == approx(iit_result):
+            if not np.nanmax(corgidrp_result-iit_result) < tol:
```
kjl0025, line 321: Change to np.abs() > tol. It works without this b/c the quantity turns out to be positive, but this change would make it clearer to someone seeing this test fresh that the subtraction is not negative, thus avoiding the tolerance check.

corgidrp/mocks.py
```diff
@@ -84,18 +139,28 @@ def create_default_headers():
...
+    exthdr['NAXIS'] = 2
+    exthdr['NAXIS1'] = NAXIS1
+    exthdr['NAXIS2'] = NAXIS2
     exthdr['PCOUNT'] = 0
     exthdr['GCOUNT'] = 1
     exthdr['BSCALE'] = 1
     exthdr['BZERO'] = 32768
-    exthdr['ARRTYPE'] = 'SCI'
+    exthdr['ARRTYPE'] = obstype # seems to be the same as OBSTYPE
     exthdr['SCTSRT'] = '2024-01-01T12:00:00.000Z'
```
kjl0025, line 164: I'm not sure what some of these headers are, but I assume they were all discussed and worked out already.  Otherwise, I approve of this pull request.  Given my ignorance of the headers here, I will let @semaphoreP or @maxwellmb pull the trigger to pull this in.  I made issues about the peripheral unresolved things (HIERARCH, obstypes, and simulated data that sometimes causes failed tests) that can be dealt with in another PR.

kjl0025 (commented):
I'm not sure what some of these headers are, but I assume they were all discussed and worked out already. Otherwise, I approve of this pull request. Given my ignorance of the headers here, I will let @semaphoreP or @maxwellmb pull the trigger to pull this in. I made issues about the peripheral unresolved things (HIERARCH, obstypes, and simulated data that sometimes causes failed tests) that can be dealt with in another PR.

corgidrp/mocks.py
semaphoreP: Thanks for putting those issues in. These headers are all specified in our team confluence (see "Data Formats" page). I prescribed dummy values to each of them, so we should change the values when we need them to be more reasonable.

PR #59, opened 2024-03-08
Add error term from prescan bias subtraction step

ell-bogat:
I modified the prescan_biassub() step in l1_to_l2a.py to measure and track the error associated with that step.

Since the bias is measured as the median value of the "reliable" prescan region for each row, I calculated the error using the standard error on the mean, so the error for each row is the standard deviation of the prescan values divided by the square root of the number of columns in the reliable prescan region.

The error measurement is a 2D array for each Image in the Dataset, which we keep track of in a list and then apply to each Image in the Dataset using the Image.add_error_term() class method at the end. (It's a little clunky, but since the data array sizes change between L1 and L2a, I think we have to do it this way)

I've also implemented some checks in the unit tests specific to the error terms:
* Processing an image of all zeros should return an image of all zeros and an error measurement of all zeros.
* Processing a uniform image should return an image of all zeros and an error measurement of all zeros.
* Processing an image with a gaussian distribution (mean m and standard deviation sigma) should return a bias measurement of approximately m, and an error measurement of approximately sigma / root(200), since we use 200 columns of the prescan region to do the bias subtraction.

Right now the error measurements are coming out a little higher than I expect (~20%), so I'm curious to hear if y'all think this is an acceptable deviation:

If I start with an image composed of purely gaussian noise with mean 100 and stdev 1:
image

We get a good subtraction!
image

But the error measurements are a bit inflated compared to what I expected:
image

Please let me know what you think!

### DIFF
--- a/corgidrp/l1_to_l2a.py
+++ b/corgidrp/l1_to_l2a.py
@@ -6,8 +6,7 @@ import numpy as np
 def prescan_biassub(input_dataset, bias_offset=0., return_full_frame=False):
     """
     Measure and subtract the median bias in each row of the pre-scan detector region. 
-    The error associated with this step is taken from the standard error on the median 
-    for each row. This step also crops the images to just the science area, or 
+    This step also crops the images to just the science area, or 
     optionally returns the full detector frames.
 
 
@@ -76,7 +75,7 @@ def prescan_biassub(input_dataset, bias_offset=0., return_full_frame=False):
 
             al_prescan = prescan
 
-        # Measure bias and error (standard error of the mean for each row, add this to 3D image array)
+        # Measure bias and error (standard error of the median for each row, add this to 3D image array)
         medbyrow = np.median(al_prescan, axis=1)[:, np.newaxis]
         sterrbyrow = np.std(al_prescan, axis=1)[:, np.newaxis] * np.ones_like(image_data) / np.sqrt(al_prescan.shape[1])
         new_err_list.append(sterrbyrow)   
--- a/corgidrp/l1_to_l2a.py
+++ b/corgidrp/l1_to_l2a.py
@@ -77,7 +77,7 @@ def prescan_biassub(input_dataset, bias_offset=0., return_full_frame=False):
 
         # Measure bias and error (standard error of the median for each row, add this to 3D image array)
         medbyrow = np.median(al_prescan, axis=1)[:, np.newaxis]
-        sterrbyrow = np.std(al_prescan, axis=1)[:, np.newaxis] * np.ones_like(image_data) / np.sqrt(al_prescan.shape[1])
+        sterrbyrow = np.std(al_prescan, axis=1)[:, np.newaxis] * np.ones_like(image_data) / np.sqrt(al_prescan.shape[1]) * np.sqrt(np.pi / 2.)
         new_err_list.append(sterrbyrow)   
             
 
--- a/corgidrp/l1_to_l2a.py
+++ b/corgidrp/l1_to_l2a.py
@@ -42,11 +42,6 @@ def prescan_biassub(input_dataset, bias_offset=0., return_full_frame=False):
             # Get the image area
             image_data = slice_section(frame_data, obstype, 'image')
             image_dq = slice_section(frame_dq, obstype, 'image')
-            # error maps are 3-D
-            image_err = []
-            for err_slice in frame_err:
-                image_err.append(slice_section(err_slice, obstype, 'image'))
-            image_err = np.array(image_err)
             
             # Get the part of the prescan that lines up with the image, and do a
             # row-by-row bias subtraction on it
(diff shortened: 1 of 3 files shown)


maxwellmb:
To be honest, this isn't throwing any major red flags for me.

maxwellmb:
I'm not sure what an hvon or hvoff distribution is. Do you mind expanding in the docstrings of those tests?

semaphoreP:
The error on the median is larger than the error on the mean, and I believe we are using the median for the prescan subtraction. So, it's not surprising that the measured error is larger than the expected error. In fact, I have some old notes that say the error on the median is sqrt(pi/2) ~ 1.25x larger than the error on the mean for gaussian noise, which is in line with your observations.

ell-bogat:
Thanks Jason, looks like that test can stay as-is with the slightly inflated tolerance so the checks still pass.

ell-bogat:
Max, I'm not sure what hvon and hvoff stand for exactly, as the test names were ported over from the II&T code. But practically the hvoff distribution is a simple gaussian distribution, and the hvon distribution is a gaussian, plus additional high values in the "unreliable" region of the prescan area, plus an exponential distribution, minus the mean of the exponential distribution, which represents some kind of direct current contribution from whatever "hv" is. 

Perhaps @mygouf knows more? There wasn't much more info I could find about it within the II&T code, but I've updated the docstrings with what I know so far.

maxwellmb:
Sounds good. Let's wait to hear from Marie, but otherwise this is looking good. I have those two very small comments on the code, but this works for me.

corgidrp/l1_to_l2a.py
```diff
@@ -108,8 +106,10 @@ def prescan_biassub(input_dataset, bias_offset=0., return_full_frame=False):
         frame.err = out_frames_err_arr[i]
         frame.dq = out_frames_dq_arr[i]
 
-            
-    history_msg = "Frames cropped and bias subtracted"
+        # Add new error component from this step to each frame using the Image class method
+        frame.add_error_term(new_err_list[i],"bias_sub")
```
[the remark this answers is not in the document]
ell-bogat: Sure! I didn't know if that would throw a warning for having a header keyword longer than 8 characters, hence the shorter name.

tests/test_prescan_sub.py
```diff
@@ -359,28 +361,33 @@ def test_bias_zeros_frame():
...
     bval = 100.
+    sig = 1.
     seed = 12346
 
     ###### create simulated data
     datadir = os.path.join(os.path.dirname(__file__), "simdata")
     
     for obstype in ['SCI', 'ENG']:
-        # create simulated data
+        # create simulated da.
```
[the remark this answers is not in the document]
ell-bogat: Yep, thank you!

corgidrp/l1_to_l2a.py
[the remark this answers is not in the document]
semaphoreP: You can set a value to longer than 8 characters without a warning (unlike setting keywords to a character longer than 8). (I believe)

kjl0025:
"hv" is the voltage applied in the EM gain register.  So "hvoff" means no gain applied (just a Gaussian since the noise there is due mainly to read noise), and "hvon" means the voltage is on, and the exponential distribution represents the effect of gain (exponential distribution approximately correct for few particles going through register, close enough for a simple test), and then the mean of the exponential is subtracted so that there is no pedestal or shift in the mean from this exponential distribution addition (no DC component), and then a bias is added on top.

semaphoreP:
@ell-bogat can you add some comments to document the meaning of `hv`/`hvoff`, and the fact that the error tolerance is set my the median of Gaussian noise, and not the mean? Otherwise, I think it looks good!

ell-bogat:
Docstrings and comments are updated! I also did a fresh pull from the main branch, so merging should be clean :)

corgidrp/l1_to_l2a.py
```diff
@@ -42,27 +49,37 @@ def prescan_biassub(input_dataset, bias_offset=0., return_full_frame=False):
...
+            image_err = []
+            for err_slice in frame_err: 
+                image_err.append(err_slice)
+            image_err = np.array(image_err)
+
+            al_prescan = prescan
+
+        # Measure bias and error (standard error of the median for each row, add this to 3D image array)
+        medbyrow = np.median(al_prescan, axis=1)[:, np.newaxis]
+        sterrbyrow = np.std(al_prescan, axis=1)[:, np.newaxis] * np.ones_like(image_data) / np.sqrt(al_prescan.shape[1])
```
semaphoreP, line 80: Should this have the sqrt(pi/2) as I mentioned earlier?

semaphoreP (commented):
Just one last thing I noticed about what we are calling the error. Is this intentional?

corgidrp/l1_to_l2a.py
ell-bogat: Ah, I think I misinterpreted your earlier comment, and was accidentally comparing the error - measured using the formula for std error on the mean - to the expected error for the median. It looks like the errors for the standard error on the median are still a bit inflated compared to the expectation. Here are the updated plots using the correct formulas:
image
image
image

semaphoreP: Hmm, why are the values in the colorbar in the "measured error" plot higher than before. Did something change? I would have thought that if we are still using the median for bias subtraction that the measured error shouldn't change?

ell-bogat: Yeah that's because I was accidentally using the formula for the mean before. So I fixed that and now the errors are more inflated.

ell-bogat: But the median bias measurement itself is unchanged

ell-bogat:
Based on chatting with Jason, I think this PR is ready to merge! Then we'll add the bias extension in a separate PR. Just let me know if y'all have any edits/questions.

ell-bogat:
Well, I fixed a test so that we're properly checking if the pointer links remain between the Dataset.all_err and Image.err attributes, and when I use the Dataset.add_err method it does in fact break the link. I have an idea to fix it though I'm not sure it's the most elegant. Also the Dataset.add_err method doesn't currently allow you to add a different 2D error array to each image in the dataset, so this would fix that as well.

semaphoreP (approved):

semaphoreP:
Looks good to me now. Just merged it in. Yay!

PR #60, opened 2024-03-08
Issues 54 and 55

kjl0025:
__init__.py added to tests and examples folders so that setup.py's find_packages() will find them.
Changed DATA_LEVEL to HIERARCH type to avoid warning.
Shortened comments below 80 characters to prevent a long card error.
Implemented random seeds for simulated data to avoid the possibility of test failure for a case of outlier simulated data.

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -221,23 +221,23 @@ class Image():
                     self.err = err.reshape((1,)+err.shape)
             else:
                 self.err = np.zeros((1,)+self.data.shape)
-                
+
             if dq is not None:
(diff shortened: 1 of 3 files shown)


semaphoreP (commented):
Thanks! For the fixes on the issues. 

Regarding the `__init__.py`, those folders shouldn't be part of the package. There's a distinction between the "package" and the entire code base. The test and examples folder are not meant to be imported by the user. We may or may not have the "examples" folder in the future as we add in a documentation section.

semaphoreP (approved):
Looks good. Thanks!

PR #61, opened 2024-03-08
Update installation and enable toggling of errors

semaphoreP:
corgidrp config file and caldb now save to `~/.corgidrp` folder by default

New setting to enable whether individual error terms are saved or not. By default, this is set to false. 

Hopefully this didn't break anything (waiting on tests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,8 @@ pip install -e .
 ```
 Then you can import `corgidrp` like any other python package!
 
-The installation will create a configuration file in your home directory called `.corgidrp`. 
-That configuration directory will be used to locate things on your computer such as the location of the calibration database. 
+The installation will create a configuration folder in your home directory called `.corgidrp`. 
+That configuration directory will be used to locate things on your computer such as the location of the calibration database and the pipeline configuration file. The configuration files stores setting such as whether to track each individual error term added to the noise. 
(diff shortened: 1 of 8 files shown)


semaphoreP:
actually this will certainly cause tests to fail. will need to change behaviors of tests before this can be merged

maxwellmb (commented):
I think that with the introduction of the config file, we should briefly write down (for other developers) its purpose so that people know when to use it and when not to use it (e.g. when the constants.py file is more appropriate). 

Otherwise this looks good to me. 

I think we should merge #63 into main first and then have you merge that here.

maxwellmb:
The other thing that just occurred to me: For reproducibility we may want to write the config file (or at least certain keywords) to fits headers.

JuergenSchreiber:
I agree with @maxwellmb. Documentation is essential. Good thing is to track the errors in the HISTORY header.

semaphoreP:
Ok I fixed the tests that were failing, added a keyword in the error HDU to indicate if we're tracking individual errors, and added to the README on when to use it. @maxwellmb can you look over again, and approve if it looks good?

maxwellmb (approved):
Ok, I think this overall looks good to me!

PR #63, opened 2024-03-12
Darksub error

JuergenSchreiber:
I moved the error determination of the dark to create_dark_calib. I assume there is only the one error for the dark, so no other error to propagate.


maxwellmb (approved):
This all looks good to me!

PR #71, opened 2024-04-05
Save Prescan Bias as 1D Fits Extension

ell-bogat:
Hello! This PR is to track the bias measured in the prescan bias subtraction step as a 1D, single precision array. There are changes in 4 files.
* `data.py`: 
    * Added a `bias` and `bias_hdr` attribute to the `Image` class to store the measured bias.
    * Modified the `Image.__init__()`, `Image.copy()`, and `Image.save()` methods to read in, copy, and save the bias as a new fits extension called "BIAS", and maintain only single point precision (`dtype=np.float32`). If a new Image object is initialized manually and no bias is specifies, the bias array is initialized as an array of all zeros.
* `l1_to_l2a.py`:
    * Saved the measured bias for each row in the image to the bias attribute of the Image class
    * Removed the 'MED_BIAS' header keyword which previously held just the median bias measured in that image, as it's now redundant.
* `test_prescan_sub.py`:
    * Added tests to make sure the bias array is the correct shape and data type
    * Added tests to make sure measured bias is correct within machine precision limits for an input array of all zero, and correct within the specified tolerance for a normal (hvoff) distribution.
    * Removed old/unnecessary comments & print statements
* `mocks.py`:
    * Fixed a bug that caused some mock data to be generated as integer datatype.

Example: performance for gaussian input data is as expected!
image

image

Just let me know if you have any thoughts/edits/questions!

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -153,22 +153,26 @@ class Image():
         ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw 2D data is passed in)
         err (np.array): 2-D/3-D uncertainty data
         dq (np.array): 2-D data quality, 0: good, 1: bad
+        bias (np.array): 1-D bias data
         err_hdr (astropy.io.fits.Header): the error extension header
         dq_hdr (astropy.io.fits.Header): the data quality extension header
+        bias_hdr (astropy.io.fits.Header): the bias extension header
 
     Attributes:
         data (np.array): 2-D data for this Image
         err (np.array): 2-D uncertainty
         dq (np.array): 2-D data quality
+        bias (np.array): 1-D bias data
         pri_hdr (astropy.io.fits.Header): primary header
         ext_hdr (astropy.io.fits.Header): image extension header. Generally this header will be edited/added to
         err_hdr (astropy.io.fits.Header): the error extension header
         dq_hdr (astropy.io.fits.Header): the data quality extension header
+        bias_hdr (astropy.io.fits.Header): the bias extension header
         filename (str): the filename corresponding to this Image
         filedir (str): the file directory on disk where this image is to be/already saved.
         filepath (str): full path to the file on disk (if it exists)
     """
-    def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq = None, err_hdr = None, dq_hdr = None):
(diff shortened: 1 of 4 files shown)


corgidrp/data.py
```diff
@@ -216,6 +220,17 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq
...
+                if bias is not None:
+                    if (np.shape(self.data)[0],) != np.shape(bias):
+                        raise ValueError("The shape of bias is {0} while we are expecting shape {1}".format(bias.shape, self.data.shape))
+                    self.bias = bias
+                # we assume that the bias extension is index 4 of hdulist
+                elif len(hdulist)>4:
+                    self.bias = hdulist[4].data
+                    self.bias_hdr = hdulist[4].header
+                else:
+                    self.bias = np.zeros(self.data.shape[0], dtype = np.float32)
```
semaphoreP, line 232: I forget the specifications now. Are we purposely saving it as float32 for a reason here?

tests/test_prescan_sub.py
```diff
@@ -400,8 +393,8 @@ def test_bias_hvoff():
             
             output_dataset = prescan_biassub(dataset, return_full_frame=return_full_frame)
 
-            # Compare median bias measurement to expectation
-            if np.abs(output_dataset[0].ext_hdr['MED_BIAS'] - bval) > tol:
+            # Compare bias measurement to expectation
+            if np.any(np.abs(output_dataset[0].bias - bval) > tol):
```
semaphoreP, line 397: Don't need to change this now, as it is clear what you are doing here, but `bias == pytest.approx(bval, tol)` is a thing you can use in the future.

semaphoreP (commented):
I think it looks good! I just had one main question on whether the bias array has to be `float32` (I forget the specs).

corgidrp/data.py
ell-bogat: I remember someone suggesting explicitly that we save the bias as an array of single-precision floats, maybe to reduce the data size, but I can't find exactly where the recommendation came from. If it doesn't matter, I can remove it!

semaphoreP: Ok, I see it in the specifications now. We actually haven't downsized the float/int size for the other HDUs, but I guess we can start here. Let's keep it then.

semaphoreP (approved):
I think it looks good.

Issue #49, opened 2024-02-21
Port L2a -> L2b JPL II&T code: Divide by EM gain


mygouf:

maxwellmb:
There is some useful discussion about this in Issue #46 (which I'm closing, to have it superseded by this one).

semaphoreP:
Baseline is to simply used the commanded EM gain. Assume this is in the header. See "Data header wishlist" section in "Data Formats" page.

JuergenSchreiber:
It is strange, in the cgi_iit_drp code they look for those keywords: 
        if 'EM_GAIN' in hdul[1].header:
            em_gain = hdul[1].header['EM_GAIN']
        if 'CMDGAIN' in hdul[1].header:  
            em_gain = hdul[1].header['CMDGAIN'] , so CMDGAIN would get the preference over EM_GAIN?

maxwellmb:
I believe CMDGAIN is the K-gain, not the commanded EM Gain. I've been.
Pushing to include that in the header as well.

JuergenSchreiber:
As far as I understand this: The K-gain is the number of electrons per DN (unit e/DN), and the EM_GAIN is the number of electrons activated by a photon (unit e/photons), is this correct?

maxwellmb:
The EM gain is more of a electrons-out per electron-in-from-your-pixel as I
understand it (after the electron multiplying process). Your understanding
of K-gain is the same as mine.

JuergenSchreiber:
Thanks, I see, it is the electron multiplication gain. So EM gain is number of electrons per photo-electron. Do we plan to have a unit keyword in the header, I don't find it in the data format document?

mygouf:
> It is strange, in the cgi_iit_drp code they look for those keywords: if 'EM_GAIN' in hdul[1].header: em_gain = hdul[1].header['EM_GAIN'] if 'CMDGAIN' in hdul[1].header: em_gain = hdul[1].header['CMDGAIN'] , so CMDGAIN would get the preference over EM_GAIN?

[...]

PR #73, opened 2024-04-10
issue #49 divide by em gain

JuergenSchreiber:
There are several things to clarify:
-I am a bit unsure about the unit naming after the division: "e/phot" or "photoelectrons"?
-I read the "CMDGAIN" value only from the first image of the dataset assuming the emgain does not change within a dataset, might not be correct in any cases?
-I added a dict parameter header_entries to Dataset.update_after_processing_step to update the headers and err_hdrs of all images in the dataset at once.

### DIFF
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -92,7 +92,7 @@ def em_gain_division(input_dataset):
     """
     
     Convert the data from electrons to photo-electrons by dividing the commanded em_gain. 
-    update the change in units in the header [e/phot].
+    update the change in units in the header [photoelectrons].
 
     Args:
         input_dataset (corgidrp.data.Dataset): a dataset of Images (L2a-level)
@@ -111,7 +111,7 @@ def em_gain_division(input_dataset):
     history_msg = "data divided by em_gain {0}".format(str(emgain))
 
     # update the output dataset with this em_gain divided data and update the history
-    emgain_dataset.update_after_processing_step(history_msg, new_all_data=emgain_cube, new_all_err=emgain_error, header_entries = {"BUNIT":"e/phot"})
+    emgain_dataset.update_after_processing_step(history_msg, new_all_data=emgain_cube, new_all_err=emgain_error, header_entries = {"BUNIT":"photoelectrons"})
 
     return emgain_dataset
     
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -103,12 +103,22 @@ def em_gain_division(input_dataset):
     
     # you should make a copy the dataset to start
     emgain_dataset = input_dataset.copy()
-
-    emgain = emgain_dataset[0].ext_hdr["CMDGAIN"]
-    emgain_cube = emgain_dataset.all_data/emgain
-    emgain_error = emgain_dataset.all_err/emgain
+    emgain_cube = emgain_dataset.all_data
+    emgain_error = emgain_dataset.all_err
     
-    history_msg = "data divided by em_gain {0}".format(str(emgain))
+    unique = True
+    emgain = emgain_dataset[0].ext_hdr["CMDGAIN"]
+    for i in range(len(emgain_dataset)): 
+        if emgain != emgain_dataset[i].ext_hdr["CMDGAIN"]:
+            unique = False
+            emgain = emgain_dataset[i].ext_hdr["CMDGAIN"] 
+        emgain_cube[i] /= emgain
+        emgain_error[i] /= emgain
+    
+    if unique:
+        history_msg = "data divided by em_gain {0}".format(str(emgain))
+    else:
+        history_msg = "data divided by non-unique em_gain"
 
     # update the output dataset with this em_gain divided data and update the history
     emgain_dataset.update_after_processing_step(history_msg, new_all_data=emgain_cube, new_all_err=emgain_error, header_entries = {"BUNIT":"photoelectrons"})
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -54,7 +54,7 @@ def dark_subtraction(input_dataset, dark_frame):
     history_msg = "Dark current subtracted using dark {0}".format(dark_frame.filename)
 
     # update the output dataset with this new dark subtracted data and update the history
-    darksub_dataset.update_after_processing_step(history_msg, new_all_data=darksub_cube)
+    darksub_dataset.update_after_processing_step(history_msg, new_all_data=darksub_cube, header_entries = {"BUNIT":"photoelectrons"})
 
     return darksub_dataset
 
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -62,6 +62,35 @@ def convert_to_electrons(input_dataset):
 
     return None 
 
+def em_gain_division(input_dataset):
+    """
+    
+    Convert the data from electrons to photo-electrons by dividing the commanded em_gain. 
+    update the change in units in the header [e/phot].
+
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (L2a-level)
+
+    Returns:
+        corgidrp.data.Dataset: a version of the input dataset with the data in units photo-electrons
+    """
+    
+    # you should make a copy the dataset to start
+    emgain_dataset = input_dataset.copy()
+
+    emgain = emgain_dataset[0].ext_hdr["CMDGAIN"]
+    emgain_cube = emgain_dataset.all_data/emgain
+    emgain_error = emgain_dataset.all_err/emgain
+    
+    history_msg = "data divided by em_gain {0}".format(str(emgain))
+
+    # update the output dataset with this em_gain divided data and update the history
+    emgain_dataset.update_after_processing_step(history_msg, new_all_data=emgain_cube, new_all_err=emgain_error, header_entries = {"BUNIT":"e/phot"})
+
+    return emgain_dataset
+    
+    
+
 def cti_correction(input_dataset):
     """
     
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -91,14 +91,14 @@ def convert_to_electrons(input_dataset):
 def em_gain_division(input_dataset):
     """
     
-    Convert the data from electrons to photo-electrons by dividing the commanded em_gain. 
-    update the change in units in the header [photoelectrons].
+    Convert the data from detected EM electrons to detected electrons by dividing the commanded em_gain. 
+    Update the change in units in the header [detected electrons].
 
     Args:
         input_dataset (corgidrp.data.Dataset): a dataset of Images (L2a-level)
 
     Returns:
-        corgidrp.data.Dataset: a version of the input dataset with the data in units photo-electrons
+        corgidrp.data.Dataset: a version of the input dataset with the data in units "detected electrons"
     """
     
     # you should make a copy the dataset to start
@@ -121,7 +121,7 @@ def em_gain_division(input_dataset):
         history_msg = "data divided by non-unique em_gain"
 
     # update the output dataset with this em_gain divided data and update the history
-    emgain_dataset.update_after_processing_step(history_msg, new_all_data=emgain_cube, new_all_err=emgain_error, header_entries = {"BUNIT":"photoelectrons"})
+    emgain_dataset.update_after_processing_step(history_msg, new_all_data=emgain_cube, new_all_err=emgain_error, header_entries = {"BUNIT":"detected electrons"})
 
     return emgain_dataset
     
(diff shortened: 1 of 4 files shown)


semaphoreP:
Might be good to ask @mygouf or @kjl0025 if they know whether CMDGAIN would ever change within a dataset? If we don't see any obvious case, we can leave it be for now, and change it into a for loop in the future.

semaphoreP:
For the units, I had thought the units would be "photons" after this conversion?

kjl0025:
It should be photoelectrons after gain division.  One would have to divide by the QE (quantum efficiency) to convert to photons.  In theory, that can vary per pixel.

semaphoreP:
It doesn't seem to me we have a QE calibration/correction step? Were we planning on such a step?

semaphoreP:
@kjl0025 and @mygouf, would one of you be interested in co-reviewing this with me, just to check the functionality is what you expect?

JuergenSchreiber:
> It doesn't seem to me we have a QE calibration/correction step? Were we planning on such a step?
> 
> Let's call it photoelectron then for the units!

OK, I think that depends on how we organize the absolute flux calibration. We could convert directly from DN to physical flux or starting from photoelectrons or photons etc. If you start from DN every conversion is hidden in the absolute flux calibration and you miss some information. So, it is a matter of what information you want to see before the flux calibration.

mygouf:
> @kjl0025 and @mygouf, would one of you be interested in co-reviewing this with me, just to check the functionality is what you expect?

Yes I can review tomorrow. Today's full with meetings and I have something urgent I need to finalize first.

corgidrp/l2a_to_l2b.py
```diff
@@ -88,6 +88,35 @@ def convert_to_electrons(input_dataset):
...
+    Returns:
+        corgidrp.data.Dataset: a version of the input dataset with the data in units photo-electrons
+    """
+    
+    # you should make a copy the dataset to start
+    emgain_dataset = input_dataset.copy()
+
+    emgain = emgain_dataset[0].ext_hdr["CMDGAIN"]
+    emgain_cube = emgain_dataset.all_data/emgain
+    emgain_error = emgain_dataset.all_err/emgain
```
kjl0025, line 109: I'm not sure why the error is divided by the gain.  The error is the uncertainty in knowledge of the signal photo-electrons?

corgidrp/l2a_to_l2b.py
```diff
@@ -88,6 +88,35 @@ def convert_to_electrons(input_dataset):
 
     return None 
 
+def em_gain_division(input_dataset):
+    """
+    
+    Convert the data from electrons to photo-electrons by dividing the commanded em_gain.
```
kjl0025, line 94: This designation of unit change may be a bit confusing to the user.  Any residual noise (due to dark current or clock-induced charge, for example) would be electrons and not signal photo-electrons.  I would argue that the step that most closely approximates the specification of electrons to photo-electrons would be dark subtraction.  Dividing by gain would simply get rid of the scaling to the signal.

corgidrp/l2a_to_l2b.py
JuergenSchreiber: Yes, I think the error should always be relative to the signal, so have the same units, also the different kind of errors.

mygouf:
> > @kjl0025 and @mygouf, would one of you be interested in co-reviewing this with me, just to check the functionality is what you expect?
> 
> Yes I can review tomorrow. Today's full with meetings and I have something urgent I need to finalize first.

- unit: what Kevin said => photoelectrons
- CMDGAIN: Please add a check to make sure all frames from the data set have the same commanded gain. This was done at in the II&T pipeline but at a different location. It wouldn't hurt performing this check directly in the em_gain_division function, unless there's a good reason to put it elsewhere. Thoughts? If we decide to put it somewhere else, let's add a new issue to not forget about it. 

Otherwise the function looks good to me.

mygouf:
> > It doesn't seem to me we have a QE calibration/correction step? Were we planning on such a step?
> > Let's call it photoelectron then for the units!
> 
> OK, I think that depends on how we organize the absolute flux calibration. We could convert directly from DN to physical flux or starting from photoelectrons or photons etc. If you start from DN every conversion is hidden in the absolute flux calibration and you miss some information. So, it is a matter of what information you want to see before the flux calibration.

Absolute flux calibration will be done at a higher level than level 4. We're in level 2 here and requirement is photoelectrons up to level 4.

maxwellmb:
> > > It doesn't seem to me we have a QE calibration/correction step? Were we planning on such a step?
> > > Let's call it photoelectron then for the units!
> > 
> > 
> > OK, I think that depends on how we organize the absolute flux calibration. We could convert directly from DN to physical flux or starting from photoelectrons or photons etc. If you start from DN every conversion is hidden in the absolute flux calibration and you miss some information. So, it is a matter of what information you want to see before the flux calibration.
> 
> Absolute flux calibration will be done at a higher level than level 4. We're in level 2 here and requirement is photoelectrons up to level 4.



> > > @kjl0025 and @mygouf, would one of you be interested in co-reviewing this with me, just to check the functionality is what you expect?
> > 
> > 
> > Yes I can review tomorrow. Today's full with meetings and I have something urgent I need to finalize first.
> 
>     * unit: what Kevin said => photoelectrons
> 
>     * CMDGAIN: Please add a check to make sure all frames from the data set have the same commanded gain. This was done at in the II&T pipeline but at a different location. It wouldn't hurt performing this check directly in the em_gain_division function, unless there's a good reason to put it elsewhere. Thoughts? If we decide to put it somewhere else, let's add a new issue to not forget about it.
> 
> 
> Otherwise the function looks good to me.


I feel like we had a discussion at some point about assuming uniform datasets for the step functions, but maybe checking elsewhere for dataset uniformity. I can't recall the outcome of that discussion. 

If we do decide we want to care about those things on the step-function level, then if in the end we're going to read every CMDGAIN anyway, why don't we just divide each image individually by whatever CMDGAIN is for that frame, such that the function isn't sensitive to different gains and is more generally applicable?

mygouf:
> > > > It doesn't seem to me we have a QE calibration/correction step? Were we planning on such a step?
> > > > Let's call it photoelectron then for the units!
> > > 
> > > 
> > > OK, I think that depends on how we organize the absolute flux calibration. We could convert directly from DN to physical flux or starting from photoelectrons or photons etc. If you start from DN every conversion is hidden in the absolute flux calibration and you miss some information. So, it is a matter of what information you want to see before the flux calibration.
> > 
> > 
> > Absolute flux calibration will be done at a higher level than level 4. We're in level 2 here and requirement is photoelectrons up to level 4.
> 
> > > > @kjl0025 and @mygouf, would one of you be interested in co-reviewing this with me, just to check the functionality is what you expect?
> > > 
> > > 
> > > Yes I can review tomorrow. Today's full with meetings and I have something urgent I need to finalize first.
> > 
> > 
> > ```
> > * unit: what Kevin said => photoelectrons
> > 
> > * CMDGAIN: Please add a check to make sure all frames from the data set have the same commanded gain. This was done at in the II&T pipeline but at a different location. It wouldn't hurt performing this check directly in the em_gain_division function, unless there's a good reason to put it elsewhere. Thoughts? If we decide to put it somewhere else, let's add a new issue to not forget about it.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Otherwise the function looks good to me.
> 
> I feel like we had a discussion at some point about assuming uniform datasets for the step functions, but maybe checking elsewhere for dataset uniformity. I can't recall the outcome of that discussion.
> 
> If we do decide we want to care about those things on the step-function level, then if in the end we're going to read every CMDGAIN anyway, why don't we just divide each image individually by whatever CMDGAIN is for that frame, such that the function isn't sensitive to different gains and is more generally applicable?

That works too!

JuergenSchreiber:
Thanks for the comments,
so I will divide the gain for every image

corgidrp/l2a_to_l2b.py
semaphoreP: For astronomers, we want to convert out of electrons at some point, so the point is to have a pipeline step that does this. We can put it after dark subtraction in the pipeline process if that makes sense. This should be done when every electron should be due to a photoelectron

kjl0025: I agree.  For reference, Roman doesn't do that until after L4 (after field-distortion correction).

semaphoreP: Our requirements for CGI require L2b products be in photoelectrons. So we need to do this somewhere in the L2a->L2b step.

kjl0025: Oh sorry, I was thinking "photon" instead of "photo-electron".  Yes, electron -> photo-electron would be via dark subtraction.

semaphoreP (approved):
Looks great to me!

semaphoreP:
@kjl0025 or @mygouf, if one of you approves the changes as well, we can merge this in

kjl0025 (changes requested):
I'm good with everything except for the units.  The units should change from electrons to photo-electrons later in the pipeline, via `dark_subtraction()`, and so the units coming in and leaving this gain division function would be electrons.  So I would like the updating of the header to be done in `dark_subtraction()` (whenever that function is written) instead of here.  And the doc string for this function would need to take out the part about unit conversion.

semaphoreP:
This feels like a definition issue of what should be in what units (the dark is being divided by the gain as well according to FDD specifications). It just feels a bit clunky to not change units when we are doing the scalar multiplication here.

JuergenSchreiber:
Would also be my question: Why should the dark subtraction change the units? In the JPL FDD Document the L2a -> L2b  output units are defined as: photoelectrons / pixel / frame.
Well, we could discuss whether my first idea of "e/phot" is a clearer description of the units, but in the end its the same.

rzellem:
> It doesn't seem to me we have a QE calibration/correction step? Were we planning on such a step?
> 
> Let's call it photoelectron then for the units!

Isn't the QE calibration/correction step the flatfield?

rzellem:
All of the different units are confusing me as well- I personally only think of units being either DN or photoelectrons. I do see @kjl0025 's point about needing some language to distinguish between taking into account the EM gain (specific to our EMCCD) vs. the "normal" gain (like QE variations and converting measured electrons on the CCD to photons that actually hit the detector).

Maybe standardizing that language, and maybe also documenting it somewhere on the JPL Wiki (like the FDD? or data level doc?) would help alleviate some confusion?

rzellem:
so @mygouf got me up to speed- sounds like there is a step that converts the images from DN to some new unit (whatever we want to call that) which contains *both* signal from photons hitting the detector *and* other annoying things like dark current; since it contains both of these signals, @kjl0025 is saying that we cannot say that the units are called "photoelectrons" because there's technically the other signal buried in there as well- and that we cannot call them "photoelectrons" until we do the dark subtract...and since we're providing the user data at all different data levels (which is awesome, imho), we need to figure out a new name to call these units (or at least something different than photoelectrons)

is that correct, @mygouf and @kjl0025 ?

kjl0025:
Yes, thanks @rzellem , that is what I mean.  (And the "unit conversion" from electron to photoelectron is "tricky" in the sense that it isn't due to a multiplicative factor of conversion but rather a subtraction.) But if electron is not distinguished from photoelectron in common terminology in this case, that's fine.  I just didn't think it made physical sense to say that EM gain division resulted in this unit conversion.  In any kind of frame, whether dark or an observation frame, EM gain would simply multiply the number of (photo)electron counts per pixel by an average factor, but I don't see any unit conversion happening.  Unless we want to define a unit of "gained (photo)electrons" or something like that?

And yes, I agree about the flat.  The flat division should effectively take care of non-uniformity of pixel response as well as QE in one fell swoop.  (I mistakenly referred to L4 above for conversion to photons from the Tech Demo FDD, but I think the table on that FDD meant something a little different when it talked about photons at that level.)

rzellem:
Thanks for confirming that I am following what's going on, @kjl0025 ! I was worried I wasn't understanding things fully :)

so could the image where DN->electrons by dividing by EM gain, but just before the dark subtraction happens, could the header just say "electrons" (is it really electrons though? is just the DN divided by the EMCCD gain or also the detector gain? well, technically both steps measure electrons in the system...just not "photoelectrons", as you've noted) and then maybe a little more details after that (to make it abundantly clear to a new user), and then the next step, where the dark is subtracted, call that "photoelectrons"?

rzellem:
and sorry if I'm just repeating things...I clearly need more coffee ☕

maxwellmb:
DN-> electrons is dividing by K-gain.

JuergenSchreiber:
I think the easiest way to get around this is to place the dark subtraction at the start of the L2 pipeline, where we still have DN units, isn't it?

maxwellmb:
@JuergenSchreiber, I've lost track of things. Is this ready to be re-reviewed or is there still something to be resolved?

semaphoreP:
I believe @kjl0025 just needs to see if the changes have sufficiently addressed his comments.

JuergenSchreiber:
@kjl0025 , I think I have addressed your requested changes, could you please review again? Thanks.

kjl0025 (approved):
Looks good!

PR #76, opened 2024-05-02
Recipe and pipeline-ization framework

semaphoreP:
This introduces the idea of recipes as JSON file and the Walker module. The walker module generates recipes from recipe templates when given an input dataset as well as executing recipes (running the dataset through the various step functions).

### DIFF
--- a/corgidrp/l1_to_l2a.py
+++ b/corgidrp/l1_to_l2a.py
@@ -168,9 +168,9 @@ def correct_nonlinearity(input_dataset, non_lin_correction):
 
 def update_to_l2a(input_dataset):
     """
-    Updates the data level to L2a
+    Updates the data level to L2a. Only works on L1 data. 
 
-    Currently does not do any verification. Just updates the header keyword
+    Currently only checks that data is at the L1 level
 
     Args:
         input_dataset (corgidrp.data.Dataset): a dataset of Images (L1-level)
@@ -178,13 +178,19 @@ def update_to_l2a(input_dataset):
     Returns:
         corgidrp.data.Dataset: same dataset now at L2-level
     """
+    # check that we are running this on L1 data
+    for orig_frame in input_dataset:
+        if orig_frame.ext_hdr['DATA_LEVEL'] != "L1":
+            err_msg = "{0} needs to be L1 data, but it is {1} data instead".format(orig_frame.filename, orig_frame.ext_hdr['DATA_LEVEL'] != "L1")
+            raise ValueError(err_msg)
+
     # we aren't altering the data
     updated_dataset = input_dataset.copy(copy_data=False)
 
     for frame in updated_dataset:
-        frame.ext_hdr['DATA_LEVEL'] = "L2"
+        frame.ext_hdr['DATA_LEVEL'] = "L2a"
 
-    history_msg = "Updated Data Level to L2"
+    history_msg = "Updated Data Level to L2a"
     updated_dataset.update_after_processing_step(history_msg)
 
     return updated_dataset
--- a/corgidrp/l1_to_l2a.py
+++ b/corgidrp/l1_to_l2a.py
@@ -114,7 +114,7 @@ def detect_cosmic_rays(input_dataset):
         corgidrp.data.Dataset: a version of the input dataset of the input dataset where the cosmic rays have been identified. 
     """
 
-    return None
+    return input_dataset
 
 def correct_nonlinearity(input_dataset, non_lin_correction):
     """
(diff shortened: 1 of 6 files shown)


corgidrp/walker.py
```diff
@@ -0,0 +1,128 @@
...
+all_steps = {
+    "prescan_biassub" : corgidrp.l1_to_l2a.prescan_biassub,
+    "detect_cosmic_rays" : corgidrp.l1_to_l2a.detect_cosmic_rays,
+    "correct_nonlinearity" : corgidrp.l1_to_l2a.correct_nonlinearity,
+    "update_to_l2a" : corgidrp.l1_to_l2a.update_to_l2a
+}
+
+recipe_dir = os.path.join(os.path.dirname(__file__), "recipe_templates")
+
+def autogen_reicpe(filelist, outputdir, template=None):
```
maxwellmb, line 18: typo in function name

tests/test_walker.py
```diff
@@ -0,0 +1,63 @@
...
+    prihdr, exthdr = mocks.create_default_headers("SCI")
+    new_nonlinearity.pri_hdr = prihdr
+    new_nonlinearity.ext_hdr = exthdr
+    new_nonlinearity.ext_hdr.set('DRPCTIME', time.Time.now().isot, "When this file was saved")
+    new_nonlinearity.ext_hdr.set('DRPVERSN', corgidrp.version, "corgidrp version that produced this file")
+    mycaldb = caldb.CalDB()
+    mycaldb.create_entry(new_nonlinearity)
+
+    # generate recipe
+    recipe = walker.autogen_reicpe(filelist, outputdir)
```
maxwellmb, line 49: similar typo in function name to above

corgidrp/walker.py
```diff
@@ -0,0 +1,128 @@
...
+        setattr(corgidrp, setting, recipe['drpconfig'][setting])
+
+    # read in data
+    filelist = recipe["inputs"]
+    curr_dataset = data.Dataset(filelist)
+    
+    # execute each pipeline step
+    for step in recipe["steps"]:
+        if step["name"].lower() == "save":
+            # special save instruction
```
maxwellmb, line 121: After this PR I think we'll want to add a feature request to enable adding suffixes easily for a "save" recipe call.

tests/test_walker.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # process_recipe
+    walker.run_recipe(recipe)
+
```
maxwellmb, line 53: I think want a test or two here at the end to make sure that it has run correctly. At the least if we check that the two new files were saved that ensures the recipe was run to the end.

maxwellmb (changes requested):
Overall I think this is looking great and fairly simple. Other than the minor comments, somethings we'll want to think about: codifying some of the expected recipe behaviour somewhere. For example, how to add in calibration filenames if you want to specify your own, how to use the drpconfig keyword, etc. Also, we'll want to mention that calibrations and recipe autogeneration are based solely on the first file, meaning that datasets are expected to be homogeneous (at least when it comes to cal files). All this we probably want in the implementation doc. 

I think the only real changes needed here is at least one more test after the walk is over.

corgidrp/walker.py
semaphoreP: good idea!

tests/test_walker.py
semaphoreP: Just added some checks to ensure the data can be loaded, and the recipe is written into the header!

corgidrp/l1_to_l2a.py
```diff
@@ -146,3 +146,26 @@ def correct_nonlinearity(input_dataset, non_lin_correction):
     linearized_dataset.update_after_processing_step(history_msg, new_all_data=linearized_cube)
 
     return linearized_dataset
+
+def update_to_l2a(input_dataset):
```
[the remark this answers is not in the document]
semaphoreP: I added this check, and added a test for it.

corgidrp/walker.py
semaphoreP: should be fixed now!

tests/test_walker.py
semaphoreP: should be fixed now!

corgidrp/l1_to_l2a.py
```diff
@@ -146,3 +146,26 @@ def correct_nonlinearity(input_dataset, non_lin_correction):
...
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (L1-level)
+
+    Returns:
+        corgidrp.data.Dataset: same dataset now at L2-level
+    """
+    # we aren't altering the data
+    updated_dataset = input_dataset.copy(copy_data=False)
+
+    for frame in updated_dataset:
+        frame.ext_hdr['DATA_LEVEL'] = "L2"
```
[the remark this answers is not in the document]
semaphoreP: Good catch!

maxwellmb:
Ok, so my last thought here is that I think we might want to check that the recipe saved to the header is the same as the input recipe. This keeps it so that we're sure if we were to re-run it, then we'd get the same thing.

semaphoreP:
Ok added that check on the recipe

maxwellmb (approved):

maxwellmb:
Things look good to me!

PR #79, opened 2024-05-07
bug_fixes_related to dark.err being only 2D

maxwellmb:
It turns out that when we create a new master Dark calibration file, letting it have a 2D error term, rather than a 3D error, can cause errors, in part because of the new ability to turn off tracking of the individual error terms. I suggest this bug fix that increase the dimension of the error array just after creation. There were a couple places where I had to include [0]s to address the new zeroth dimension.

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -206,9 +206,11 @@ class Image():
                 elif len(hdulist)>2:
                     self.err = hdulist[2].data
                     self.err_hdr = hdulist[2].header
+                    if self.err.ndim == 2:
+                        self.err = self.err.reshape((1,)+self.err.shape)
                 else:
                     self.err = np.zeros((1,)+self.data.shape)
-
+           
                 if dq is not None:
                     if np.shape(self.data) != np.shape(dq):
                         raise ValueError("The shape of dq is {0} while we are expecting shape {1}".format(dq.shape, self.data.shape))
--- a/corgidrp/detector.py
+++ b/corgidrp/detector.py
@@ -23,6 +23,7 @@ def create_dark_calib(dark_dataset):
     
     # determine the standard error of the mean: stddev/sqrt(n_frames)
     new_dark.err = np.nanstd(dark_dataset.all_data, axis=0)/np.sqrt(len(dark_dataset))
+    new_dark.err = new_dark.err.reshape((1,)+new_dark.err.shape) #Get it into the right dimensions
 
     return new_dark
 
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -46,7 +46,7 @@ def dark_subtraction(input_dataset, dark_frame):
     
     # propagate the error of the dark frame
     if hasattr(dark_frame, "err"):
-        darksub_dataset.add_error_term(dark_frame.err, "dark_error")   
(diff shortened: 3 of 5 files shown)


JuergenSchreiber:
Hi Max, yes, it looks fine, you can do it like this, although I do not quite understand why the master dark has to have a 3d error.

semaphoreP:
Max, can you please post a re-production of the bug to make it easier for other's to understand? It should just be a few lines of python.

semaphoreP:
[deleting my last comment]. It may be good to in the future to have separate attributes for total error and individual error terms in the Image class (separate from how it's stored in the FITS files) to avoid such indexing confusing in the future.

JuergenSchreiber:
Looks like an excellent idea

maxwellmb:
FYI this reproduces the bug: 
```
import os
import corgidrp.caldb as caldb
import corgidrp.mocks as mocks
import corgidrp.detector as detector

################################
##### Create a master Dark #####
################################

datadir = os.path.join(os.path.dirname(__file__), "simdata")
if not os.path.exists(datadir):
    os.mkdir(datadir)
dark_dataset = mocks.create_dark_calib_files(filedir=datadir)
dark_frame = detector.create_dark_calib(dark_dataset)

calibdir = os.path.join(os.path.dirname(__file__), "testcalib")
dark_filename = "sim_dark_calib.fits"
if not os.path.exists(calibdir):
        os.mkdir(calibdir)
dark_frame.save(filedir=calibdir, filename=dark_filename)

################################
##### Create a master Dark #####
################################


testcaldb_filepath = os.path.join(calibdir, "test_caldb.csv")
testcaldb = caldb.CalDB(filepath=testcaldb_filepath)

testcaldb.scan_dir_for_new_entries(calibdir)
```

and the error is as follows: 
image

maxwellmb:
@JuergenSchreiber, I also added a new commit that makes errors 3D when read in, if they aren't already. Still good to go?

semaphoreP:
Max, can you add a test for the bug you just fixed? based data the reproduction code and make it a unit test. That way we don't run into this case in the future (and it seems like it is fast the run, so it won't be a bottleneck to add)

maxwellmb:
> Max, can you add a test for the bug you just fixed? based data the reproduction code and make it a unit test. That way we don't run into this case in the future (and it seems like it is fast the run, so it won't be a bottleneck to add)

Test added! Good to go?

JuergenSchreiber:
Ok, now I looked a bit more deeper what the problem was. The sim_dark_calib.fits file was generated (detector.py) with a 2D Error, which is fine I think? The problem is now that the image class does not check whether we have a 2D error. So I think the following would solve this in the image __init__():
if not corgidrp.track_individual_errors and self.err.ndim > 2 and self.err.shape[0] > 1:

semaphoreP:
I think we're making the design decision to force err maps to always be 3-D (even if the 3rd dimension is 1).

maxwellmb:
Yes, that basically just means that always, no matter what the error should be 3D.

JuergenSchreiber:
Agreed, sorry, I always find it a bit "awkward" to have 3d error on 2d data, especially at calibration files like dark. Could be merged from my site.

maxwellmb:
If we want to think about restructuring I'm definitely open to the idea, but we should probably do so soon before we need to make commitments about the archiving of the data.

PR #80, opened 2024-05-07
E per dn

JuergenSchreiber:
I set up a new cal class KGain in data.py with a convenient property value, which is the kgain value. I set the unit
keyword to "detected EM electrons/DN", default filename "KGain.fits".
The function "convert_to_electrons" in l2a_to_l2b.py does the unit conversion by multiplying the kgain value with the data. A keyword with the kgain value is added, the unit after conversion is "detected EM electrons". Test is called test_kgain.py.

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -544,12 +544,13 @@ class KGain(Image):
     """
     Class for KGain calibration file. Until further insights it is just one float value.
 
-     Args:
+    Args:
         data_or_filepath (str or np.array): either the filepath to the FITS file to read in OR the calibration data. See above for the required format.
         pri_hdr (astropy.io.fits.Header): the primary header (required only if raw data is passed in)
         ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw data is passed in)
-     Attrs:
-        value (float): the value of kgain
+     
+    Attributes:
+        _kgain (float): the value of kgain
     """
     def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None):
 
@@ -582,15 +583,35 @@ class KGain(Image):
     @property
     def value(self):
         return self._kgain
-
-    def copy(self):
-        """
-        just return the reference to get the value field
-        Returns:
-           KGain: returns the copy
-        """
-        return self        
  
+    def copy(self, copy_data = True):
+        """
+        Make a copy of this KGain file. including data and headers.
+        Data copying can be turned off if you only want to modify the headers
+        Headers should always be copied as we should modify them any time we make new edits to the data
+
+        Args:
+            copy_data (bool): (optional) whether the data should be copied. Default is True
+
+        Returns:
+            corgidrp.data.KGain: a copy of this KGain
+        """
+        if copy_data:
+            new_data = np.copy(self.data)
+        else:
+            new_data = self.data # this is just pointer referencing
+        new_kg = KGain(new_data, pri_hdr=self.pri_hdr.copy(), ext_hdr=self.ext_hdr.copy())
+        
+        # annoying, but we got to manually update some parameters. Need to keep track of which ones to update
+        new_kg.filename = self.filename
+        new_kg.filedir = self.filedir
+
+        # update DRP version tracking
+        self.ext_hdr['DRPVERSN'] =  corgidrp.version
+        self.ext_hdr['DRPCTIME'] =  time.Time.now().isot
+
+        return new_kg
+  
         
 datatypes = { "Image" : Image,
               "Dark"  : Dark,
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -548,6 +548,8 @@ class KGain(Image):
         data_or_filepath (str or np.array): either the filepath to the FITS file to read in OR the calibration data. See above for the required format.
         pri_hdr (astropy.io.fits.Header): the primary header (required only if raw data is passed in)
         ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw data is passed in)
+     Attrs:
+        value (float): the value of kgain
     """
     def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None):
 
@@ -584,6 +586,8 @@ class KGain(Image):
     def copy(self):
         """
         just return the reference to get the value field
+        Returns:
+           KGain: returns the copy
         """
         return self        
  
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -549,7 +549,8 @@ class KGain(Image):
         pri_hdr (astropy.io.fits.Header): the primary header (required only if raw data is passed in)
         ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw data is passed in)
      
-    Attributes:
+    Attrs:
+        value: the getter of the kgain value
         _kgain (float): the value of kgain
     """
     def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None):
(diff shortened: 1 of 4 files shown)


corgidrp/data.py
```diff
@@ -551,10 +556,84 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+        # double check that this is actually a KGain file that got read in
+        # since if only a filepath was passed in, any file could have been read in
+        if 'DATATYPE' not in self.ext_hdr or self.ext_hdr['DATATYPE'] != 'KGain':
+            raise ValueError("File that was loaded was not a KGain Calibration file.")
+    
+    @property
+    def value(self):
+        return self._kgain
+ 
+    def copy(self, copy_data = True):
```
kjl0025, line 604: Why make a new copy() method?  Since the KGain class inherits the Image class, why not just use Image.copy()?  I don't think we'd want to track potential future errors that propagate through 2 different copy() methods.

corgidrp/data.py
```diff
@@ -551,10 +556,84 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+
+        # File format checks
+        if self.data.shape != (1,1):
+            raise ValueError('The KGain calibration data should be just one float value')
+
+        self._kgain = self.data[0,0] 
+        # additional bookkeeping for a calibration file
+        # if this is a new calibration file, we need to bookkeep it in the header
+        # b/c of logic in the super.__init__, we just need to check this to see if it is a new KGain file
+        if ext_hdr is not None:
```
kjl0025, line 585: If `data_or_filepath` were a file path, then the `Image` class doesn't require pri_hdr or ext_hdr, and then the new headers below are not added to the file, presumably because the user assumes the file represented by the file path already includes these KGain-specific headers.  The error case check in line 597 only applies to the case where a file path is read in.  So if we require a file read in to have the right header at least for 'DATATYPE' (so that not just any file is read in), by the same token we should require an input ext_hdr to already have these KGain-specific headers, in my opinion.  I know this is a thing that is done in all these data classes which inherit from Image in this Python file, but that is my thought for all of them.  Either that, or we simply add these KGain-specific headers in even in the case when a file path is used as the input, which would effectively take away the check that happens at line 597.

tests/test_kgain.py
```diff
@@ -0,0 +1,53 @@
...
+import os
+import numpy as np
+from corgidrp.mocks import create_default_headers
+import corgidrp.data as data
+import corgidrp.l2a_to_l2b as l2a_to_l2b
+import pytest
+
+def test_kgain():
+    """
+    test the KGain class and the calibration file and the unit conversion
```
kjl0025, line 11: I think you should also test the case where the KGain class input is a file path where pri_hdr and ext_hdr is None.

tests/test_kgain.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    #test KGain class and cal file
+    prhd, exthd = create_default_headers()
+    dat = np.ones([1024,1024]) * 2 
+    err = np.ones([1,1024,1024]) * 0.5
+    
+    gain_value = np.array([[9.55]])
+    kgain = data.KGain(gain_value, pri_hdr = prhd, ext_hdr = exthd)
+    kgain.save("kgain.fits")
```
kjl0025, line 20: Are we okay with saving a file to a local user's current directory and then deleting it in line 26?  I've written tests before that create a temp file instead...

kjl0025 (changes requested):
Sorry for the delay in my review.  It must have slipped through the cracks.  Here it is now, though.  My changes requested are in the comments.

corgidrp/data.py
JuergenSchreiber: See #81,, the copy of Image does not work.

tests/test_kgain.py
JuergenSchreiber: This is already done in line 24

corgidrp/data.py
JuergenSchreiber: I am not sure whether I understand what you mean. You are right, line 585 is just a check whether it is a new KGain file with only the data as input. Then the DATATYPE is added to the ext_hdr. It might already be in the ext_hdr input but then it does not harm to overwrite. As you stated the test at line 597 is only needed in case of a file as input. It checks whether there is by chance a wrong calibration file used as input. Maybe @semaphoreP could comment?

tests/test_kgain.py
JuergenSchreiber: Maybe @semaphoreP could comment. I could also create a caldb entry for KGain and read it again. The reason that I did this is that I did not want to leave a lot of  files around on the disk.

JuergenSchreiber:
Thanks @kjl0025 for reviewing. Another question: Up to now  I assumed that we know the kgain value and just save it as cal file. I think we should also write a creation routine in detector.py. I think I learned the kgain is the linear slope of signal vs stddev(signal)?

corgidrp/data.py
kjl0025: Okay, after thinking about it, I'm okay with that.  For me, it was really an issue of the philosophy of the check.  If a user can't get away with inputting a file path without a KGain-specific header, then by the same token a user shouldn't be able to enter an array without an ext_hdr that includes a KGain-specific header to minimize the chances of accidentally inputting non-KGain data.  But we aren't being draconian with inputs in general, and I think the typical user of these classes wouldn't mess this up, so I'm okay with this.  And I can understand the convenience of this in practice.

tests/test_kgain.py
kjl0025: Oh, my mistake, I see that now.

kjl0025:
I agree, a routine should be in detector.py.  I don't think it would involve export-controlled material (such as DACs), so I think a port of the code from GSW for that should be fine.  I am currently reviewing that code for GSW, and it is supposed to be done by the end of May.

tests/test_kgain.py
kjl0025: I think this may cause issues for a Windows user.  I think it would be better to use a the `tempfile` module for creating and deleting and cleaning up:
```
import tempfile
tmp_fd, tmp_fits = tempfile.mkstemp(suffix='.fits')
# do stuff with tmp_fits here

os.close(tmp_fd)
os.unlink(tmp_fits)
```

kjl0025 (changes requested):
I'm fine with this after the tempfile change.

tests/test_kgain.py
semaphoreP: Check out some of our other tests (e.g., https://github.com/roman-corgi/corgidrp/blob/main/tests/test_dark_sub.py). We do save files, but we save them inside the `tests` folder, and preferrably inside a subdirectory like `simdata` and `testcalib`. These folders generally won't be tracked, and avoids saving test files in the arbitrary current working directory. 

The use of the `os.path` module should remove any issues in paths between Mac/Windows/Linux.

semaphoreP: I'm a windows user, and can confirm that the `os` module should be OS agnostic.

kjl0025: I remember running into an issue with Windows not with the use of `os` but something about doing a particular thing with a file while it's already opened in Windows vs Mac, and `tempfile` solved the issue for me.  But it looks like it's not applicable in this case then since all this works, and I don't remember the context of my previous issue. :)

kjl0025 (approved):
I'm good to go with this.  I'll let Jason or Max pull the trigger.

PR #82, opened 2024-05-12
Add rowreadtime to detector.py

ghost:
I have added "rowreadtime" to detector.py through a get function. I usually add the units to the parameters names (sec in this case). If corgidrp prefers to not include them in the get function's name, let me know.

### DIFF
--- a/corgidrp/detector.py
+++ b/corgidrp/detector.py
@@ -223,7 +223,7 @@ def make_detector_areas(detector_areas, areas=('image', 'prescan', 'prescan_reli
         detector_areas[area] = detector_area_mask(detector_areas, area=area)
     return detector_areas
 
-def get_rowreadtime_sec(datetime=Time('2026-12-01 00:00:00', scale='utc'):
+def get_rowreadtime_sec(datetime=Time('2026-12-01 00:00:00', scale='utc')):
     """
     Get the value of readrowtime. The EMCCD is considered sensitive to the
     effects of radiation damage and, if this becomes a problem, one of the
@@ -237,20 +237,21 @@ def get_rowreadtime_sec(datetime=Time('2026-12-01 00:00:00', scale='utc'):
 
     Args:
         datetime (astropy Time object): Observation's starting date. Its default
-        value is sometime during the Roman mission.
+        value is sometime between the first collection of ground data (Full
+        Functional Tests) and the duration of the Roman Coronagraph mission.
 
     Returns:
         rowreadtime (float): Current value of rowreadtime in sec.
 
     """ 
-    # Launch datetime (should we store it somewhere else?)
-    datetime_launch = Time('2026-10-01 00:00:00', scale='utc')
+    # IIT datetime
+    datetime_iit = Time('2023-11-01 00:00:00', scale='utc')
     # Date well in the future to always fall in this case, unless rowreadtime
     # gets updated. One may add more datetime_# values to keep track of changes.
     datetime_1 = Time('2040-01-01 00:00:00', scale='utc')
     
-    if datetime < datetime_launch:
-        raise ValueError('The observation datetime cannot be before launch.')
+    if datetime < datetime_iit:
+        raise ValueError('The observation datetime cannot be earlier than first collected data on ground.')
     elif datetime < datetime_1:
         rowreadtime_sec = 223.5e-6
     else:
--- a/corgidrp/detector.py
+++ b/corgidrp/detector.py
@@ -222,3 +222,26 @@ def make_detector_areas(detector_areas, areas=('image', 'prescan', 'prescan_reli
     for area in areas:
         detector_areas[area] = detector_area_mask(detector_areas, area=area)
     return detector_areas
+
+def get_rowreadtime_sec():
+    """
+    Get the value of readrowtime. The EMCCD is considered sensitive to the
+    effects of radiation damage and, if this becomes a problem, one of the
+    mitigation techniques would be to change the row read time to reduce the
+    impact of charge traps.
+ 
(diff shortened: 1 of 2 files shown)


corgidrp/detector.py
```diff
@@ -222,3 +222,26 @@ def make_detector_areas(detector_areas, areas=('image', 'prescan', 'prescan_reli
     for area in areas:
         detector_areas[area] = detector_area_mask(detector_areas, area=area)
     return detector_areas
+
+def get_rowreadtime_sec():
```
semaphoreP, line 226: I would add an argument of the time of observation (e.g., as an astropy.time object). This could be an optional argument for now. In the future, we can check on the time of observation to determine which rowreadtime to use.

corgidrp/detector.py
```diff
@@ -222,3 +222,26 @@ def make_detector_areas(detector_areas, areas=('image', 'prescan', 'prescan_reli
...
+    There’s no formal plan/timeline for this adjustment,  though it is possible
+    to change in the future should it need to.
+
+    Its default value is 223.5e-6 sec.
+
+    Args:
+        None
+
+    Returns:
+        Current value of rowreadtime in sec.
```
semaphoreP, line 242: Please check the docstring examples of other functions here to format the syntax of the Args and Returns variables.

corgidrp/detector.py
```diff
@@ -222,3 +222,26 @@ def make_detector_areas(detector_areas, areas=('image', 'prescan', 'prescan_reli
...
+
+    """ 
+    rowreadtime_sec = 223.5e-6
+
+    return rowreadtime_sec
```
semaphoreP, line 247: Please also write a test for the function to demonstrate function behavior (I know it's trivial, but we want to get in the habit of writing tests for all the code we commit).

semaphoreP (changes requested):
Hi Sergi, a few changes requested. let me know if you have any questions about them.

hsergi:
I think so, although one could just define a variable at the top of detector.py before any function. It depedns on the coding style.

When I read this response in slack from @maxwellmb : "[@Sergi Hildebrandt]
, you could put it in detector.py I think for now", I looked at detector.py and all I found were functions. So, despite being overkilling, I did so. 

Any other option is fine with me. Just let me know

corgidrp/detector.py
hsergi: @semaphoreP: this function does not have arguments. I did not find an example in coorgidrp with no arguments. Is Args: None not acceptable? 

For the Return, it seems: " rowreadtime_sec (float): Current value of rowreadtime in sec." could be compliant with other examples.

maxwellmb:
Given that we expect that it could one day chance, I think keeping it as a function for now is fine, since that'll make it easier to implement a date-specific value if we need to. The most ideal case would be to have the value written into the L1 data products.

corgidrp/detector.py
hsergi: I now found the first comment. I'll add that argument. Good idea.

corgidrp/detector.py
hsergi: Should we use JD or ISO?

hsergi: In https://collaboration.ipac.caltech.edu/display/romancoronagraph/Header+Information, there's DATETIME, which is TBD UTC, and the other two keywords related to time are in TAI (SCTSRT, SCTEND). So, maybe UTC and  ISO 8601?

hsergi: I found 'DATE', '2024-02-10T00:17:48.172768' in tests/test_data/nonlin_sample.fits that Max created some time ago. I'll use then UTC and ISO8601

corgidrp/detector.py
semaphoreP: You just need to remove the "Args:" section if there are no arguments. But yeah, let's add one instead!

hsergi:
@semaphoreP I implemented the suggested changes. Let me know if you have any comments about them or any additional comments

corgidrp/detector.py
hsergi: @semaphoreP I am confused about your last comment below about using primitive types like None in the keyword of this function. I am not sure which comment said that. The only argument is the datetime. You asked: "I would add an argument of the time of observation (e.g., as an astropy.time object)" Do you mean not to use datetime=Time('2026-12-01 00:00:00', scale='utc') in the function definition? Can you clarify and provide an alternative if that's the case?

hsergi:
(Copying my comment here from an "outdated" thread above) 

@semaphoreP I am confused about your last comment below about using primitive types like None in the keyword of this function. I am not sure which comment said that. The only argument is the datetime. You asked: "I would add an argument of the time of observation (e.g., as an astropy.time object)" Do you mean not to use datetime=Time('2026-12-01 00:00:00', scale='utc') in the function definition? Can you clarify and provide an alternative if that's the case?

semaphoreP:
Even though you are expecting `datetime` to by an astropy.time.Time object, the default value of `datetime` should be `None` instead, because Python keywords should not use instantiated types as default values (default values should only be primitive types). Since there is no primitive type for astropy.time.Time, it should use `None` instead. Inside the function, if `None` is passed in, you should set `datetime` with a new instance of astropy.time.Time instead. 

For example, see the function definition of `corgidrp.walker.autogen_recipe` https://github.com/roman-corgi/corgidrp/blob/b3173895e28bf29649140f2b038b3ac1ea5b3045/corgidrp/walker.py#L18

hsergi:
Thanks. I've done that and added one more test using the default value. Please @semaphoreP review syntax and any other details before approval.

semaphoreP (approved):

semaphoreP:
Looks good!

PR #88, opened 2024-05-22
Flat calibration

ramya-anche:
Added functions to perform flat division using a master flat, as shown in the demo. 
1. Created a simulated dataset with 10 flat-fielded images.
2. Make one master flat by combining them 
3. Perform flat division of the simulated dataset using the master flat


corgidrp/l2a_to_l2b.py
```diff
@@ -58,6 +58,38 @@ def dark_subtraction(input_dataset, dark_frame):
...
+    
+     # copy of the dataset
+    flatcal_dataset = input_dataset.copy()
+    
+    #Divide by the master flat
+    flatcal_cube = flatcal_dataset.all_data/master_flat.data
+    
+    # propagate the error of the master flat frame # Check this error prop 
+    if hasattr(master_flat, "err"):
+        flatcal_dataset.add_error_term(master_flat.err, "masterflat_error")
```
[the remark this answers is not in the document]
maxwellmb: Also: 
Here the error in the test is happening because the dataset.add_error_term is expecting a 3D cube of shape: [n_frames,xdim,ydim]
or a 2D frame of shape [xdim,ydim]. Since we're trying to keep ALL of our image.err formats as 3D cubes [n_layer,xdim,ydim] (and we have hidden functions that try to enforce this when reading in data), I think you're inadvertantly passing in a 3D cube here.

tests/test_flat_div.py
```diff
@@ -0,0 +1,64 @@
...
+    flat_filepath = os.path.join(calibdir, flat_filename)
+    new_masterflat = data.Masterflat(flat_filepath)
+    # divide sim_data from masterflat
+    flat_dataset = l2a_to_l2b.flat_division(flat_dataset, new_masterflat)
+    #assert(flat_filename in str(flat_dataset[0].ext_hdr["HISTORY"]))
+
+    # check the level of the dataset is now approximately 0 
+    assert np.mean(flat_dataset.all_data) == pytest.approx(1, abs=1e-2)
+    # check the propagated errors
+    assert flat_dataset[0].err_hdr["Layer_2"] == "masterflat_error"
```
maxwellmb, line 56: I think the multi-layer error terms are turned off by default for testing so this might fail.

tests/test_flat_div.py
```diff
@@ -0,0 +1,64 @@
...
+    assert(np.mean(flat_dataset.all_err) == pytest.approx(np.mean(flat_frame.err), abs = 1e-2))
```
maxwellmb, line 57: If the data already has errors (which we expect it will for real data) this test will fail, so maybe the right test is to add errors to the mock data and still check here if it's correct (but the correct value will be different than just the flat_frame.err)

maxwellmb (changes requested):
just changing the status of my last review comments to be "Request changes"

corgidrp/detector.py
```diff
@@ -27,6 +27,27 @@ def create_dark_calib(dark_dataset):
 
     return new_dark
 
+def create_master_flat(flat_dataset):
```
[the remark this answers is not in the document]
mygouf: After confirming with Jason, we want the L2b processed flat data as an input.

maxwellmb:
@JuergenSchreiber, yes thanks for the error catch there! I had written out the errors for a multiplication even though I said the function was a division. You've got it correct for division.

corgidrp/l2a_to_l2b.py
```diff
@@ -58,6 +58,39 @@ def dark_subtraction(input_dataset, dark_frame):
...
+
+    Returns:
+        corgidrp.data.Dataset: a version of the input dataset with the flat field divided out
+    """
+    
+     # copy of the dataset
+    flatdiv_dataset = input_dataset.copy()
+    
+    #Divide by the master flat
+    flatdiv_cube = flatdiv_dataset.all_data /  flat_field.data
```
ecady-jpl, line 78: How are zero values in the master flat handled?  Are they special-cased in the division step, or precleaned/validated to ensure they can't exist, or...

ramya-anche: Currently, I don't think I am considering having zero values in the master flat when I generate them. In the test function, I have—assert np.mean(flat_frame.data) == pytest.approx(1, abs=1e-2), which would be a check.
Do you have any other suggestions I could add to the flat_division function?

maxwellmb:
Do we have to even do anything? I think the divide by zero will just create
a NaN, which isn't really a problem. And any pixel that is unresponsive
should be flagged by both the error extension of the flat field (when it's
created) and should be detected by the BadPixelMap creation.

corgidrp/data.py
```diff
@@ -440,6 +465,39 @@ def add_error_term(self, input_error, err_name):
...
+        if input_error.ndim != 2 or input_error.shape != self.data.shape:
+            raise ValueError("we expect a 2-dimensional error layer with dimensions {0}".format(self.data.shape))
+        
+        #first layer is always the updated combined error
+        self.err = self.err*input_error
+        self.err_hdr["Layer_1"] = "combined_error"
+
+        if corgidrp.track_individual_errors:
+            #append new error as layer on 3D cube
+            self.err=np.append(self.err, [input_error], axis=0)
```
maxwellmb, line 491: We should only use this for an additive error term.

corgidrp/data.py
```diff
@@ -440,6 +465,39 @@ def add_error_term(self, input_error, err_name):
...
+
+            layer = str(self.err.shape[0])
+            #self.err_hdr["Layer_1"] = "combined_error"
+            self.err_hdr["Layer_" + layer] = err_name
```
maxwellmb, line 495: Similarly, we don't need to update this.

corgidrp/detector.py
```diff
@@ -27,6 +27,27 @@ def create_dark_calib(dark_dataset):
...
+    """
+    Turn this dataset of image frames that were taken for performing the flat calibration and
+    to make one master flat image
+
+    Args:
+        flat_dataset (corgidrp.data.Dataset): a dataset of Image frames (L2a-level)
+
+    Returns:
+        data.masterflat: a master flat for flat calibration
```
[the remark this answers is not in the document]
maxwellmb: Don't forget to update the Return datatype here.

maxwellmb (changes requested):
A very small handful of final changes, then we're good to go! 

There's also now a merge conflict that maybe came from other recently PRs.

ecady-jpl:
> Do we have to even do anything? I think the divide by zero will just create a NaN, which isn't really a problem. And any pixel that is unresponsive should be flagged by both the error extension of the flat field (when it's created) and should be detected by the BadPixelMap creation.

I mean, I'm mostly looking to make sure the case is handled with intention.  If I do a quick check in a Python interpreter:
```
>>> import numpy as np
>>> a = np.ones((3, 3))
>>> b = np.zeros((3, 3))
>>> a/b
<stdin>:1: RuntimeWarning: divide by zero encountered in divide
array([[inf, inf, inf],
       [inf, inf, inf],
       [inf, inf, inf]])
```
which is infs rather than NaNs.

maxwellmb:
> > Do we have to even do anything? I think the divide by zero will just create a NaN, which isn't really a problem. And any pixel that is unresponsive should be flagged by both the error extension of the flat field (when it's created) and should be detected by the BadPixelMap creation.
> 
> I mean, I'm mostly looking to make sure the case is handled with intention. If I do a quick check in a Python interpreter:
> 
> ```
> >>> import numpy as np
> >>> a = np.ones((3, 3))
> >>> b = np.zeros((3, 3))
> >>> a/b
> <stdin>:1: RuntimeWarning: divide by zero encountered in divide
> array([[inf, inf, inf],
>        [inf, inf, inf],
>        [inf, inf, inf]])
> ```
> 
> which is infs rather than NaNs.

Yes sorry, these kinds of discussions are definitely good to have. I didn't mean to discourage discussion on it, and good point about the infs. 

For now this _should_ be covered by the bad pixel maps, but its worth being conscious of.

maxwellmb (approved):

PR #89, opened 2024-05-24
Desmear step function

hsergi:
Now that detetcor has the method to get rowreadtime, I am requesting the review of the desmearing step. It has a unit test associated to it: It checks data types and that the result of desmearing a smeared frame of 1024x1024 elements is equivalent within 1e-12 absolute error. Let me know any changes you'd like and any further tests.

### DIFF
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -1,5 +1,6 @@
 # A file that holds the functions that transmogrify l2a data to l2b data 
 import numpy as np
+import corgidrp.detector as detector
 
 def add_photon_noise(input_dataset):
     """
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -118,27 +118,39 @@ def flat_division(input_dataset, master_flat):
 
     return None
 
-def correct_bad_pixels(input_dataset):
+def desmear(input_dataset):
     """
-    
-    Compute bad pixel map and correct for bad pixels. 
 
-    MMB Notes: 
-        - We'll likely want to be able to accept an external bad pixel map, either 
-        from the CalDB or input by a user. 
-        - We may want to accept just a list of bad pixels from a user too, thus 
-        saving them the trouble of actually making their own map. 
-        - We may want flags to decide which pixels in the DQ we correct. We may 
-        not want to correct everything in the DQ extension
-        - Different bad pixels in the DQ may be corrected differently.
-
+    Desmear data frames.
 
     Args:
         input_dataset (corgidrp.data.Dataset): a dataset of Images (L2a-level)
 
     Returns:
-        corgidrp.data.Dataset: a version of the input dataset with bad pixels corrected
+        corgidrp.data.Dataset: a version of the input dataset with desmear applied
+
     """
 
-    return None
+    data = input_dataset.copy()
+    data_cube = data.all_data
+
+    rowreadtime_sec = detector.get_rowreadtime_sec()
+
+    for i in range(data_cube.shape[0]):
+        exptime_sec = float(datacube[i].ext_hdr['EXPTIME'])
+        smear = np.zeros_like(data_cube[i])
+        m = len(smear)
+        for r in range(m):
+            columnsum = 0
+            for s in range(r+1):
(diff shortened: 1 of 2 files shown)


corgidrp/l2a_to_l2b.py
```diff
@@ -157,27 +157,39 @@ def flat_division(input_dataset, master_flat):
...
+    data = input_dataset.copy()
+    data_cube = data.all_data
+
+    rowreadtime_sec = detector.get_rowreadtime_sec()
+
+    for i in range(data_cube.shape[0]):
+        exptime_sec = float(data[i].ext_hdr['EXPTIME'])
+        smear = np.zeros_like(data_cube[i])
+        m = len(smear)
+        for r in range(m):
```
semaphoreP, line 182: this double for loop here could be vectorized in the future if we need to speed up this function. However, I'm fine with optimizing for runtime later.

semaphoreP (approved):
Overall looks good to me. If we run into runtime issues, I think the function can be substantially sped-up with vectorization. However, this likely won't be the main thing slowing down the pipeline.

corgidrp/l2a_to_l2b.py
hsergi: Yeah. it is as-is in the IIT version.

corgidrp/l2a_to_l2b.py
```diff
@@ -157,27 +157,39 @@ def flat_division(input_dataset, master_flat):
...
-    MMB Notes: 
-        - We'll likely want to be able to accept an external bad pixel map, either 
-        from the CalDB or input by a user. 
-        - We may want to accept just a list of bad pixels from a user too, thus 
-        saving them the trouble of actually making their own map. 
-        - We may want flags to decide which pixels in the DQ we correct. We may 
-        not want to correct everything in the DQ extension
-        - Different bad pixels in the DQ may be corrected differently.
 
+    Desmear data frames.
```
maxwellmb, line 163: Please add more of a description as to what the function does

corgidrp/l2a_to_l2b.py
```diff
@@ -157,27 +157,39 @@ def flat_division(input_dataset, master_flat):
...
+        exptime_sec = float(data[i].ext_hdr['EXPTIME'])
+        smear = np.zeros_like(data_cube[i])
+        m = len(smear)
+        for r in range(m):
+            columnsum = 0
+            for s in range(r+1):
+                columnsum = columnsum + rowreadtime_sec/exptime_sec*((1 
+                + rowreadtime_sec/exptime_sec)**((s+1)-(r+1)-1))*data_cube[i,s,:]
+            smear[r,:] = columnsum
+        data_cube[i] -= smear
```
maxwellmb, line 188: Given that this is an analytical function, we should be able to calculate the added error that this operation applies to the data. Could please make an issue about this as a future to-do item so we can move this PR forward, but still keep track of the fact that it's not here now.

tests/test_desmear.py
```diff
@@ -0,0 +1,62 @@
...
+from corgidrp.mocks import create_default_headers
+from corgidrp.data import Image, Dataset, BadPixelMap
+
+old_err_tracking = corgidrp.track_individual_errors
+
+def test_desmear():
+    # Tolerance for comparisons
+    tol = 1e-12
+
+    corgidrp.track_individual_errors = True
```
maxwellmb, line 17: same comment as on the other PR. I think it's more important to test when this is False, since that's the standard use case.

corgidrp/l2a_to_l2b.py
hsergi: https://github.com/roman-corgi/corgidrp/issues/102

corgidrp/l2a_to_l2b.py
hsergi: Added. There's a question regarding the effective exposure time that results after applying desmearing. Once I know what to do about it, I can resubmit this PR, or work on any further step needed to deal with that

hsergi: It is an old comment that was inaccurate and does not apply any longer

tests/test_desmear.py
hsergi: Run tests succesfully

PR #92, opened 2024-05-28
Adding copy() method to BadPixelMap

hsergi:
As per Kevin's suggestion, I added a copy() method to BadPixelMap. It follows the implementation in the e_per_dn branch, KGain class in data.py, line 612.


kjl0025 (approved):
Looks good to me.  And when you are doing your tests for the bad-pixel step function, you can include a test that involves this copy method (similar to what is done in the e_per_dn branch).  I'll let @semaphoreP or @maxwellmb pull the trigger to merge this in.

PR #101, opened 2024-06-12
Add hash function to data

semaphoreP:
We now get obtain a unique str for any Image class that is a hash of the data, err, and dq maps. This is useful to quickly see in the caldb if any two entries are functionally the same. We aren't using it in the caldb yet for selection/matching, but we have the option to now. 

Should be pretty straightforward but let me know if there are any comments.


dsavransky (approved):
All looks good to me.

maxwellmb (approved):

Issue #68, opened 2024-03-22
We need a function that can do frame selection


maxwellmb:
We don't know how we'll select frames yet, but we still need this function. For now it can just check whether or not a frame is all NANs and if so get rid of it. That way we can get it set up and still test it.

maxwellmb:
If a frame is flagged as bad, it should set an IS_BAD header keyword, with the header keyword comment describing the reason it was rejected.

maxwellmb:
@nasavbailey can you confirm that for R3.0.1. this is a no-op code since we don't have any selection criteria yet?

nasavbailey:
I'd like to see what we can come up with for a beta version. Even if we aren't sure what the final criteria will be, can we define types of data we'd need to pass to enable frame selection, so we can define the function's interface? Your example of more than X% pix have bad DQ flag, where X is TBD, is a good one.  Another example, reject if RMS tip/tilt exceeds some TBD threshold.  Maybe there are other header keywords we'd also want to compare? 

So we want the ability to pass in the name of a keyword, a value, and a comparison type (=, !=, >, ≥, <, ≤, contains, does not contain). Can we define how we'd pass in those criteria, even if we haven't set them yet?

maxwellmb:
Another option would be to just code this into the step function as these things arise. In the case you mention, `bad_pixel_percentage` could be a kwarg to the step function, and we could add other kwargs like that (e.g. `max_rms_tiptilt`). Passing in all those things (keyword, value, operator) seems a bit clunky to me, but I'm sure we could figure out a way to do it that way if we wanted.

semaphoreP:
Do we need to define every aspect of the interface now? We can just add arguments in the future as we find useful filters. 

In my experience, I rarely useful keywords or DQ flags to throw away frames. I usually use some sort of image metric (correlation coefficient, least squares distance, flux at some separation, satellite spot properties), but it really depends on the data which one is effective. I find it hard to anticipate something that is very empirical in nature.

maxwellmb:
Based on a slack conversation with Jason and Vanessa, we will at least include keyword arguments that define a maximum tip/tilt residuals (from a currently TBD header keyword), and maximum % of bad pixels in the frame. We will also want to see if the OVEREXP keyword is set to True.

mygouf:
[...]

PR #104, opened 2024-06-14
Frame seelction interface.

semaphoreP:
This addresses the part of issue #68 that is part of the next release. After this is merged in, #68 can be moved to R3.0.2 milestone.


maxwellmb (approved):
easy peazy lemon squeezy

PR #105, opened 2024-06-15
function to create calibrated noise maps

kjl0025:
I ported over the function for making calibrated noise maps, which will be used by a step function I will port over to produce a synthesized master dark of a desired exposure time and EM gain.  The outputs of this PR's function will be used for making the master dark .fits files, along with the DQ and err maps that go into that.

I also moved the detector geometry parameters that were in detector.py into a .yaml file, and I added a class (Metadata) which can extract those numbers from the .yaml file, along with other parameters stored on the .yaml file.  I adjusted the function that used the detector geometry numbers (prescan_biassub) and the associated tests accordingly.  There was a function in detector.py which created a visualization of the SCI frame geometry which I also adjusted accordingly and put into corgidrp.util.metadata_visualize.py.


kjl0025:
I went ahead and added a calibration class for the noise maps and made the function that creates the master dark from these noise maps.

kjl0025:
Also tagging @maxwellmb for review since Jason is out on vacation, I believe.

kjl0025:
Okay, @maxwellmb , this is all good to go for review now.

maxwellmb:
@kjl0025 FYI, Jason and I have split up some of the PRs to review, and this one is on his list, so you'll probably have to wait until he's back.

tests/test_masterdark_from_noisemaps.py
```diff
@@ -0,0 +1,267 @@
+import unittest
```
semaphoreP, line 1: switch to pytest

tests/test_mean_combine.py
```diff
@@ -0,0 +1,175 @@
+import unittest
```
semaphoreP, line 1: switch to pytest

semaphoreP (changes requested):
Hi Kevin, I went through with some bigger picture comments. I think the main ones are:

 * I think we should restructure how NoiseMap and Darks are defined. I think they can be a single class, and the NoiseMap class should incorporate all of the arrays (not have separate instances for each map). 
 * Our DRP functions should input a single dataset and output a single dataset/calibration file. Thus, we shouldn't take a list of datasets, but rather one big dataset that we slice into smaller datasets inside the code. I'm happy to write some code to help slice a big dataset into smaller datasets by header values if that's useful.
 * I think we can centralize a lot of the noisemap/dark code into a single module. I don't want a proliferation of too many modules, especially if no other pieces of code use these functions.
 * The functions that take a detector_area dictionary as argument should have it be an optional keyword and default to the default one. This is to avoid conflicts with what Max is working on in parallel
 * Switch from unittest to pytest. It should be pretty straightforward. 
 * I'm also unsure what the BiasOffset class is for, that isn't already handled with how we are currently tracking bias offset values. 

Let me know what you think. We can discuss in the individual comment threads if that's easier.

corgidrp/data.py
```diff
@@ -459,47 +459,85 @@ def get_hash(self):
 
 class Dark(Image):
```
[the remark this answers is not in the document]
maxwellmb: Seems simpler to me!

corgidrp/data.py
```diff
@@ -724,6 +756,177 @@ def copy(self, copy_data = True):
...
+        new_nm.filename = self.filename
+        new_nm.filedir = self.filedir
+
+        # update DRP version tracking
+        self.ext_hdr['DRPVERSN'] =  corgidrp.version
+        self.ext_hdr['DRPCTIME'] =  time.Time.now().isot
+
+        return new_nm
+
+class BiasOffset(Image):
```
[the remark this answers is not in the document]
maxwellmb: Is it used in other steps or calibrations?

corgidrp/data.py
```diff
@@ -742,17 +945,56 @@ class DetectorParams(Image):
...
+        # Overhead time, in seconds, for each collected frame.  Used to compute
+        # total wall-clock time for data collection
+        'overhead': 3,
+        # Maximum allowed electrons/pixel/frame for photon counting
+        'pc_ecount_max': 0.1,
+        # number of read noise standard deviations at which to set the
+        # photon-counting threshold
+        'T_factor': 5,
+        ######################
+        # These below are ultimately sourced from calibration files
```
maxwellmb, line 975: I don't think I understand this comment. Are these parameters redundant?

corgidrp/calibrate_darks_lsq.py
```diff
@@ -0,0 +1,504 @@
...
+from corgidrp.detector import slice_section, imaging_slice, imaging_area_geom, detector_areas
+import corgidrp.check as check
+from corgidrp.mean_combine import mean_combine
+from corgidrp.data import NoiseMap, BiasOffset
+
+
+class CalDarksLSQException(Exception):
+    """Exception class for calibrate_darks_lsq."""
+
+def calibrate_darks_lsq(datasets, detector_params, d_areas=detector_areas):
```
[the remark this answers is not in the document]
maxwellmb: Throughout this PR it would be great to have variable names be more descriptive. It should be a quick find and replace. Some that caught my attention are d_area, all the F D C array-like things, that g_arr, t_arr, k_arr. The motivation here is future maintenance and code readability.

corgidrp/data.py
```diff
@@ -794,18 +1044,23 @@ def __init__(self, data_or_filepath, date_valid=None):
         self.params = {}
         # load back in all the values from the header
         for key in self.default_values:
-            self.params[key] = self.ext_hdr[key]
-        
+            if len(key) > 8:
```
maxwellmb, line 1047: Not a hard requirements, but might make things simpler: can we just keep the keys to 8 characters? I find headers really hard to read when they include hierarchical keys. Worth a discussion

corgidrp/l2a_to_l2b.py
```diff
@@ -43,24 +44,28 @@ def dark_subtraction(input_dataset, dark_frame):
...
     # propagate the error of the dark frame
     if hasattr(dark_frame, "err"):
-        darksub_dataset.add_error_term(dark_frame.err[0], "dark_error")   
+        darksub_dataset.add_error_term(dark_frame.err[0], "dark_error")
     else:
         raise Warning("no error attribute in the dark frame")
-    
+
+    if hasattr(dark_frame, "dq"):
+        new_all_dq = darksub_dataset.all_dq + dark_frame.dq[0]
```
maxwellmb, line 55: It’s not immediately clear to me that we want to combine DQs between calibrations and data, but I haven’t thought it through really. Do you have an example of a case where we’d want that?

should we move forward with it, we’ll want to combine the DQs using numpy.bitwise_or

corgidrp/l2a_to_l2b.py
```diff
@@ -43,24 +44,28 @@ def dark_subtraction(input_dataset, dark_frame):
...
+    else:
+        new_all_dq = None
+
     #darksub_dataset.all_err = np.array([frame.err for frame in darksub_dataset.frames])
-    history_msg = "Dark current subtracted using dark {0}".format(dark_frame.filename)
+    history_msg = "Dark subtracted using dark {0}".format(dark_frame.filename)
```
maxwellmb, line 60: It would also be good to include a message about changing units

corgidrp/l2a_to_l2b.py
```diff
@@ -172,6 +177,17 @@ def flat_division(input_dataset, master_flat):
 
 def desmear(input_dataset, detector_params):
     """
```
maxwellmb, line 179: I think these are here by accident

maxwellmb: (The comments)

corgidrp/masterdark_from_noisemaps.py
```diff
@@ -0,0 +1,116 @@
...
+    in the course of acquisition, alignment, and HOWFSC.  Better to take data
+    sets that don't vary.
+
+    Output is a bias-subtracted, gain-divided master dark in electrons.
+    (Bias is inherently subtracted as we don't use it as
+    one of the building blocks to assemble the dark frame.)
+
+    Master dark = (F + g*t*D + g*C)/g = F/g + t*D + C
+
+    Arguments:
```
maxwellmb, line 26: Same comment as before: please make the variable names more descriptive.

corgidrp/mocks.py
```diff
@@ -1,12 +1,204 @@
...
+            'r0c0': [104, 108]
+            },
+        'serial_overscan' : {
+            'rows': 220,
+            'cols': 5,
+            'r0c0': [0, 215]
+            },
+        }
+}
+
```
maxwellmb, line 70: Same as above, for all the testing more descriptive variable names would be ideal

corgidrp/data.py
maxwellmb: (just to emphasize that this is pretty low priority and is probably fine as is)

corgidrp/data.py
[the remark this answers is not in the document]
kjl0025: This class exists because bias offset is a calibration product that comes out of the calibrate_darks_lsq() function.  That value would then be an input for the prescan bias subtraction step function (or I guess that function would have as an input an instance of the BiasOffset class).

corgidrp/mean_combine.py
```diff
@@ -0,0 +1,119 @@
+import numpy as np
```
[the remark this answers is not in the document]
kjl0025: I left it as a separate function because other things in the pipeline (https://collaboration.ipac.caltech.edu/pages/viewpage.action?spaceKey=romancoronagraph&title=Corgi-DRP+Implementation+Document#CorgiDRPImplementationDocument-II&Tmeancombine) use this (things that haven't been added to the DRP yet).

corgidrp/data.py
kjl0025: I mean that the parameters below come from calibration activities.  So if the plan is to exclusively draw calibration numbers from the corresponding calibration class instances in data.py, then yes, these are redundant.

corgidrp/l2a_to_l2b.py
kjl0025: This is the sum of the flag values (similar to how it is done here: https://github.com/roman-corgi/corgidrp/blob/e86d27a700bee3d47e5d42d78bbecd71507f63d6/corgidrp/l1_to_l2a.py#L216).  So in the end, this value can be done via bitwise_or, but the flag values (powers of 2) are added and stored in the dq.

corgidrp/data.py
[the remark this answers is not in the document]
semaphoreP: Can we just integrate it as part of the NoiseMaps class?

corgidrp/mean_combine.py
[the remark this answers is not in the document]
semaphoreP: What functions are those?

corgidrp/data.py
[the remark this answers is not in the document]
maxwellmb: Ahhh, I see. This would be the input to the prescan_bias that is currently a `bias_offset` keyword? 

`def prescan_biassub(input_dataset, bias_offset=0., return_full_frame=False):`

corgidrp/data.py
maxwellmb: I guess from the comment I just didn't understand what calibration files you're talking about here. If they're derived from other existing calibration files in the corgi-DRP then I don't think we want them here too, right?

corgidrp/l2a_to_l2b.py
maxwellmb: I think ultimately its most robust to use bitwise_or both here and in the cosmic ray code. Doing it as a sum puts assumptions on what's already in the all_dq file, and the bitwse_or does not.

kjl0025:
Should be all good to go.  Note:  I added one new flag to calibrate_darks_lsq.py and to another function in darks.py (and Irenamed the containing file for that function to darks.py as you suggested).  The new flag is 256.  Details are in the doc string for that function.

semaphoreP (changes requested):
Just two more comments (I think the amount of code this changes is minor), but otherwise good. In the future, please avoid running a refactoring program on the entire repository to limit the number of files that get changed in a PR.

kjl0025:
I'll make these changes soon.  And sorry about all the weird spacing/deletions that happened there.  I'm really not sure why that happens b/c I didn't run any refactoring program.  I'll look into that.  I use VS Code to edit, and that hasn't caused this issue in the past.

kjl0025:
Also, just wanted to note that I removed the read noise output from `calibrate_darks_lsq()`, and the official calibrated read noise should come from the output of the k gain calibration function.  I don't think the read noise is needed in the processing pipeline, but I know it is used in the exposure time calculator and in HOWFSC.

semaphoreP:
Regarding the auto-code linting in VS Code, you could always check the gif diff to make sure to only commit relevant changes, and not commit files with auto-code linting steps.

semaphoreP (approved):
Looks good to me. I think we are ready to merge

PR #106, opened 2024-06-16
Add detector params class

semaphoreP:
This class stores semi-static detector parameters such as rowreadtime and allows us to track it like a calibration file. It also creates a default calibration file by default on installation with the default values for these parameters.


semaphoreP:
I had to change the install behavior to create the configuration files for the pipeline on first import of corgidrp, which should happen any time we need to access a variable in the configuration file. So I don't anticipate it being an issue.

corgidrp/caldb.py
```diff
@@ -26,7 +26,11 @@
     "EXCAMT",
 ]
 
-labels = {data.Dark: "Dark", data.NonLinearityCalibration: "NonLinearityCalibration", data.BadPixelMap: "BadPixelMap", "KGain": data.KGain, }
+labels = {data.Dark: "Dark", 
+          data.NonLinearityCalibration: "NonLinearityCalibration", 
+          data.BadPixelMap: "BadPixelMap", 
+          data.KGain : "KGain", 
+          data.DetectorParams : "DetectorParams"}
```
maxwellmb, line 33: Not a requirement but just a thought: Is there a way to combine this dictionary with the `datatypes` dictionary from data.py. They look like they do very similar things, but are maybe the inverse of each other. If possible having one would be great.

corgidrp/data.py
```diff
@@ -724,11 +724,110 @@ def copy(self, copy_data = True):
...
+    default_values = {
+        'kgain' : 8.7,
+        'fwc_pp' : 90000.,
+        'fwc_em' : 100000.,
+        'rowreadtime' : 223.5e-6 # seconds
+    }
+
+    def __init__(self, data_or_filepath, date_valid=None):
+
+        # if filepaht passed in, just load in from disk as usual
```
maxwellmb, line 750: minor typo

corgidrp/data.py
```diff
@@ -724,11 +724,110 @@ def copy(self, copy_data = True):
...
+        if isinstance(data_or_filepath, str):
+            # run the image class contructor
+            super().__init__(data_or_filepath)
+
+            # double check that this is actually a bad pixel map that got read in
```
maxwellmb, line 755: minor typo: not a bad pxiel map

corgidrp/data.py
```diff
@@ -724,11 +724,110 @@ def copy(self, copy_data = True):
 
         return new_bp
 
+class DetectorParams(Image):
+    """
+    Class containing detector parameters that may change over time
```
maxwellmb, line 729: It would be good to state here how this should be used. It looks like like a user will read keywords from the extension header, but it would be good to lay that out.

maxwellmb: Oh, or is it to use the .params dictionary?

maxwellmb (approved):
I'm going to let this through, but I'm going to make an issue that the docstring for the DetectorParams should be a bit more informative.

Issue #84, opened 2024-05-14
The add_photon_noise function should check the EM Gain to see if an extra noise term needs to be applied


maxwellmb:
When the EM Gain is active (and we're not photon-counting) the shot noise picks up some extra variance, we should account for this in the `add_photon_noise` function.

JuergenSchreiber:
@maxwellmb: Do you mean the sqrt(2) factor of excess noise? So, we have to multiplicate the photon noise by sqrt(2) if CMDGAIN > 1? How do we recognize whether we are photon-counting or not? Header keyword?

maxwellmb:
Yes, exactly the sqrt(2) factor of excess noise.

> So, we have to multiplicate the photon noise by sqrt(2) if CMDGAIN > 1?
That is my understanding.

> How do we recognize whether we are photon-counting or not? Header keyword?
I think this doesn't depend on us photon-counting, just on us having
CMDGAIN > 1.

Issue #25, opened 2024-01-30
Add image photon noise to error map


semaphoreP:
We need a step that computes the photon noise of the science image itself, and add that to the error map

JuergenSchreiber:
That looks like an effort I would be happy to contribute. But that will also raise questions about error propagation and how to organize that. For MIRI they have lots of extensions with different kind of errors like flatfield error, read noise and a combined uncertainty.

maxwellmb:
Yeah, I was thinking about this too. I'm wary of the proliferation of number of extensions and would also prefer using the 3-D image approach to this. I can see it being worth it to keep the different terms around sometimes, as it can help evaluate what the limiting noise term is.

JuergenSchreiber:
The photon noise seems to be just the sqrt of the signal (e/s) (shot noise), is this correct?

maxwellmb:
Yup! We may want to combine adding in the photon noise with the step that
converts from DN (ADU) to electrons. Seems like a logical place to this,
right?

semaphoreP:
Sounds reasonable. It should just be done before any dark subtraction.

JuergenSchreiber:
Should it be done also before prescan_biassub, so in l1_to_l2a.py? I guess not, so a starter in l2a_to_l2b.py?

maxwellmb:
I think there is noise associated with measuring and subtracting the bias, but I don't think that it's the same as photon noise.

PR #107, opened 2024-06-18
issue #84 excess noise

JuergenSchreiber:
multiplied sqrt(2) to photon/shot noise in case of EMGAIN > 1 -> excess noise


kjl0025:
Is the purpose of this function to estimate the noise due to the photons hitting the detector?  Or the noise of all detected EM electrons?  If just the photons, I think this step should happen after dark subtraction.  And taking the square root of each pixel would not be an estimate of the mean count from the Poisson distribution governing the photon detection process.  If a region of the detector were trustworthy as being due solely to a stellar target (with negligible effects from jitter, etc), then I suppose an average could be done over that region to find the mean count, and the noise would be the square root of that. 
 
I thought the "noise" in SNR, per pixel, would come from taking the standard deviation for a given pixel over several observation frames, and that would include all possible noise sources (photon noise and non-photon noise). An estimated noise for a single frame, especially if a trusted region of interest is not easily identified on a frame, could come from a theoretical expression using the model of the known noise sources (i.e., square root of a read noise squared (Gaussian noise) plus extra noise factor times photon noise plus a contribution due to FPN).
 
Also, the extra noise factor is a function of gain and number of gain stages (which is 604 for the EXCAM EMCCD, which should be found in the config data) which does approach sqrt(2) for large gain values.

JuergenSchreiber:
@kjl0025 well, I am not an expert on detector noises, maybe @maxwellmb should reply. But in my poor understanding the photon or shot noise has a Poisson distribution with a variance of the mean value, so the signal. You might be right in applying it after dark subtraction since this is not a photon noise? I think the standard application is high gain values?

maxwellmb:
Thanks for the useful discussion!!
My first thought is that maybe we want to build a photon-transfer curve,
which we can do with an estimate of the read noise and the Kgain. That way
we could apply a uniform treatment to any pixel value. If we're tracking
individual error terms then we'd want to make sure both components are
included. In the read noise dominates regime maybe the photo noise just
gets set to zero though.

When exactly to apply it is also good to think about. Since dark current is
a random process, I'm fairly sure it also introduces a Poisson noise term.

Remind me, when are we currently applying the poisson noise?

JuergenSchreiber:
Well, it is an independent step and can be placed where you want, in the initial issue #25 we said, it should be placed before the dark subtraction

kjl0025:
The code that Sergei is working on porting over (PR #116 ) creates a photon-transfer curve.  However, I'm just not clear on the purpose of this step function.  Ultimately, I would think we would want to know the total noise so that we can ultimately get the SNR, right?  I'm not sure why we want to isolate the photon noise specifically.  With a single frame, the noise per pixel is not really possible to read off (unless there is a reliable region of pixels over which the E field is the same, i.e., there needs to be multiple measurements of the same field in order to have items over which to take the standard deviation).  I think the best way to get the noise per pixel is to have multiple frames for a given observation so that for each pixel the standard deviation can be obtained.  And that noise would account for photon and non-photon noise (and the non-photon mean is removed from a frame via dark subtraction, but the non-photon noise is not removed by dark subtraction).  Taking the standard deviation over a stack gets the noise per pixel, whereas taking the square root of a pixel value (even if the frame is an averaged frame) is at best an approximation of the noise, photon or not.  In other words, each pixel value is not a Poisson mean but just a Poisson variate (times an extra noise factor plus a Gaussian variate plus any contribution from FPN), so I don't see how taking the square root of the pixel value gives us anything.

maxwellmb:
We'd like to keep track of the expected error for these different terms,
which is the whole point of the error extension. I agree that it will also
be useful to compare it to a set of images. Comparing the measured standard
deviation vs. the expected error will be critical for understanding the
actual error budget.

kjl0025:
Since the input to this function is a dataset, perhaps this function could take the standard deviation over the dataset stack per pixel (we'll call that 'std', a value different for each pixel) and assign each pixel in each frame in the dataset to have an error of std/sqrt(N), where N is the number of frames.... Actually, to take into account pixels that are masked in the DQ maps, one could use a `mean_combine`-type function that is used in the CTC pipeline (which I call in my PR #105 ) to handle the DQ maps when calculating the standard deviation.  In the end, if the plan aligns with the CTC pipeline, which mean-combines the frames after they have been processed, the `err` of the mean-combined frame could be calculated at this stage in the pipleline, and this standard deviation over the stack I've spoken of could be added in quadrature with any otherwise errors from each frame.  Do we want the individualized version of that error (i.e., `mean_combine`-type standard deviation divided by sqrt(N) assigned to each frame in the dataset), or would calculating this `mean_combine`-type standard deviation for the final mean-combined frame suffice?   Either way, I could write such a function and do a PR for it if that is agreeable to you both.

JuergenSchreiber:
@kjl0025, you are right, we cannot isolate single noise sources. You made a proposal to calculate the measured noise by using a sufficient number of frames in a dataset (problem of low number statistics?), while in the step the noise of each single frame is calculated. 
After reading a bit about detector noise, the main noise sources are photon noise, dark noise and read noise. Photon noise dominates pixels with more or less bright sources, where the sqrt(signal) seems to be a good estimate of the noise.

JuergenSchreiber:
To move on with this PR I try to summarize the points Kevin made. I am not an expert in detector noise, so please correct and comment on this.
1) Kevin doubts that we can distinguish between noise sources like read noise, photon noise etc. We only could determine the overall noise statistically from a high enough number of frames in a dataset.
2) We could anyhow use the PTC to determine/distinguish photon noise/ read noise? How?
I see some options: 
- Keep this step function as is as an estimate of photon noise and decide later whether to apply it where or not at all in the pipeline.
- add a new issue for a new step function determine_noise, as an estimate of the overall noise calculated as Kevin suggests.
- add a new issue for a new step function that can determine different noise sources from the PTC.

semaphoreP:
It looks like to me that this PR does help improve our existing noise measurements, and it doesn't require multiple files. It sounds like in the future, there are more intelligent ways to calculate noise with a large number of files, but that requires further work and more lines of code. Also, let's remember that the error maps for the science data are not a requirement. It's something we want to do. 

In line with Juergen's suggestion, I think it makes sense that we merge in this PR now, and open a future issue to have better error estimates. Let me know if anyone disagrees.

kjl0025:
I'm okay with that.  Sqrt(2) is only good for the extra noise for large gain, but the full function for extra noise would require getting the number of gain registers (604) from DetectorParams or detector.py (i.e., more lines of code), so I'm okay shifting the work to a future push.  

Just to try to document everything in one place, I'll try to summarize the stuff from Slack here, for the case of having noise on a per frame basis and not in a `mean_combine` sense as described in above comments:

The photon noise can't really reliably be separated for a single pixel from non-photon noise (like CIC, dark current, read noise, and FPN); dark subtraction subtracts the mean, but the noise is still present.  We could include the non-photon noise, though (say, for getting the SNR).  However, if a luminous target is being observed, the noise dominating the pixel would be photon noise, of course. 

For an estimate of the photon + non-photon noise, we could use
sqrt(rn^2 + ENF^2*other_noise^2 + FPN noise contribution)
rn: read noise, output from k gain function
other_noise: could be estimated as sqrt(pixel value) or could be estimated more accurately from the PTC curve (output from the k gain code), but how we deal with it depends on what EM gain value the PTC was taken at (see below)
ENF: extra noise function due to EM gain, defined below
FPN contribution:  usually estimated to be 0 b/c no good systematic way to deal with it and varies pixel to pixel

The pixel value is really a Gaussian variate (read noise) plus a Poisson variate (dark stuff plus photon signal) plus FPN, but since the mean of read noise is 0 (and since FPN is harder to handle and has roughly 0 noise), it would be approximately correct to assume the pixel value is Poisson (photon + non-photon) noise even though the pixel value is really a variate and not a mean value.

The photon transfer curve (PTC) is one output of the k gain code.  I could see the photon noise step function having an input for that (coming from the KGain class).  And we could require the input PTC be for EM gain=1 and then multiply the result by the extra noise function (ENF):
```
def ENF(g, Nem):
    """Returns the extra-noise function (ENF).
    Parameters
    ----------
    g : float
        EM gain.  >= 1.
    Nem : int
        Number of gain register cells.
    Returns
    -------
    ENF : float
        extra-noise function
    """
    return np.sqrt(2*(g-1)*g**(-(Nem+1)/Nem) + 1/g)
    ```
where Nem (which is 604) would come from the DetectorParams class, and this function does go to sqrt(2) for large values of g.  Or if we had a PTC for some EM gain g1 and knew that the frame in question had g2, we could always multiply by ENF(g2,Nem)/ENF(g1, Nem), although a PTC for unity EM gain would be ideal.  And this would be done before dark subtraction for this case since this curve would include contributions due to Poisson dark sources like CIC and dark current (and the FPN does not appear on the linear part of the curve, nor does the read noise).

semaphoreP (approved):
Approving this for now, acknowledging we can be doing better, but that's a new issue.

PR #113, opened 2024-06-21
Boresight calibration

semaphoreP:
Making this for @manduhmia


corgidrp/astrom.py
```diff
@@ -0,0 +1,387 @@
...
+import astropy
+import astropy.io.ascii as ascii
+from astropy.coordinates import SkyCoord
+
+import pyklip.fakes as fakes
+
+import scipy.ndimage as ndi
+import scipy.optimize as optimize
+
+def centroid(frame):
```
semaphoreP, line 14: these helper functions should have docstrings (they can be simple)

corgidrp/astrom.py
```diff
@@ -0,0 +1,387 @@
+import numpy as np
+
+import corgidrp.data
+
+import astropy
+import astropy.io.ascii as ascii
+from astropy.coordinates import SkyCoord
+
+import pyklip.fakes as fakes
```
semaphoreP, line 9: you'll need to add pyklip to requirements.txt

corgidrp/astrom.py
```diff
@@ -0,0 +1,387 @@
...
+    xstamp += stampsize//2 + dx
+    ystamp += stampsize//2 + dy
+    
+    return ndi.map_coordinates(frame * flux, [ystamp, xstamp], mode='constant', cval=0.0).ravel()
+    
+def measure_offset(frame, xstar_guess, ystar_guess, xoffset_guess, yoffset_guess, guessflux=1, rad=5, stampsize=10):
+    """
+    Helper function to compute the relative offset between stars.
+    
+    Args:
```
semaphoreP, line 36: you'll get build errors until you fill in the docstrings each argument needs to be documented, and the return function needs to be documented.

corgidrp/astrom.py
```diff
@@ -0,0 +1,387 @@
...
+            angle = - np.arctan(xdif/ydif)
+    else:
+        if ydif<0:
+            angle = - np.arctan(xdif/ydif) + np.pi
+        else: 
+            angle = (2*np.pi) - np.arctan(xdif/ydif)
+            
+    return angle * 180/np.pi
+
+def compute_northangle(image, source_guesses, center_coord, skycoords_):
```
semaphoreP, line 214: I think the north angle and platescale computation can be combined together, since they both use the same structure (just look at a different measurable). This should simplify and speed up the code

corgidrp/astrom.py
```diff
@@ -0,0 +1,387 @@
...
+
+    platescale = compute_platescale(image=image, source_guesses=guesses, skycoords_=guesses)
+
+    northangle = compute_northangle(image=image, source_guesses=guesses, center_coord=target_coordinate, skycoords_=guesses)
+
+    cal_properties = (platescale, northangle)
+    ra, dec = compute_boresight(image=image, target_coordinate=target_coordinate, cal_properties=cal_properties)
+
+    astrom_data = np.array([ra, dec, platescale, northangle])
+    astrom_cal = corgidrp.data.AstrometricCalibration(astrom_data, pri_hdr=dataset[0].pri_hdr, ext_hdr=dataset[0].ext_hdr)
```
semaphoreP, line 381: you can just return the single astrometriccalibration file (with the updated history). It's ok that it isn't a dataset object, since we want to output a single calibration file at the end.

corgidrp/data.py
```diff
@@ -708,11 +706,54 @@ def copy(self, copy_data = True):
...
+        pri_hdr (astropy.io.fits.Header): the primary header (required only if raw 2D data is passed in)
+        ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw 2D data is passed in)
+        
+    Attrs:
+        boresight (np.array): the [(RA, Dec)] of the center pixel in ([deg], [deg])
+        platescale (float): the platescale value in [mas/pixel]
+        northangle (float): the north angle value in [deg]
+
+    """
+    def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None):
```
semaphoreP, line 724: It should take another argument `input_dataset` that's the `input_dataset` that gets fed into `astrometric_calibration()`. We want to reference it to record all the files that were used to make this astrometric calibration file, for future documentation reference. (see other calibration classes like `Dark` and `NonLinearityCalibration`

corgidrp/mocks.py
```diff
@@ -202,3 +206,143 @@ def create_default_headers(obstype="SCI"):
     exthdr['MISSING'] = False
 
     return prihdr, exthdr
+
+def create_astrom_data(filedir=None, field_path='~/corgidrp/tests/test_data/JWST_CALFIELD2020.csv'):
```
semaphoreP, line 210: we shouldn't have a path that's hard coded that depends on the user's directory. I think the best thing to do is to require it as an input argument (instead of an optional keyword). Then, your test code can reconstruct the file path.

semaphoreP: You can use python path manipulation functions in the test script to navigate to the corgidrp/tests/test_data folder. e.g.,:
```
# reconstruct where the dark is saved
calibdir = os.path.join(os.path.dirname(__file__), "testcalib")
dark_filename = "sim_dark_calib.fits"
```

semaphoreP (commented):
Here's a few comments to get started, even though it looks like there's still some work in progress.

corgidrp/astrom.py
```diff
@@ -0,0 +1,431 @@
...
+
+    return image_center_RA, image_center_DEC
+
+def astrometric_calibration(input_dataset, guesses, target_coordinate):
+    """
+    Perform the boresight calibration of a dataset.
+    
+    Args:
+        input_dataset (corgidrp.data.Dataset): Dataset containing a single image for astrometric calibration
+        guesses (str): Path to file with x,y [pixel] locations of sources AND RA,DEC [deg] true source positions
```
semaphoreP, line 391: why do we need both `guesses` and `target coordinate`

corgidrp/astrom.py
```diff
@@ -0,0 +1,431 @@
...
+    # average all offsets in x,y directions
+    boresight_x, boresight_y = np.mean(boresights[:,0]), np.mean(boresights[:,1])
+
+    # convert back to RA, DEC
+    image_center_RA = target_coordinate[0] - ((boresight_x * cal_properties[0]) * astropy.units.mas).to(astropy.units.deg).value
+    image_center_DEC = target_coordinate[1] - ((boresight_y * cal_properties[0]) * astropy.units.mas).to(astropy.units.deg).value
+
+    return image_center_RA, image_center_DEC
+
+def astrometric_calibration(input_dataset, guesses, target_coordinate):
```
semaphoreP, line 385: maybe call this function boresight calibration if it is just doing the boresight calibration? We'll probably also have a function that does precise astrometric calibration).

corgidrp/mocks.py
```diff
@@ -356,3 +359,146 @@ def create_badpixelmap_files(filedir=None, col_bp=None, row_bp=None):
...
+
+        center = Table()
+        center['RA'] = [target[0]]
+        center['DEC'] = [target[1]]
+        ascii.write(center, filedir+'/target_guess.csv', overwrite=True)
+
+        frame.save(filedir=filedir, filename=filename)
+
+    frames.append(frame)
+    dataset = data.Dataset(frames)
```
semaphoreP, line 502: is this a single image dataset? do we want to have the option to have more frames simulated?

tests/test_astrom.py
```diff
@@ -0,0 +1,60 @@
...
+    # perform the astrometric calibration
+    astrom_cal = corgidrp.astrom.astrometric_calibration(input_dataset=dataset, guesses=guess_path, target_coordinate=target_path)
+    assert len(astrom_cal.data) == 4
+
+    # the data was generated to have the following image properties
+    expected_platescale = 21.8
+    expected_northangle = 45
+
+    # check orientation is correct within 0.3 [deg]
+    assert np.abs(expected_platescale - astrom_cal.platescale) < 0.5  # not sure how accurate this needs to be
```
semaphoreP, line 41: I guess this value gets folded into the boresight requirement. errors on the platescale will affect the ability to know that boresight is good to within 30 mas.

semaphoreP (commented):
Looks good. Just a few more minor questions/comments.

corgidrp/astrom.py
manduhmia: seemed easier than adding a chunk of code to find the center star in a field since the user would know the target coords and we are already passing in values. should I add code to find the target coordinate given all the guess coordinates?

tests/test_astrom.py
manduhmia: there was no formal requirement on plate scale accuracy that I could find in the literature (maybe for the reason it gets folded into boresight?) I can get rid of this assertion

semaphoreP: let's keep it for now, since it is useful for our own sanity.

corgidrp/astrom.py
semaphoreP: I guess maybe it depends on what kind of progress you've made on the automatic source ID code (sounds like the same task as the auto source ID code?)

manduhmia: my thought was more like: assume symmetry in source guesses and then find the guess thats closest to the min+max/2 in both axes

semaphoreP: I'm not sure how robust that is.. it would be nice to have some way to double/triple check we are right.

hsergi:
@manduhmia @semaphoreP: I've read thru the code and unit test. Looks good to me. Just wondering about the 72 arcsec radius to search for the target star. match_sources() has this hard coded parameter rad=0.02 # deg. Is it a good idea to have it hardcoded? How does it compare with the CBE of Roman’s observatory pointing (RA, DEC)? P.S. Same (hard coded) radius value is used in mocks.create_astrom_data()

manduhmia:
> @manduhmia @semaphoreP: I've read thru the code and unit test. Looks good to me. Just wondering about the 72 arcsec radius to search for the target star. match_sources() has this hard coded parameter rad=0.02 # deg. Is it a good idea to have it hardcoded? How does it compare with the CBE of Roman’s observatory pointing (RA, DEC)? P.S. Same (hard coded) radius value is used in mocks.create_astrom_data()

That radius value is kind of arbitrary, I just needed to set a large enough search radius to not 'miss' any stars in the source matching function-- I made it tunable so that the user can now decide how big that should be. I also made it tunable in the mocks.create_astrom_data(), although that one is just used to speed up computation in translating LMC sources to pixel space in the mock image.

corgidrp/astrom.py
```diff
@@ -0,0 +1,702 @@
...
+    return image_center_RA, image_center_DEC
+
+def boresight_calibration(input_dataset, field_path, threshold=100, sigma=3, mask_rad=1, comparison_threshold=25):
+    """
+    Perform the boresight calibration of a dataset.
+    
+    Args:
+        input_dataset (corgidrp.data.Dataset): Dataset containing a single image for astrometric calibration
+        field_path (str): Full path to directory with search field data (ra, dec, vmag, etc.)
+        threshold (int): Number of stars to find (default 25)
```
semaphoreP, line 669: isn't the default 100?

corgidrp/astrom.py
```diff
@@ -0,0 +1,702 @@
...
+        sigma (int): Stddev of point source psf (default: 3 pixels)
```
semaphoreP, line 670: Could we input the `fwhm` instead of `sigma`. We usually remember the value for FWHM more easily, and it is a more direct connection to the `mask_rad` keyword.

semaphoreP (commented):
Just a couple minor API comments.

semaphoreP:
@hsergi if you think the modifications look good to you, can you give your approval to the PR? If not, let us know if there's anything else.

tests/test_astrom.py
```diff
@@ -0,0 +1,54 @@
+import os
+import numpy as np
+import pytest
+import sys
+sys.path.insert(0, '/Users/macuser/Roman/corgidrp')
```
semaphoreP, line 5: We should remove this. You shouldn't need this after installating the corgidrp

corgidrp/astrom.py
```diff
@@ -0,0 +1,702 @@
...
+
+    return image_center_RA, image_center_DEC
+
+def boresight_calibration(input_dataset, field_path, threshold=100, fwhm=7, mask_rad=1, comparison_threshold=25):
+    """
+    Perform the boresight calibration of a dataset.
+    
+    Args:
+        input_dataset (corgidrp.data.Dataset): Dataset containing a single image for astrometric calibration
+        field_path (str): Full path to directory with search field data (ra, dec, vmag, etc.)
```
semaphoreP, line 668: Let's make this an optional argument, and by default, it uses the JWST_CALFIELD2020.csv in test_data. You can use os.path.dirname(__file__) to get from this file to the csv file using relative paths. 

Also, add JWST_CALFIEDL2020.csv to the MANIFEST.in in the base directory.

semaphoreP (commented):
Sorry, just a couple more things I noticed. Otherwise, let's get this PR'd soon, so that future development can go into a separate PR.

hsergi (approved):
Looks good to me! Great work!

Issue #109, opened 2024-06-19
corgidrp.data.KGain does not read in an input_dataset for archiving the filenames used to produce the calibration file


kjl0025:
The class could then include the call to `_record_parent_filenames()`.

kjl0025:
@JuergenSchreiber , just wanted to document this here before I forgot.  Please correct me if I'm misunderstanding something about the implementation.

JuergenSchreiber:
I did not add this since we only had one "pre-defined" value. But I can add it.

JuergenSchreiber:
@semaphoreP merged this into main, so it can be closed

PR #115, opened 2024-06-21
#109: input_dataset for KGain

JuergenSchreiber:
added parameter input_dataset for KGain, for the copy method I had to check whether the input_dataset were already applied.


kjl0025 (approved):
I am good with this.  I'll let @semaphoreP pull the trigger if he's good with it.

PR #116, opened 2024-06-21
Kgain/NL

hsergi:
Porting Kgain and NL calibration functions to CORGI DRP, together with unit tests. 

I have used CORGI DRP native data types, including the new class KGain. Original (internal) operations related to calibration are performed with numpy arrays. 

All unit tests are passed.


hsergi:
It is necessary to add FluxMap1024.fits add to tests/test_data for running the ut. It is a pupil image:

Flux1024_image

Zipped FITS file:

[Uploading FluxMap1024.fits.zip…]()

It is 8 Mb.

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+        array of size N x 2, where N is the number of bins set by the 'signal_bins_N' 
+        parameter in detector.kgain_params. The first column is the mean (DN) and 
+        the second column is standard deviation (DN) corrected for read noise.
+    
+    """
+    # copy stack_arr and stack_arr2 and cast them into np arrays for convenience
+    stack_arr, stack_arr2 = copy_and_cast(stack_arr, stack_arr2)
+
+    # input checks
+    from corgidrp.detector import kgain_params as master_files
```
kjl0025, line 106: I think it would be better to put this line at the very top.  (Side note:  The implementation of input parameters will be overhauled in Jason's PR #106 (away from loading them from detector.py), but I think this is fine as it is, and that PR would handle any changes needed.)

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    prhd, exthd = create_default_headers()
+    gain_value = np.array([[kgain]])
+    kgain = data.KGain(gain_value, pri_hdr = prhd, ext_hdr = exthd)
+    
+    return (kgain, mean_rn_gauss_e, mean_rn_std_e, ptc)
+    
+
+
+"""
+Class to hold input-checking functions to minimize repetition
```
kjl0025, line 696: FYI, I put this same class in `check.py` in a `util` folder and `test_check.py` in the `tests` folder on my PR since the calibration code in that branch also uses these checks.

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+    means_min_max : array-like
+        minima and maxima of mean values (in DN) used the fit each for EM gain. 
+        The size of means_min_max is N x 2, where N is the length of actual_gain_arr.
+    """
+    # copy stack_arr and stack_arr2 and cast them into np arrays for convenience
+    stack_arr, stack_arr2 = copy_and_cast(stack_arr, stack_arr2)    
+
+    # input checks
+    # load in default parameters
+    from corgidrp.detector import nonlin_params as master_files
```
kjl0025, line 146: I think this should be at the top.

kjl0025 (changes requested):
A few changes requested in the comments, along with a few clarifications requested from Jason/Max.

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+from scipy.optimize import curve_fit
+
+import corgidrp.data as data
+from corgidrp.data import Image
+from corgidrp.mocks import create_default_headers
+
+class CalKgainException(Exception):
+    """Exception class for calibrate_kgain."""
+
+def calibrate_kgain(stack_arr, stack_arr2, emgain, min_val, max_val,
```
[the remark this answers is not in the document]
kjl0025: Oh, and one other thing:  An error for the k gain could be obtained and put into an `err` of the `KGain` class instantiation that makes the calibration product.  That would require more than a mere port of this code, but I don't think it wouldn't be too hard to implement.  I didn't know how precise we wanted to be with that.  Perhaps that could be made a GitHub issue as a possible enhancement?

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+import io
+import matplotlib.pyplot as plt
+from statsmodels.nonparametric.smoothers_lowess import lowess
+
+from corgidrp.calibrate_kgain import check, copy_and_cast
+
+class CalNonlinException(Exception):
+    """Exception class for calibrate_nonlin."""
+
+def calibrate_nonlin(stack_arr, exp_time_stack_arr, time_stack_arr, len_list,
```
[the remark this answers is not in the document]
kjl0025: Oh, and one other thing:  An error for the k gain could be obtained and put into an `err` of the `NonLinearityCalibration` class instantiation that makes the calibration product.  That would require more than a mere port of this code, but I don't think it wouldn't be too hard to implement.  I didn't know how precise we wanted to be with that.  Perhaps that could be made a GitHub issue as a possible enhancement?

corgidrp/calibrate_kgain.py
kjl0025: Update: Since the new implementation of the DetectorParams class has been pulled in, I think these parameters should go in there.  Then you would have a class instance of that as an input to your function to get these k gain parameters.

maxwellmb: Are some (or all) of the parameters just better as input kwargs? I haven't thought through the pros/cons, but I thought I'd throw it out there for discussion.

kjl0025: I guess that could work if the walker can handle kwargs.  I just suggested the move to DetectorParams so that all input parameters are in one place.  On that note, I was going through my list of things to do, and one of them was to suggest putting all the calibration function and step function input parameter arguments in DetectorParams so that they would all be in one place, but if you and Jason prefer to keep parameters as kwargs associated with their relevant functions (and if the walker can handle that), I suppose that's fine.  As long as the CTC can set the kwargs and DetectorParams as needed when this is in Docker, I am fine either way.

maxwellmb: This conversation has been moved over to Issue #126 for a more general discussion (thanks @kjl0025 for starting it there!).

JuergenSchreiber:
> It is necessary to add FluxMap1024.fits add to tests/test_data for running the ut. It is a pupil image:
> 
> Flux1024_image
> Zipped FITS file:
> 
> Uploading FluxMap1024.fits.zip…
> 
> It is 8 Mb.
Is it really necessary to upload a new file to the test_data or could also the existing test data like example_L1_input.fits be used?

corgidrp/calibrate_kgain.py
JuergenSchreiber: default of the boolean mkplot should be False. Do we want matplotlib in the requirement.txt?

hsergi:
The file has been added (see Jason's changes). This particular convo is closed.

tests/test_kgain_cal.py
```diff
@@ -0,0 +1,847 @@
...
+stack_arr = np.squeeze(np.stack(stack_list))
+
+# set input parameters for calibrate_kgain function
+min_val = 800
+max_val = 3000
+binwidth = 68
+
+################### define class for tests ######################
+
+class TestCalibrateKgain(unittest.TestCase):
```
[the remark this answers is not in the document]
maxwellmb: It would be good to reformat these tests to be a bit more readable and make it clearer where the main functionality is being tested (see some of the other existing tests for examples.

corgidrp/calibrate_kgain.py
hsergi: I think @maxwellmb prefers plots to be stored. Right now, default is None, which is probably fine. Should we leave None as default in the calibration functions? And let the ut do the plots and store them? If so, "Do we want matplotlib in the requirement.txt?"

maxwellmb (changes requested):
Alright, thanks all for your patience on this. Please see below for my comments. A quick summary:

- The two main calibration functions should be reformatted to accept a single Dataset as input. This will require the function to sort the files into the previous stack_arr, (and stack_arr2 where appropriate). This should mostly just be header keyword sorting.
- On the output side of things, let's output the calibration Datatype that we want with all the auxiliary data stored as header keywords or in extensions. 
- For producing the plots: we'll have it be default not show them (boolean flag), but by default  save them (another boolean flag) to a directory that should be a kwarg
- The kgain_params and nonlin_params should be mostly kwargs to the calibration functions and if we think they'll be constants we can make kgain_params and nonlin_params in their respective modules (calibrate_kgain.py and calibrate_nonlin.py). 
- I think (but I'm not 100% sure) that nonlin_tableTVAC.txt might be able to replace the existing [nonlin_sample.csv](https://github.com/roman-corgi/corgidrp/blob/main/tests/test_data/nonlin_sample.csv) and associated fits file. This is a nice-to-have not a full requirement in order to keep the number of test_data files as small as possible. 
- The tests should be reformatted into pytests and be a bit more organized so that it's easier to follow, in particular the main functionality testing. 
- There's some coordination to be done with @kjl0025 and me about some of the functions in detector.py. Whoever starts working on them first please let everyone else know so as to not double the work. 

We've previously discussed splitting this work between @JuergenSchreiber and @hsergi . I suggest @hsergi take on one of the functions (let's say KGain) and @JuergenSchreiber do the other (Non-linearity). Since there may be overlapping functions the only way to really make this work will be staying in touch regularly (ideally here, not slack) and with regular commits. I think in this case the conflicting time zones might help us. Please let me know if you have any questions about all of this or if this plan doesn't work for you.

corgidrp/calibrate_kgain.py
hsergi: I sent an email to Guillermo Gonzalez (tellus1) who wrote these functions, and CC'ed Max.

corgidrp/calibrate_kgain.py
hsergi: @maxwellm: There are two "check" py functions: calibration and tests. It seems that we should have both. 

@kjl0025: which is your PR (link)? I'd like to find the util foder with check.py and test_check.py

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: > Yes, I think it would be good to have the output be a KGain class (ideally with error, but I don't think that's a hard requirement at this point. If it's a pain to put in here we can make it a feature request for late.

@maxwellmb and @kjl0025 : the return has been a KGain class [link](https://github.com/roman-corgi/corgidrp/blob/e19f15ce1e537df724f133446591c964fa432715/corgidrp/calibrate_kgain.py#L689). Can you explain what's missing?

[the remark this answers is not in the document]
hsergi: > For the input, it should be just a single Dataset. Internal to this function we can do the sorting as need be. The EMGain should be in the header of Dataset objects, so I don't think we need `emgain` as an input parameter. `min_val` and `max_val` should be set as keyword arguments.

calibrate_kgain (and calibrate_nonlin) have stacks (of stacks) of arrays [docstring](https://github.com/roman-corgi/corgidrp/blob/e19f15ce1e537df724f133446591c964fa432715/corgidrp/calibrate_kgain.py#L16) Initially, @kjl0025 and I thought to build stacks (of stcks) of Datasets. @maxwellmb: can you explain what you mean by having a single Dataset instead?

corgidrp/calibrate_kgain.py
hsergi: Found it: https://github.com/roman-corgi/corgidrp/pull/105/files

maxwellmb: I'm not quite sure I understand why we need two different check.py files

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
maxwellmb: I think just get rid of the other outputs and put them into headers/extensions should be fine.

corgidrp/calibrate_kgain.py
hsergi: Added matplotlib to requirements.txt and order packages alphabetically

corgidrp/calibrate_nonlin.py
hsergi: @maxwellmb @semaphoreP: what is the syntax for docstrings (Args) for calibration functions in corgidrp accepting **kwargs? (I only found walker using it and there's no docstring associated with that particular use)

corgidrp/calibrate_kgain.py
hsergi: > Yes, let's add matplotlib to the requirement.txt!
> 
> Looking at the `kgain_params`, I don't know exactly what all of them are, but here's my guess as to where different ones should go:
> 
> * All of the row and column ones _might_ be constants, in which case they could stay in a `kgain_params` dictionary, but I'd move the location of that dictionary to this file rather than detector.py since I don't think they're used for anything else. They might also just be fine as kwargs here.
> * The rest of them are probably best as kwargs, since we'll likely want to be able to change them around with real data.

@maxwellmb When using kwargs, do you mean to just type their keys/values in the unit tests directly?

corgidrp/calibrate_nonlin.py
maxwellmb: you'd want the function to look _something_ like: 
`def calibrate_kgain(input_dataset, logplot1=-1, logplot2=4, logplot3=200)`
(as a very basic example) where I'm grabbing `logplot1` from the kgain_params.

hsergi:
> * The two main calibration functions should be reformatted to accept a single Dataset as input. This will require the function to sort the files into the previous stack_arr, (and stack_arr2 where appropriate). This should mostly just be header keyword sorting.

@maxwellmb Is there any specific example? I have some difficulty understanding this point. In IIT the variable was a stack of stacks of nd arrays and initially Kevin and I did a stack of stacks of Dataset (see [make_frame](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L59) and [stack](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L161)). What's non-desirable with this option?

maxwellmb:
> > * The two main calibration functions should be reformatted to accept a single Dataset as input. This will require the function to sort the files into the previous stack_arr, (and stack_arr2 where appropriate). This should mostly just be header keyword sorting.
> 
> @maxwellmb Is there any specific example? I have some difficulty understanding this point. In IIT the variable was a stack of stacks of nd arrays and initially Kevin and I did a stack of stacks of Dataset (see [make_frame](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L59) and [stack](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L161)). What's undesirable with this option?

We don't yet have an example of this because these calibrations functions are the first time we're encountering this. It's undesirable because it requires the definition and maintenance of a new type of data class, when we'll ultimately have to have the code to sort the files somewhere anyway (so we might as well put it in here). It also doesn't conform to the standard dataset-in -> dataset-out function structure, which makes it harder (impossible?) to write recipes for.

hsergi:
> > > * The two main calibration functions should be reformatted to accept a single Dataset as input. This will require the function to sort the files into the previous stack_arr, (and stack_arr2 where appropriate). This should mostly just be header keyword sorting.
> > 
> > 
> > @maxwellmb Is there any specific example? I have some difficulty understanding this point. In IIT the variable was a stack of stacks of nd arrays and initially Kevin and I did a stack of stacks of Dataset (see [make_frame](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L59) and [stack](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L161)). What's undesirable with this option?
> 
> We don't yet have an example of this because these calibrations functions are the first time we're encountering this. It's undesirable because it requires the definition and maintenance of a new type of data class, when we'll ultimately have to have the code to sort the files somewhere anyway (so we might as well put it in here). It also doesn't conform to the standard dataset-in -> dataset-out function structure, which makes it harder (impossible?) to write recipes for.

I would appreciate if someone can provide a first example: either @maxwellmb , @kjl0025 or someone more versed than I in corgidrp Dataset uses

corgidrp/calibrate_kgain.py
hsergi: I think the answer is in the next thread

corgidrp/calibrate_nonlin.py
hsergi: @maxwellmb: Okay, when saying "be kwargs", you mean explictly writing any keyword in the definition, not using **kwargs: `def calibrate_kgain(input_dataset, **kwargs)`?

maxwellmb: yeah! :D

hsergi: ty

hsergi: @JuergenSchreiber: FYI, I sent an email to Guillermo Gonzalez and Nathan Bush, CC'ed Max, asking which of each of the kgain and nonlin params are constant or variable.

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: @maxwellmb header/extensions of which variable?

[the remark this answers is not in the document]
hsergi: Also, if the only output from calibrate_kgain and calibrate_nonlin is either a KGain or NonLinearity object, respectively: how are these keywords/extensions output? The input variables are copied as per corgidrp best practices. Therefore, changes inside the code will not be accessible after the function is run, unless they are returned.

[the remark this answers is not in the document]
maxwellmb: I just meant that you should put those outputs into the KGain object header, and if need be, feel free to add an extension for arrays (like the PTC).

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    
+    binwidth : int
+        Width of each bin for calculating std devs and means from each 
+        sub-stack in stack_arr. Maximum value of binwidth is 800. NOTE: small 
+        values increase computation time.
+        (minimum 10; binwidth between 65 and 75 recommended)
+    
+    mkplot : boolean
+        Option to display plots. Default is None. If mkplot is anything other 
+        than None, then this option is chosen.
```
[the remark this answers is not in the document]
hsergi: @maxwellmb if the output directory can be freely defined and someone defines it within corgidrp repo, it will be tracked by git and produce annoying 'files changed'. Is this a possibility that you have considered?

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+        
+    verbose : boolean
+        Option to display various diagnostic print messages. Default is None. 
+        If mkplot is anything other than None, then this option is chosen.
+    
+    Returns
+    -------
```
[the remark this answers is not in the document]
hsergi: I'd appreciate a definite prescription by people who have an overview on these calibrations. I can't decide myself where it is appropriate to include all the additional output results. @kjl0025 ?

hsergi:
@semaphoreP @maxwellmb "The tests should be reformatted into pytests and be a bit more organized so that it's easier to follow, in particular the main functionality testing." Any specific example in the pipeline that ahs been recently merged that can be used as a guidance?

corgidrp/calibrate_kgain.py
maxwellmb: Yep!

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
maxwellmb: I just meant that you should put those outputs into the KGain object header, and where appropriate (e.g. for an array of values) add an extension HDU for arrays, like the PTC (including useful headerwords about the content of the extension, and a name, per the developer guidelines).

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
maxwellmb: I hadn't considered that, so feel free to direct it somewhere else. This will normally be set by the walker when we pick an output directory for files.

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
maxwellmb: Put `read_noise_gauss` and `read_noise_stdev` into the KGain header, and put ptc into an HDU extension of `KGain`. See for example, the bias attribute of the `Image` class.

maxwellmb:
> @semaphoreP @maxwellmb "The tests should be reformatted into pytests and be a bit more organized so that it's easier to follow, in particular the main functionality testing." Any specific example in the pipeline that ahs been recently merged that can be used as a guidance?

You could look to basically all the exists tests for guidance. In terms of changing over unittests, perhaps @kjl0025 has gotten started on doing this for his PR?

maxwellmb:
> > > > * The two main calibration functions should be reformatted to accept a single Dataset as input. This will require the function to sort the files into the previous stack_arr, (and stack_arr2 where appropriate). This should mostly just be header keyword sorting.
> > > 
> > > 
> > > @maxwellmb Is there any specific example? I have some difficulty understanding this point. In IIT the variable was a stack of stacks of nd arrays and initially Kevin and I did a stack of stacks of Dataset (see [make_frame](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L59) and [stack](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L161)). What's undesirable with this option?
> > 
> > 
> > We don't yet have an example of this because these calibrations functions are the first time we're encountering this. It's undesirable because it requires the definition and maintenance of a new type of data class, when we'll ultimately have to have the code to sort the files somewhere anyway (so we might as well put it in here). It also doesn't conform to the standard dataset-in -> dataset-out function structure, which makes it harder (impossible?) to write recipes for.
> 
> I would appreciate if someone can provide a first example: either @maxwellmb , @kjl0025 or someone more versed than I in corgidrp Dataset uses

I'm not sure this will actually work, but here's something to give you inspiration. 
Let's say you needed to sort your files by exposure time: 

#Grab the exposure times
exposure_times = [frame[0].header['EXPTIME'] for frame in dataset.frames]
#Find the unique ones
unique_exposure_times = np.unique(exposure_times)
#Make a list of data 
stacked_array = [dataset.all_data[exposure_times == exp_time] for exp_time in unique_exposure_times]

You can use different logic to split up the dataset object how you please, but this is an example of how you might do that.

hsergi:
> > > > > * The two main calibration functions should be reformatted to accept a single Dataset as input. This will require the function to sort the files into the previous stack_arr, (and stack_arr2 where appropriate). This should mostly just be header keyword sorting.
> > > > 
> > > > 
> > > > @maxwellmb Is there any specific example? I have some difficulty understanding this point. In IIT the variable was a stack of stacks of nd arrays and initially Kevin and I did a stack of stacks of Dataset (see [make_frame](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L59) and [stack](https://github.com/roman-corgi/corgidrp/blob/b0396fe2c441be96864a4902c4f6cdebbfed6d26/tests/test_kgain_cal.py#L161)). What's undesirable with this option?
> > > 
> > > 
> > > We don't yet have an example of this because these calibrations functions are the first time we're encountering this. It's undesirable because it requires the definition and maintenance of a new type of data class, when we'll ultimately have to have the code to sort the files somewhere anyway (so we might as well put it in here). It also doesn't conform to the standard dataset-in -> dataset-out function structure, which makes it harder (impossible?) to write recipes for.
> > 
> > 
> > I would appreciate if someone can provide a first example: either @maxwellmb , @kjl0025 or someone more versed than I in corgidrp Dataset uses
> 
> I'm not sure this will actually work, but here's something to give you inspiration. Let's say you needed to sort your files by exposure time:
> 
> ```
> #Grab the exposure times
> exposure_times = [frame[0].header['EXPTIME'] for frame in dataset.frames]
> #Find the unique ones
> unique_exposure_times = np.unique(exposure_times)
> #Make a list of data 
> stacked_array = [dataset.all_data[exposure_times == exp_time] for exp_time in unique_exposure_times]
> ```
> 
> You can use different logic to split up the dataset object how you please, but this is an example of how you might do that.

@semaphoreP: thanks for volunteering in today's meeting to provide a template code tat can be used to deal with stack_array and stack_array2 in both Kgain and Nonlinearity, different to the approach I took of defining a stack of (stack of) Dataset Images.

tests/test_nonlin_cal.py
```diff
@@ -0,0 +1,434 @@
...
+    min_value = 800.0
+    max_value = 10000.0
+    f_nonlin = np.polyval(coeffs, DN)
+    # Control values outside the min/max range
+    f_nonlin = np.where(DN < min_value, np.polyval(coeffs, min_value), f_nonlin)
+    f_nonlin = np.where(DN > max_value, np.polyval(coeffs, max_value), f_nonlin)
+    
+    return f_nonlin
+
+def make_frame(f_map, bias, kgain, rn, emgain, time, coeffs, nonlin_flag):
```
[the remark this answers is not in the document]
hsergi: Working on it

corgidrp/calibrate_nonlin.py
hsergi: Guillermo's reply for both kgain and nonlin parameters: "the only parameters that might need adjusted are the 'rms_low_limit' and 'rms_upp_limit' for nonlin, which were determined with simulated frames."

hsergi: @JuergenSchreiber @maxwellmb After reviewing the list with Guillermo in more detail, these are the parameters that may be keywords in the Kgain and nonlin functions:  **'logplot1', 'logplot2', ‘logplot3’, lowess_frac’, 'rms_low_limit' and 'rms_upp_limit'.**

I'll ask again Guillermo on Thu 07/11 to confirm if there are any changes to this list.

tests/test_nonlin_cal.py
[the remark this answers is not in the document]
hsergi: Done

hsergi:
"The kgain_params and nonlin_params should be mostly kwargs to the calibration functions and if we think they'll be constants we can make kgain_params and nonlin_params in their respective modules (calibrate_kgain.py and calibrate_nonlin.py)." 

@maxwellmb, @JuergenSchreiber  This is done, tested and committed for both kgain and nonlinearity

JuergenSchreiber:
I have updated the docstrings to our flake conventions and converted the unittests to pytests.
Tests of kgain are not running since there is a function used in calibrate_kgain which has no import: count_contiguous_repeats(exp_time_loop)

corgidrp/calibrate_nonlin.py
hsergi: Guillermo confirms as of 07/11/24 that these are the parameters in Kgain and Nonlin that may vary along the mission: **'logplot1', 'logplot2', ‘logplot3’, ‘lowess_frac’, 'rms_low_limit', 'rms_upp_limit’, 'fit_upp_cutoff1', 'fit_upp_cutoff2', 'fit_low_cutoff1' and 'fit_low_cutoff2'.**

maxwellmb: Ok great. These look like appropriate keyword arguments (but maybe with clearer names).

corgidrp/detector.py
```diff
@@ -265,3 +265,129 @@ def get_rowreadtime_sec(datetime=None):
...
+    colroi2: ROI constant
+    rn_bins1: read noise bins range limit 1
+    rn_bins2: read noise bins range limit 2
+    max_DN_val: maximum DN value to be included in PTC
+    signal_bins_N: number of bins in the signal variables
+    log_plot1: log plot min value
+    log_plot2: log plot max value
+    log_plot3: log plot number of elements
+"""
+kgain_params= {
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/detector.py
```diff
@@ -265,3 +265,129 @@ def get_rowreadtime_sec(datetime=None):
...
+    min_mask_factor: factor for making mask
+    lowess_frac: factor to use in lowess smoothing function
+    rms_low_limit: rms relative error selection limit
+    rms_upp_limit: rm relative error selection limit
+    fit_upp_cutoff1: linear fit upper cutoff
+    fit_upp_cutoff2: linear fit upper cutoff
+    fit_low_cutoff1: linear fit upper cutoff
+    fit_low_cutoff2: linear fit upper cutoff
+"""
+nonlin_params= {
```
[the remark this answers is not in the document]
hsergi: Done

hsergi:
@mygouf, @maxwellmb , @semaphoreP: Unit tests are running (not passing all, but that's fine right now). I'd appreciate if @JuergenSchreiber can deal with the output of calibrate_kgain.py, and Jason write the Dataset object for stack_array and stack_array2 for both calibration functions: calibrate_kgain and calibrate_nonlin (bear in mind, as detailed by Max in some of the comments above, that there's a bunch of info, like exposure times that will be retrieved from that Dataset). I think I can deal with the rest of the tasks

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+        
+        good_mean_frame += frame  # Accumulate into good_mean_frame
+    
+    good_mean_frame = good_mean_frame / nFrames2
+    
+    # Slice the good_mean_frame array
+    frame_slice = good_mean_frame[rowroi, colroi]
+    
+    # If requested, create a figure and plot the sliced frame
+    if mkplot is not None:
```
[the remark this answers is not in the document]
hsergi: done

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    # Use the mask to filter out NaN values
+    y_err = logged_deviations_error[non_nan_mask]
+    x_vals = logged_averages[non_nan_mask]
+    y_vals = logged_deviations[non_nan_mask]
+    
+    # make array of kgain values from variance and mean values
+    # equation: k_gain = mean / signal variance
+    kgain_arr = 10**(x_vals)/(10**y_vals)**2
+    
+    if mkplot is not None:
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+        plt.title('kgain versus mean')
+        plt.show()
+    
+    kgain_clipped, _ = sigma_clip(kgain_arr)
+    mode_kgain,_ = calculate_mode(kgain_clipped)
+
+    # adopt 'mode_kgain' as the final value to return
+    kgain = mode_kgain
+
+    if mkplot is not None:
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+    
+    min_write : float
+        Minimum mean value in DN to output in nonlin_arr and csv_lines. 
+        (800.0 recommended)
+    
+    max_write : float
+        Maximum mean value in DN to output in nonlin_arr and csv_lines. 
+        (10000.0 recommended)
+    
+    mkplot : boolean
```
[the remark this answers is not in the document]
hsergi: Done

tests/test_kgain_cal.py
```diff
@@ -0,0 +1,847 @@
...
+import numpy as np
+from astropy.io import fits
+
+from corgidrp.data import Image, Dataset
+from corgidrp.mocks import create_default_headers
+from corgidrp.calibrate_kgain import (calibrate_kgain, CalKgainException, check)
+
+######################## function definitions ###############################
+
+def nonlin_coefs(filename,EMgain,order):
```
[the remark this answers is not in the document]
hsergi: Done

semaphoreP:
While I wrote the dataset helper function in the most recent PR (#131), I think it's most efficient if Sergi or Juergen implements the dataset stacks here. You can look at the test function I wrote in PR #131 for how to call it, but you guys probably are much more familiar what stacks you need to break it up in this code.

tests/test_kgain_cal.py
```diff
@@ -0,0 +1,847 @@
...
+    min_value = 800.0
+    max_value = 10000.0
+    f_nonlin = np.polyval(coeffs, DN)
+    # Control values outside the min/max range
+    f_nonlin = np.where(DN < min_value, np.polyval(coeffs, min_value), f_nonlin)
+    f_nonlin = np.where(DN > max_value, np.polyval(coeffs, max_value), f_nonlin)
+    
+    return f_nonlin
+
+def make_frame(f_map, bias, kgain, rn, emgain, time, coeffs, nonlin_flag):
```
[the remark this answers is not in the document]
hsergi: Done. New function name is make_fluxmap_frame, to leave room for other similar names related with other functions in corgidrp (make_frame is too generic)

tests/test_kgain_cal.py
```diff
@@ -0,0 +1,847 @@
...
+
+"""
+Class to hold input-checking functions to minimize repetition
+Note: This module, used by test_kgain_cal.py and test_nonlin.py is included
+here because at this moment it is not clear if the functions in the module are
+of general utility for corgidrp
+"""
+
+"""Unit tests for check.py."""
+class ut_check:
```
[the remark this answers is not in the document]
hsergi: Done

tests/test_kgain_cal.py
```diff
@@ -0,0 +1,847 @@
...
+    # shift so that function passes through unity at 3000 DN for these tests
+    fitVals0 = np.polyval(coeffs, DNs)
+    ind = np.where(DNs == 3000)
+    unity_val = fitVals0[ind][0]
+    coeffs[3] = coeffs[3] - (unity_val-1.0)
+    fitVals = np.polyval(coeffs,DNs)
+    
+    return coeffs, DNs, fitVals
+
+def nonlin_factor(coeffs,DN):
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    check.positive_scalar_integer(max_val, 'max_val', TypeError)
+    if min_val >= max_val:
+        raise CalKgainException('max_val must be greater than min_val')
+    check.positive_scalar_integer(binwidth, 'binwidth', TypeError)
+    if binwidth < 10:
+        raise CalKgainException('binwidth must be >= 10.')
+    if binwidth > 800:
+        raise CalKgainException('binwidth must be < 800.')
+    
+    ################### function defs #####################
```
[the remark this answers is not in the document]
hsergi: @maxwellmb can you clarify for me what you mean? (these functions are in the file, defined before calibrate_kgain. I do not think we want to add sub-functions inside calibrate_kgain. Adding them to mock.py or elsewhere is not a good idea since they are only used here.)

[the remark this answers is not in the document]
maxwellmb: they're currently defined in the `calibrate_kgain` function which makes the function just overall long and harder to read/follow. I was just suggesting that they get defined earlier in the file begfore the `calibrate_kgain` function. This is mostly just to help with code readability.

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    if not isinstance(master_files['logplot3'], (float, int)):
+        raise TypeError('logplot3 is not a number')
+    
+    # check parameters
+    if type(stack_arr) != np.ndarray:
+        raise TypeError('stack_arr must be an ndarray.')
+    if np.ndim(stack_arr) != 4:
+        raise CalKgainException('stack_arr must be 4-D (i.e., a stack of '
+                '3-D sub-stacks)')
+    if len(stack_arr) <= 10:
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: @maxwellmb Oh! That's how it is now -and might have been for a couple of days before I read this conversation-. I'm going to close the convo, unless you want to change it.

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+        is_increasing = np.all(ctim_datetime[:-1] <= ctim_datetime[1:])
+        if not is_increasing:
+            raise CalNonlinException('Elements of time_stack_arr must be '
+                    'in increasing time order for each EM gain value.')
+    if type(stack_arr2) != np.ndarray:
+        raise TypeError('stack_arr2 must be an ndarray.')
+    if np.ndim(stack_arr2) != 3:
+        raise CalNonlinException('stack_arr2 must be 3-D (i.e., a stack of '
+                '2-D sub-stacks')
+    if len(stack_arr2) < 30:
```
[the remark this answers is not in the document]
hsergi: Done

tests/test_nonlin_cal.py
```diff
@@ -0,0 +1,434 @@
...
+ctime_strings_plus_one_day1 = ctime_datetime_plus_one_day1.strftime('%Y-%m-%dT%H:%M:%S').tolist()
+time_stack_arr1[index:index+len_list0[3]] = ctime_strings_plus_one_day1
+
+# set input parameters for calibrate_nonlin
+local_path = os.path.dirname(os.path.realpath(__file__))
+norm_val = 3000
+min_write = 800
+max_write = 10000
+
+class TestCalibrateNonlin(unittest.TestCase):
```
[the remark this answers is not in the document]
hsergi: Done. Although we may need to open new conversations when reviewing the unit tests for Kgain and Non-linearity. Add/remove some given all the I/O/variable changes

hsergi:
@kjl0025 @maxwellmb @semaphoreP: The input data for both calibrations (KGain and Non-Linearity) have a set of parameters: bias,kgain,rn,emgain,exp_time (I think they are all self explanatory). When generating mock data, I think we want to store these values in the header, right?. 

Looking at mocks.py, the default header has a few keys. For instance, 'CMDGAIN': is it the same as EMGAIN? We always use the latter, so I'm not sure why would we not use EMGAIN if they were both the same.

Surely someone else is already using some header keywords for all or some of Bias, Read Noise, KGain, ... Is there a document/file that keeps track of newly defined keywords, and even the ones in the default case?

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: Working on the *input* datasets (stack_arr and stack_arr2)

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    
+    stack_arr2 : array-like
+        The stack of EXCAM illuminated pupil L1 SCI frames (counts in DN). 
+        stack_arr2 contains a stack of at least 30 frames of uniform exposure 
+        time, such that the net mean counts in the pupil region is a few thousand 
+        DN (2000 to 4000 DN recommended; note: unity em gain is recommended 
+        when k-gain is the primary desired product, since it is known more 
+        accurately than non-unity values. stack_arr and stack_arr2 must be 
+        obtained under the same positioning of the pupil relative to the 
+        detector.
```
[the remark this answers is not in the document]
hsergi: stack_arr2 is used to build a mean frame that is then subtracted from the data in stack_arr. stack_arr2 is made of different exposure times, and always has EM=1. So, both stack_arr and stack_arr2 are independent. 

I have sent an email to Guillermo asking whether this is the case, or stack_arr2 could somehow be a subset of stack_arr (assuming one always has some frames with EM=1 with the same exposure time, enough to build a stack_arr2, N>= 30)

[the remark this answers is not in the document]
hsergi: @maxwellmb: Confirmed with Guillermo: `stack_arr2` cannot be just a subset of `stack_arr`. It usually has more frames with EM=1 than the unity gain frames in `stack_arr`. As I mention before, `stack_arr2` is used to builda mean frame that is then subtracted from `stack_arr` frames.

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    log_plot2 = constants_config['logplot2']
+    log_plot3 = constants_config['logplot3']
+
+    if mkplot is not None:
+        # Avoid issues with importing matplotlib on headless servers without GUI
+        # support without proper configuration
+        import matplotlib.pyplot as plt
+    
+    # Prescan region
+    offset_rowroi = slice(offset_rowroi1,offset_rowroi2)
```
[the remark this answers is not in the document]
kjl0025: Oh, let me correct one part of the above comment:  The offset_* parameters are used for getting the prescan region for read noise measurement.  So we could keep those parameters in play for now and potentially change to using `detector_areas` dictionary and `slice_section` from detector.py in the future.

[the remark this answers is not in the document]
maxwellmb: @kjl0025 this is a really good catch. If there are steps that are already happening in other step functions we shouldn't repeat them here! Indeed the prescan bias subtraction is the first step function. I believe we expect all data to automatically be processed to L2b. Should the data input to this function be L2b?

hsergi:
@maxwellmb I have to implement the single object return in calibrate_nonlin.py. I asked a couple of questions on slack. There's many comments so far in this PR. I'd like to confirm the steps for the implementation:

calibrate_nonlin.py now returns 4 variables: 

     **headers** (np.array): 1-D array of headers used to build csv-lines. The length is equal to
        the number of columns in 'nonlin_arr' and is one greater than the
        length of 'actual_gain_arr'.
      **nonlin_arr** (np.array): 2-D array with nonlinearity values for input signal level (DN) in rows
        and EM gain values in columns. The input signal in DN is the first column.
        Signal values start with min_write and run through max_write in steps
        of 20 DN.
      **csv_lines** (list): List of strings containing the contents of 'headers' and 'nonlin_arr'.
      **means_min_max** (np.array): minima and maxima of mean values (in DN) used the fit each for EM gain. The size of means_min_max is N x 2, where N is the length of actual_gain_arr.

Is the output ``nonlin_arr``, ``nonlin_raw`` in the NonLinearity class?

Should I add ``headers``, ``csv_lines`` and ``means_min_max`` as new data arrays

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: Done. Kgain value and error are returned as part of a single object

corgidrp/calibrate_kgain.py
hsergi: test_check.py is the unit test of check.py.

corgidrp/data.py
```diff
@@ -558,6 +558,84 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
         if 'DATATYPE' not in self.ext_hdr or self.ext_hdr['DATATYPE'] != 'NonLinearityCalibration':
             raise ValueError("File that was loaded was not a NonLinearityCalibration file.")
 
+class KGain(Image):
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+        values increase computation time.
+        (minimum 10; binwidth between 65 and 75 recommended)
+    
+    mkplot : boolean
+        Option to display plots. Default is None. If mkplot is anything other 
+        than None, then this option is chosen.
+        
+    verbose : boolean
+        Option to display various diagnostic print messages. Default is None. 
+        If mkplot is anything other than None, then this option is chosen.
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_kgain.py
hsergi: Done

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,1099 @@
...
+    
+    # Gaussian read noise value in DN
+    mean_rn_gauss_DN = np.mean(rn_gauss)
+    mean_rn_gauss_e = mean_rn_gauss_DN * kgain
+    
+    mean_rn_std_DN = np.mean(read_noise)
+    mean_rn_std_e = mean_rn_std_DN * kgain
+    
+    # If requested, plotting
+    if mkplot is not None:
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+    stack_arr : array-like
+        stack_arr is a stack of frames of dimention 3, which is implicitly 
+        subdivided into smaller ranges of grouped frames. The frames are EXCAM 
+        illuminated pupil L1 SCI frames. There must be one or more unique EM 
+        gain values and at least 20 unique exposure times for each EM gain. The 
+        number of frames for each EM gain can vary. The size of stack_arr is: 
+        Sum(N_t[g]) x 1200 x 2200, where N_t[g] is the number of frames having 
+        EM gain value g, and the sum is over g.
+    
+    exp_time_stack_arr : array-like
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+        exp_time_stack_arr is an array of dimension 1 of exposure times (in s) 
+        corresponding to the frames in stack_arr in the order found there. The 
+        length of exp_time_stack_arr must equal the number of frames used to 
+        construct stack_arr. There must be at least 20 unique exposure times at 
+        each EM gain. The values must be greater than 0.
+    
+    time_stack_arr : array-like
```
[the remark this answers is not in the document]
hsergi: Done

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+        must equal to the number of sub-stacks in stack_arr. The number of 
+        elements (= the number of unique em gain values) in len_list must be 
+        one or greater.
+    
+    stack_arr2 : array-like
+        stack_arr2 is a stack array of EXCAM unity EM gain illuminated pupil L1 
+        SCI frames. stack_arr2 contains a stack of frames of uniform exp time, 
+        such that the mean signal in the pupil regions is a few thousand DN.
+    
+    actual_gain_arr : array-like
```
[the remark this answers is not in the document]
hsergi: Added to the list of future work on these functions.

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,806 @@
...
+                                 'max_write.')
+    check.real_nonnegative_scalar(rms_low_limit, 'rms_low_limit', TypeError)
+    check.real_nonnegative_scalar(rms_upp_limit, 'rms_upp_limit', TypeError)
+    if rms_low_limit >= rms_upp_limit:
+        raise CalNonlinException('rms_upp_limit must be greater than rms_low_limit')
+    
+    ######################### start of main code #############################
+    
+    # Define offset columns ROI range
+    offset_colroi = slice(offset_colroi1,offset_colroi2)
```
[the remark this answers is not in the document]
hsergi: Added to the list of future work on these functions.

tests/test_kgain_cal.py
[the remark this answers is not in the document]
hsergi: Done

hsergi:
Summary of the PR above (I may have missed something)

**To-do**

1/ Max’s comment in the PR: I think (but I'm not 100% sure) that nonlin_tableTVAC.txt might be able to replace the existing [nonlin_sample.csv](https://github.com/roman-corgi/corgidrp/blob/main/tests/test_data/nonlin_sample.csv) and associated fits file. This is a nice-to-have not a full requirement in order to keep the number of test_data files as small as possible: https://github.com/roman-corgi/corgidrp/pull/116#pullrequestreview-2163979641:

2/ Sergi. Return single object for Non-linearity. Confirmation/questions for Max: https://github.com/roman-corgi/corgidrp/pull/116#issuecomment-2243517984

3/ Sergi. Check plots/results

**Future work?**

1/ ``actual_gain_array``: https://github.com/roman-corgi/corgidrp/pull/116#discussion_r1669132879:

2/ ROI regions: A future PR (or this one possibly) could implement the use of detector_areas and slice_section, tools in detector.py for that instead of relying on the offset_roi parameters. (Kevin): https://github.com/roman-corgi/corgidrp/pull/116#discussion_r1674701793

3/ (Kgain error is now part of Kgain. Not sure this is necessary any longer)An error for the k gain could be obtained and put into an err of the NonLinearityCalibration class instantiation that makes the calibration product. That would require more than a mere port of this code, but I don't think it wouldn't be too hard to implement. I didn't know how precise we wanted to be with that. Perhaps that could be made a GitHub issue as a possible enhancement?: https://github.com/roman-corgi/corgidrp/pull/116#discussion_r1651599405

4/ Add readout noise estimation to KGain: https://github.com/roman-corgi/corgidrp/issues/135

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,918 @@
...
+        plt.title('kgain versus mean')
+        plt.savefig(f'{plot_outdir}/{fname}')
+        if verbose:
+            print(f'Figure {fname} stored in {plot_outdir}')
+        if show_plot:
+            plt.show()
+        plt.close()
+    
+    kgain_clipped, _ = sigma_clip(kgain_arr)
+    mode_kgain,_ = calculate_mode(kgain_clipped)
```
kjl0025, line 715: I think the `err` for the KGain class should be `np.std(kgain_clipped)` and not what it currently is in line 812.

kjl0025: Also, I can't remember if we discussed this or not already, but I think the read noise (and its uncertainty would be nice, too) should be saved as header info in the `KGain` instance that is the output of this function.  We could probably discuss this at the meeting tomorrow.

semaphoreP:
Hi Sergi, as mentioned on slack, the input to NonLinearityCalibration is the `np.vstack((header,nonlin_arr))`

corgidrp/calibrate_nonlin.py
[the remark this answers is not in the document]
hsergi: Issue opened to deal with this is a future release: https://github.com/roman-corgi/corgidrp/issues/136

corgidrp/calibrate_nonlin.py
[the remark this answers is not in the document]
hsergi: See https://github.com/roman-corgi/corgidrp/issues/137

corgidrp/calibrate_kgain.py
hsergi: Thanks @kjl0025! First point done. UT run and OK. As I shared with you `err` was before your suggestion `mean_rn_gauss_DN` = np.mean(rn_gauss), where `rn_gauss = [Single_peakfit(bin_locs,counts_diff[x])/np.sqrt(2) for x in range(len(frames_diff))]` and `Single_peakfit' is just std(): "Fit Gauss function to x, y data. Returns mean and sigma. Only sigma
    is used in main code call"

Second one: https://github.com/roman-corgi/corgidrp/issues/135

hsergi (commented):
I thin all comments have been addressed and the one related to the test_data file nonlin_tableTVAC.txt included as an Issue for future work: https://github.com/roman-corgi/corgidrp/issues/138

hsergi (commented):
I think I have addressed all comments and moved ROI, read noise, actual array of EM gains to Issues (Feature/Enhacement) for future work.

corgidrp/calibrate_kgain.py
kjl0025: Sorry, I don't understand (concerning the first point).  I see that `err` in line 812 is the read noise estimate you just mentioned, but that is not the standard deviation of the array `kgain_clipped`.

kjl0025: I see that was just changed in the latest commit.  Great, I'll resolve this conversation.

hsergi:
New comment from @maxwellmb: "one useful thing to do for both K-gain and NL would be to define in the docstring the detailed info about the expectation on the input data: For Example - What previous steps have been run? Is it L1, L2a, or is it L2a, but with a skipped step. You could take a look at Kevin’s Docstring for the NoiseMaps."

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,924 @@
...
+    max_DN = max_DN_val; # maximum DN value to be included in PTC
+    nrow = len(cal_list[0][0])
+    ncol = len(cal_list[0][0][0])
+    
+    # prepare "good mean frame"
+    good_mean_frame = np.zeros((nrow, ncol))
+    nFrames2 = len(mean_frame_list)
+    for mean_frame_count in range(nFrames2):
+        frame = mean_frame_list[mean_frame_count]
+        frame = frameProc(frame, colroi_fp, actual_gain)
```
maxwellmb, line 505: The `frameProc ` function is superseded by the `prescan_biassub` step function in l1_to_l2a.py and the `em_gain_division` step function in `l2a_to_l2b.py`. I think it would be fair to make it a requirement that the input dataset has already been processed by these steps.

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,960 @@
...
+    frame = frame/emgain
+    return frame
+
+def calibrate_nonlin(dataset_nl, actual_gain_arr, actual_gain_mean_frame,
+                     n_cal=20, n_mean=30, norm_val = 2500, min_write = 800.0, max_write = 10000.0,
+                     lowess_frac = 0.1, rms_low_limit = 0.004, rms_upp_limit = 0.2,
+                     pfit_upp_cutoff1 = -2, pfit_upp_cutoff2 = -3,
+                     pfit_low_cutoff1 = 2, pfit_low_cutoff2 = 1,
+                     make_plot=True, plot_outdir='figures', show_plot=False,
+                     verbose=False):
```
maxwellmb, line 160: Same comment as for kgain: the gains we can get from the headers and don't need to be input arguments here. For now `CMDGAIN` is appropriate since we haven't implemented actual gain.

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,960 @@
...
+    good_mean_frame = np.zeros((nrow, ncol))
+    nFrames = len(mean_frame_arr)
+
+    good_mean_frame = good_mean_frame / nFrames
+    
+    mean_frame_index = 0
+    # Loop over the mean_frame_arr frames
+    for i in range(nFrames):
+        frame = mean_frame_arr[i]
+        frame = frameProc(frame.astype(np.float64), offset_colroi,
```
maxwellmb, line 422: same comment as in Kgain. frameProc has been superseded by `prescan_biassub` and `em_gain_division`. I think it would be fair to make it a requirement that the input dataset has already been processed by these steps.

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,924 @@
...
+        if show_plot:
+            plt.show()
+        plt.close()
+    
+    # prepare PTC array
+    ptc_list = [compiled_binned_averages, 
+             compiled_binned_shot_deviations]
+    ptc = np.column_stack(ptc_list)
+
+    prhd, exthd = create_default_headers()
```
maxwellmb, line 815: You should grab a set of headers from the input data and add a history comment instead of filling the KGain object with mock headers.

corgidrp/calibrate_nonlin.py
```diff
@@ -0,0 +1,960 @@
...
+    # See data.NonLinearityCalibration doc string for more details:
+    # [0, 1:]: Gain axis values
+    # [1:, 0]: "count" axis value
+    actual_gain_arr = np.insert(actual_gain_arr, 0, np.nan)
+    n_col = len(nonlin_arr3) + 1
+    n_row = len(actual_gain_arr)
+    nonlin_data=np.insert(nonlin_arr3, 0, actual_gain_arr).reshape(n_col,n_row)
+    
+    # Return NonLinearity instance
+    prhd, exthd = create_default_headers()
```
maxwellmb, line 875: Same comment as KGain: we shouldn't fill in dummy headers, we should be using real headers from the input data.

tests/test_kgain_cal.py
```diff
@@ -0,0 +1,220 @@
...
+import test_check
+from corgidrp import check
+from corgidrp.data import Image, Dataset
+from corgidrp.mocks import (create_default_headers, make_fluxmap_image)
+from corgidrp.calibrate_kgain import (calibrate_kgain, CalKgainException, kgain_params)
+
+######################## function definitions ###############################
+
+def nonlin_coefs(filename,EMgain,order):
+    # filename is the name of the csv text file containing the TVAC nonlin table
```
maxwellmb, line 26: I think this function is already defined in mocks.py, right? If its just the same function, then you can just import it here.

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,924 @@
...
+            print(f'Figure {fname} stored in {plot_outdir}')
+        if show_plot:
+            plt.show()
+        plt.close()
+    
+    # parameter for plot
+    parm1 = -0.5*np.log10(kgain)
+    
+    # Gaussian read noise value in DN
+    mean_rn_gauss_DN = np.mean(rn_gauss)
```
maxwellmb, line 743: Is anything done with these? I feel like we discussed putting them into the ext_hdr as a keyword. Both the values and the errors.

maxwellmb (commented):
I think we're looking good with just a few last comments! They're mostly about the inputs to the step functions and frame preprocessing: a) we should only have a dataset or a calibration datatype as arguments, b) frameProc can be deleted and you can make it a requirement that the input data has been processed by the appropriate step functions. See e.g. the docstring for `calibrate_darks_lsq` in `darks.py`.

corgidrp/calibrate_kgain.py
```diff
@@ -0,0 +1,924 @@
...
+        
+    return clipped_data, mask
+    
+######################### start of main code #############################
+
+def calibrate_kgain(dataset_kgain, actual_gain, actual_gain_mean_frame,
+                    n_cal=10, n_mean=30, min_val=800, max_val=3000, binwidth=68,
+                    make_plot=True,plot_outdir='figures', show_plot=False,
+                    logspace_start=-1, logspace_stop=4, logspace_num=200, verbose=False):
+    """
```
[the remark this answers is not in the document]
hsergi: @maxwellmb: Are you suggesting to use the actual gain values in CMDGAIN? That could be an issue. First, there are checks for it to be 1 in Kgain, and measured values need not be exactly 1.0. More importantly, in non-linearity the values of CMDGAIN are used to split the dataset. This only works if there's a set with the same value (commanded). The actual values may be different by fractional values and the split would not work as expected.

Can we not just create a new Keyowrd, as I had to to with the mean frame, and update the Issue https://github.com/roman-corgi/corgidrp/issues/136 accordingly? And add this keyword, together with the one of the mean frame to the list of desired keywords? They can be renamed at any time and scripts be updated.

corgidrp/calibrate_kgain.py
hsergi: Last thing I heard is that it could be mentioned in the doc strings, as it is now. I'm fine changing it to not allow it and expect it to be processed earlier on.

corgidrp/calibrate_kgain.py
hsergi: This was implemented by Juergen. I thought it had been reviewed. Sounds good.

corgidrp/calibrate_kgain.py
hsergi: @maxwellmb: Read noise is a future task according to the latest convo on slack and the last telecon. https://github.com/roman-corgi/corgidrp/issues/135 

 Do you mean to include it now in this release? This implies changing KGain class accordingly. Is the read noise estimated elsewhere besides in calibrate_kgain? @kjl0025 I would like to avoid two different sources for the same type of noise, unless there's a good reason and some documentation on what to do with both estimations. If it is only in K-Gain, it's easier.

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
maxwellmb: Sorry if that was confusing.  We should definitely not change the CMDGAIN, but you should have your code use CMDGAIN for now since we don't have any actual gain keywords or infrastructure implemented. I'd prefer this to adding an extra keyword for now. However if you could comment on the issues where this is relevant so that we can keep track of places where we need to change it.

corgidrp/calibrate_kgain.py
maxwellmb: Yes, if we had to I suppose we could just have this function accept L1 data. It ends up bypassing much of the header tracking of frames, which is not ideal and means that if we ever change anything about either the EMGAIN or the pre-scan bias routines we'd have to change them here too, so it's really not ideal from a code maintenance point of view. I'm not sure I understand the advantage of doing it this way.

corgidrp/calibrate_kgain.py
maxwellmb: Ok, good point - we shouldn't have two sources of read noise. My first instinct is that it should be where the PTC is defined, but I'm certainly open to other thoughts. @kjl0025 do you re-derive it yourself in your routine? 

It would be incredibly simple to just add these variables to the header at this point, isn't that what we talked about? I could be forgetting things.

tests/test_kgain_cal.py
hsergi: Good catch! I had added it to mocks.py from calibrate_nonlin.py. Done

corgidrp/calibrate_kgain.py
hsergi: wrt the latter: my understanding from slack was yes and then no. I'm fine if it is yes again. The only caveat is what I raised above.

corgidrp/calibrate_kgain.py
hsergi: I'm not advocating for it. I was just paraphrasing the decision on the meeting. I'll implement the idea that bias and EM gain division are already applied. Maybe @kjl0025 has some objection/suggestion not to do it?

corgidrp/calibrate_kgain.py
maxwellmb: as far as I can tell the read noise is not saved when creating the NoiseMaps

JuergenSchreiber: if it is only a new ext_hdr keyword, you do not need to change the KGain class. just
kgain.ext_hdr["RNOISE"] = mean_rn_gauss_DN

kjl0025: Read noise is calculated simply as a check in my noise maps code, but it isn't supposed to be the official calibration product for read noise, and as Max said, it isn't returned (in the DRP version of the code).  The way it is calculated in this k gain code is the official (more accurate) way to do it.  The official return for the read noise was called `mean_rn_gauss_e` originally (and perhaps still retains that name in this DRP version; I just haven't checked it lately).  Yeah, I thought we discussed putting that in a header.  Like Juergen said, we could do "RNOISE", and another header could be "RN_ERR".  Or to have them more similar and both <= 8 characters (to avoid HIERARCH cards), we could do maybe "RN" and "RN_ERR"?  So "RN" should be `mean_rn_gauss_e`, and the error could be an estimate of the standard deviation error (fractional error goes like 1/sqrt(2N)?  I'll have to check that later).

corgidrp/calibrate_kgain.py
semaphoreP: I agree with Max here. We should defer these steps to the other step functions we wrote.

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
kjl0025: Yes, I think anywhere arrays are sorted according to EM gain, CMDGAIN is all that is needed.  Anywhere an actual EM gain value is utilized (divided by, for example), a more accurate EM gain if available should be used.  And the good thing is that a commanded gain of 1.0 should be physically an applied gain of 1 (i.e., no gain applied), by definition, regardless of the parametrization of the EM gain curve, and that's because no voltage was applied to the HV electrode in that case.

corgidrp/calibrate_kgain.py
kjl0025: I agree that the frames should be assumed to be already processed as much as possible before going through this function.  In the case of calibrate_kgain and calibrate_nonlin, the frames are bias-subtracted, and in addition, if EM gain is >1 for the input data for calibrate_kgain, EM gain division is also needed.  So the bias subtraction that happens here (and in the nonlin code) needs to be removed, along with the EM gain division in this function.  Consequently, we don't need to worry about changing the bias subtraction implementation here to use `slice_section`, and thus we also don't need offset_rowroi1, offset_rowroi2, offset_colroi1, or offset_colroi2. And the doc string should be changed accordingly to specify that these pipeline steps should have already been taken for the input data.

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: Agreement (Max) is to follow what Kevin says/did: "eah, the CMDGAIN is supposed to be the actual applied gain, assuming a good fit to the parametrization function.  In the future, we may implement a methodology of reading off the EM gain from a given frame and assign some new header like ESTGAIN or EMGAIN_M for that read-off value.  But since that is still a little up in the air, and since CMDGAIN in practice should be close enough to the applied gain, he wanted us to just use CMDGAIN for now.
If you really wanted to, you could do something like I did in my calibrate_darks_lsq() function with a try and except statement so that even if EMGAIN_M is never implemented, the code doesn't fail:
```
try: # if EM gain measured directly from frame TODO change hdr name if necessary
     EMgain_arr = np.append(EMgain_arr, datasets[i].frames[0].ext_hdr['EMGAIN_M'])
except: # use commanded gain otherwise TODO change hdr name if necessary
     EMgain_arr = np.append(EMgain_arr, datasets[i].frames[0].ext_hdr['CMDGAIN']):"
```

corgidrp/calibrate_kgain.py
hsergi: The input dataset has a set of frames, each has its own primary/extended header. The operations with the dataset are:

`dataset_nl.add_error_term(                dataset_nl.frames
dataset_nl.all_data                       dataset_nl.rescale_error(
dataset_nl.all_dq                         dataset_nl.save(
dataset_nl.all_err                        dataset_nl.split_dataset(
dataset_nl.copy(                          dataset_nl.update_after_processing_step(`

I do not see an option to collect all the information of the headers in a structured way. Is there one? That'd be great. If not, which object structure should the call to 
```
data.NonLinearityCalibration(nonlin_data,
        pri_hdr = prhd, ext_hdr = exthd, input_dataset=dataset_nl)
```
have when the dataset has a set of frames with different primary/extended headers? An example would help

maxwellmb: you could just take the headers from the first frame: 
`prhd = nonlin_data.frames[0].pri_hdr`
`exthd = nonlin_data.frames[0].ext_hdr`

corgidrp/calibrate_kgain.py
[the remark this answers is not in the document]
hsergi: Done.

corgidrp/calibrate_nonlin.py
hsergi: Done

corgidrp/calibrate_kgain.py
hsergi: @maxwellmb If they are different for different frames as I wrote, how does this propagate each difference?

maxwellmb: Yeah, that's one of the downsides to this approach, but its better than having just dummy headers. Nobody should be using the GAIN keyword from a Non Linearity calibration anyway. Ultimately the files used will be kept track of (because we keep track of the input datasets), so if we ever need to trace it back its possible. The HISTORY keyword should be detailed enough to make it clear that the current fits file is a combination of several fits files.

corgidrp/calibrate_kgain.py
hsergi: @kjl0025: "don't need offset_rowroi1, offset_rowroi2, offset_colroi1, or offset_colroi2." These are used in calibrate_kgain.py for other functions than the bais subtraction. Similarly for calibrate_nonlin.py. I think I have to keep them and remove frameProc, change doc strings and inout bias subtracted (EM divided, in Kgain) in the unit tests

corgidrp/calibrate_kgain.py
hsergi: We could remove the useless headers too

corgidrp/calibrate_kgain.py
hsergi: I have committed the changes above, except removing "ffset_rowroi1, offset_rowroi2, offset_colroi1, or offset_colroi2" for the reasons I mention. If there's something else to do about them, let me know. Otherwise, we can resolve this converstaion.

corgidrp/calibrate_nonlin.py
hsergi: Done. Please check for any further suggestions

corgidrp/calibrate_kgain.py
hsergi: Done. Please check for any further suggestions

corgidrp/calibrate_nonlin.py
hsergi: Waiting to hear about the changes done in calibrate_kgain in this respect

corgidrp/calibrate_nonlin.py
maxwellmb: Looks good!

corgidrp/calibrate_kgain.py
maxwellmb: Sure, deleting non-sense headers is definitely a nice-to-have to catch those subtle bugs (but naturally, low priority).

corgidrp/calibrate_nonlin.py
maxwellmb: Sure this looks fine.

corgidrp/calibrate_kgain.py
hsergi: Done. I added two keywords as suggested above. Build and tests pass.

corgidrp/calibrate_kgain.py
kjl0025: Oh, true.  It looks like the offset_* parameters are only used for specifying the prescan region (other than bias subtraction via frameProc).  We can specify the prescan portion of a frame instead using `slice_section`.  But as discussed before, if that is being put off to a future pull request, then you'll need to keep the offset_* parameters for now.  I don't think that change would be hard to do.  Anywhere you see a frame region is specified as `frame[offset_rowroi, offset_colroi]`, you could instead use `prescan = slice_section(frame, 'SCI', 'prescan_reliable', detector_regions)` where `detector_regions` is as found in `prescan_biassub()`.

hsergi: Done.

Thanks so much everyone and kudos for @kjl0025 and @maxwellmb for all the extensive details! 

Please, let me know if you have any further suggestions!

maxwellmb (approved):
I think we're finally here with this one! Huge thanks to @hsergi for all the hard work getting this through and to @JuergenSchreiber for the extra help, and to @kjl0025 for the extra support and discussions.

maxwellmb:
@kjl0025 I think you need to approve here too since at some point you "Requested changes"

kjl0025 (approved):
I committed two small changes, despite what the diff appears to be!  I apologize for that; I thought I had made the random whitespace changes that my editor does go away, but apparently not.  I will make sure this doesn't happen again.  I promise that the only changes I actually made were these two small changes:
-I added a period at the end of a sentence in the nonlin doc string.
-I changed the err for read noise to incorporate the error in k gain since multiplication by k gain is involved:
```
    # rn err depends on spread of data that determines rn and the error in kgain,
    # so use error propagation to find error in (rn in DN)*kgain
    kgain_err = np.std(kgain_clipped)
    rn_err_DN = np.std(rn_gauss)
    rn_err_e = np.sqrt((kgain*rn_err_DN)**2 + (mean_rn_gauss_DN*kgain_err)**2)
    exthd['RN_ERR'] = rn_err_e
```

Everything else looks good, and I will let Jason or Max pull the trigger to pull it in.

kjl0025:
And the nonlin code currently does not have an err output yet, but I believe Sergi made an issue for that for a future pull request (and it is also covered in Issue #124).

PR #119, opened 2024-06-25
adding templates for bug reports, feature requests, and pull requests

dsavransky:
Adding standardized templates for bug reports, feature requests, and pull requests.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,7 +8,6 @@ Please delete options that are not relevant (and this line).
 - Bug fix (non-breaking change which fixes an issue)
 - New feature (non-breaking change which adds functionality)
 - Breaking change (fix or feature that would cause existing functionality to not work as expected)
-- This change requires a documentation update
 
 ## Reference any relevant issues (don't forget the #)
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,19 @@
+## Describe your changes
+
+
+## Type of change
+
+Please delete options that are not relevant (and this line).
+
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- This change requires a documentation update
+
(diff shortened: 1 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+## Type of change
+
+Please delete options that are not relevant (and this line).
+
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- This change requires a documentation update
+
+## Reference any relevant issues (don't forget the #)
```
semaphoreP, line 13: do we need this? I think you can reference issues as part of the "Development" sidebar (although doing so does close the issue upon merge of PR).

.github/pull_request_template.md
```diff
@@ -0,0 +1,19 @@
...
+
+
+## Type of change
+
+Please delete options that are not relevant (and this line).
+
+- Bug fix (non-breaking change which fixes an issue)
+- New feature (non-breaking change which adds functionality)
+- Breaking change (fix or feature that would cause existing functionality to not work as expected)
+- This change requires a documentation update
```
semaphoreP, line 11: I don't understand this particular kind of change. shouldn't the documentation update be part of the PR

semaphoreP (commented):
Had a couple of minor questions on the PR one. otherwise, looks good. Thanks for putting this together.

.github/pull_request_template.md
dsavransky: I think I got this from a default pull request template a while ago.  It means that the pull pops up in the history of an issue even before its accepted/merged, possibly avoiding duplicative effort.  We can remove it, if you prefer.

.github/pull_request_template.md
dsavransky: I'll remove that option.

.github/pull_request_template.md
semaphoreP: Let's try it for now, and if it feels too redundant, we can remove.

semaphoreP (approved):

PR #120, opened 2024-06-28
Cosmic ray update

kjl0025:
Updated code to match that of latest CTC cosmic masking.  (Ell ported it just before the final update, which we weren't expecting at the time.)


kjl0025:
This branch was made from my branch from my other PR,  #105 .

PR #122, opened 2024-07-01
frame selection tests

### LINKED ISSUE #68: We need a function that can do frame selection

semaphoreP:
## Describe your changes

Adds the functionality to the frame selection step function

## Type of change

Closes #68 

- [x] I have linted my code


corgidrp/l2a_to_l2b.py
```diff
@@ -91,23 +91,55 @@ def flat_division(input_dataset, flat_field):
...
     Returns:
         corgidrp.data.Dataset: a version of the input dataset with only the frames we want to use
     """
-    return input_dataset.copy()
+    pruned_dataset = input_dataset.copy()
+    select_flags = np.ones(len(input_dataset))
+
+    for i, frame in enumerate(input_dataset.frames):
+        if bpix_frac < 1:
+            numbadpix = np.size(np.where(frame.dq > 0)[0])
```
maxwellmb, line 113: This logic might include pixels that we're comfortable including, e.g. cosmic rays that have been corrected for. I'm not quite sure how to handle it, but maybe we just make this  a future enhancement issue for now. You could imagine e.g. counting the number of dq pixels here that have a specific set of bit values....just thinking out loud.

corgidrp/l2a_to_l2b.py
```diff
@@ -91,23 +91,55 @@ def flat_division(input_dataset, flat_field):
...
+        raise ValueError("No good frames were selected. Unable to continue")
+
+    pruned_frames = pruned_dataset.frames[good_frames]
+    pruned_dataset = data.Dataset(pruned_frames)
+
+    # history message of which frames were removed.
+    history_msg = "Removed {0} frames:".format(np.size(bad_frames))
+    for bad_frame in input_dataset.frames[bad_frames]:
+        history_msg += " {0},".format(bad_frame.filename)
+    history_msg = history_msg[:-1] # remove last comma or :
```
maxwellmb, line 138: According to the FDD not only do we need to report the frames we rejected, but also the criteria.

maxwellmb: I think we can probably just the kwargs as header items.

maxwellmb (commented):
Overall this is looking good, I think the only real thing to update is that we need to track the frame rejection criteria, which I think we can either do with header keywords or in the header history.

corgidrp/l2a_to_l2b.py
semaphoreP: we could have as input a custom bpvalue mask, that species which DQ flags are allowed. Or, we can just hard code it to ignore the DQ = 2 flag for fixed cosmic rays.

maxwellmb: Yeah, inputting a custom bpvalue mask was basically what I was thinking. I think it would just have to be a single integer. We're currently using DQ=2 to also flag pixels replaced in the Correct bad Pixel step too, so I don't love that option as much.

semaphoreP: Ok, added new argument to pass in allowed DQ values that the function will ignore

corgidrp/l2a_to_l2b.py
semaphoreP: Added reason for each file to be removed (a file could have multiple reasons as well)

maxwellmb (approved):

PR #131, opened 2024-07-11
add dataset splitting feature

semaphoreP:
## Describe your changes

Add a helper function to split dataset into a list of sub datasets based on header keywords. Take a look at the test to see how it gets used

Minor: also moved the hash test to the same test file to be better categorized. 

## Type of change

- [x] I have linted my code

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -4,6 +4,7 @@ import numpy.ma as ma
 import corgidrp
 import astropy.io.fits as fits
 import astropy.time as time
+import pandas as pd
 
 class Dataset():
     """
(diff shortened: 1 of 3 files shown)


corgidrp/data.py
```diff
@@ -184,6 +185,57 @@ def rescale_error(self, input_error, err_name):
...
+            list of tuples: list of each set of unique header keywords. pri_hdr keywords occur before ext_hdr keywords
+        """
+        if prihdr_keywords is None and exthdr_keywords is None:
+            raise ValueError("No prihdr or exthdr keywords passed in to split dataset")
+        
+        col_names = []
+        col_vals = []
+        if prihdr_keywords is not None:
+            for key in prihdr_keywords:
+                dataset_vals = [frame.pri_hdr[key] for frame in self.frames]
```
maxwellmb, line 208: what if the key is present in some but not all files? is that too much of an edge case to care about? we might want to be error tolerant to a key not existing though.

semaphoreP: It'll throw an error here that the key doesn't exist. I think that's what we want since that means we can't split the dataset on this entry because some of the data is missing it. Likely points to a problem with the data.

kjl0025 (approved):
This all looks good to me!

Issue #112, opened 2024-06-21
Calibration Algorithm Generation: CTI Pump Trap Data (II&T Port)


maxwellmb:
This is a break-up from issue #20 which currently includes the step function and the calibration generation. This part just covers the generation of the calibrations. 

We need to port over the II&T code. I've started and there's a branch [cti_correction](https://github.com/roman-corgi/corgidrp/tree/cti_calibration). This is a requirement for R3.0.1, even if the step function becomes a "should" (i.e. not a strict requirement).

maxwellmb:
@kjl0025 says: 
"There are a few minor changes to the tpumpanalysis code that were made after the port to cgi_iit_drp repo.  Those changes are on the cosmic_update_code branch [here](https://github.com/roman-corgi/cgi_iit_drp/tree/cosmic_update_code/Calibration_NTR/cal/tpumpanalysis). The changes are found in example_script.py, tpump_final.py, and ut_tpump_final.py."

maxwellmb:
I'm dropping my current working to-do list here mostly to keep track of myself, but also in case other people get involved: 
- [ ] make the input a dataset object. We need to split the data by temperature (header keyword 'EXCAMT' <span style="color:blue">TBC this is the correct keyword</span>.
) and scheme (header keyword 'TPSCHEME')
- [ ] figure out where we get the tpump phase time from. Should it be a L1 header requirement? 
- [ ] Identify what other information that is output by `tpump_analysis` should be saved in headers/extensions
- [ ] Write tests
- [ ] Update slice_section functions based on what is done in the NoiseMaps and Kgain/NL branches

maxwellmb:
Tagging the following requirement numbers from the FDD for my reference: 1090879 and 1090880. 

The input data needs to be "clean focal plane images". From what I can tell, "Clean" means L2 data. @mygouf (and maybe @kjl0025), do either you know exactly what steps should/shouldn't be run for the trap pumping data. I'd like to make it clear  in the headers what the input expectation is.

maxwellmb:
Some other general questions/thoughts: 
- is illumination_correction needed if we do a flatfield correction?  
- where will the num_pumps input come from? It is the "Number of cycles of pumping performed for each trap-pumped frame."). Will it be in the CPGS file or the header? 
- It looks like the phase time was previously assumed to be in the 'PHASE_T' header keyword. Are we happy with this keyword. It's fine for me. We need it added to the header list.

kjl0025:
[...]

PR #134, opened 2024-07-19
A proposal for how to add more flexible hdus without hardcoding attributes

maxwellmb:
## Describe your changes

I've added an `hdu_list` attribute to the `Image` class that allows us to flexibly add new HDUs to an Image without having to hardcode in attributes, such as how we are currently treating `Image.bias`. In fact, this PR demonstrates the functionality by completely replacing `Image.bias`. Part of the motivation for this came up in a conversation with @neilzim about future spectroscopic dataset. There is a new `Image.add_extension_hdu()` function. 

Another advantage of this implementation is that means that if we decide at some point we no longer need to carry around some information then we can drop HDUs, e.g. when saving data to a new level. For example, do level 4 data products need to drag around the pre-scan bias information?

## Type of change

If we decide to go this route it will affect several current PRs, such as PR #116, PR #105, as it will provide a convenient way to attach new HDUs (e.g. the PTC for the KGain class). Its also relevant for issue #112. 

- [x] I have linted my code

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -265,7 +265,7 @@ class Image():
         filedir (str): the file directory on disk where this image is to be/already saved.
         filepath (str): full path to the file on disk (if it exists)
     """
-    def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq = None, bias=None, err_hdr = None, dq_hdr = None, bias_hdr=None):
+    def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq = None, err_hdr = None, dq_hdr = None, input_hdulist = None:
         if isinstance(data_or_filepath, str):
             # a filepath is passed in
             with fits.open(data_or_filepath) as hdulist:
@@ -278,7 +278,7 @@ class Image():
                 self.data = first_hdu.data
 
                 #A list of extensions
-                hdu_names = [hdu.name for hdu in hdulist]
+                self.hdu_names = [hdu.name for hdu in hdulist]
 
                 # we assume that if the err and dq array is given as parameter they supersede eventual err and dq extensions
                 if err is not None:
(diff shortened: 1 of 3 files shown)


corgidrp/data.py
```diff
@@ -353,33 +357,39 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq
             else:
                 self.dq = np.zeros(self.data.shape, dtype = int)
 
-            if bias is not None:
-                if (np.shape(self.data)[0],) != np.shape(bias):
-                    raise ValueError("The shape of bias is {0} while we are expecting shape {1}".format(bias.shape, self.data.shape))
-                self.bias = bias.astype(np.float32)
-            else:
-                self.bias = np.zeros(self.data.shape[0], dtype = np.float32)
+            #The default hdu extensions
```
semaphoreP, line 360: Just to clarify, this is to remake the hdu_names since the hudlist passed in may not have ERR, DQ?

corgidrp/data.py
```diff
@@ -246,34 +246,38 @@ class Image():
         pri_hdr (astropy.io.fits.Header): the primary header (required only if raw 2D data is passed in)
         ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw 2D data is passed in)
         err (np.array): 2-D/3-D uncertainty data
-        dq (np.array): 2-D data quality, 0: good. Other values track different causes for bad pixels and other pixel-level effects in accordance with the DRP implementation document.
-        bias (np.array): 1-D bias data
+        dq (np.array): 2-D data quality, 0: good. Other values track different causes for bad pixels and other pixel-level effects in accordance with the DRP implementation document.x
         err_hdr (astropy.io.fits.Header): the error extension header
         dq_hdr (astropy.io.fits.Header): the data quality extension header
-        bias_hdr (astropy.io.fits.Header): the bias extension header
+        hdu_list (astropy.io.fits.HDUList): an astropy HDUList object that contains any other extension types.
```
semaphoreP, line 252: should explicitly say this excludes everything except the data, err, and dq?

semaphoreP (approved):
I like the idea. It would be good to clarify that hdu_list is an list of hdus that don't include the image, err, and dq maps.

PR #151, opened 2024-07-29
creation of onsky flatfield

ramya-anche:
Creates flat fields using HST images of Neptune and Uranus.

1. 36 Raster scanned images are generated using the median combined HST image frames
2. The images are cropped and combined to create a matched filter in groups of 3.
3. The raster images are divided by the matched filter and combined to create a flatfield.


semaphoreP:
Hi Ramya, can you fix the conflicts with the main branch, so that the tests can run? Let me know if you need any help with conflict fixing!

corgidrp/detector.py
```diff
@@ -520,7 +528,204 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+    percens_resel[nrr>165] = None
+    full_qe[nrr>165]= None
+    full_qe=np.pad(full_qe, ((302,302),(302,302)), mode='constant',constant_values=(np.nan,))
+    full_residuals=np.pad(full_residuals, ((302,302),(302,302)), mode='constant',constant_values=(np.nan,))
+    percens_resel=np.pad(percens_resel, ((302,302),(302,302)), mode='constant',constant_values=(np.nan,))
+    
+    return (full_qe,full_residuals,percens_resel)
+    
+    
+def create_onsky_flatfield(filedir):
```
semaphoreP, line 632: This looks like the "step function" to make the flat field. The input of this function must be a `Dataset` object. It cannot be a file directory (Assume that the fits files have already be read in as a Dataset).

corgidrp/detector.py
```diff
@@ -520,7 +528,204 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
         np.array: lowest full well capacity saturation threshold for frames
     """
-    possible_sat_fwcs_arr = np.append((emgain_arr * fwcpp_arr)[:,np.newaxis], fwcem_arr[:,np.newaxis],axis=1)
-    sat_fwcs = sat_thresh * np.min(possible_sat_fwcs_arr,axis=1)
+	possible_sat_fwcs_arr = np.append((emgain_arr * fwcpp_arr)[:,np.newaxis], fwcem_arr[:,np.newaxis],axis=1)
+	sat_fwcs = sat_thresh * np.min(possible_sat_fwcs_arr,axis=1)
+
+	return sat_fwcs
+
+def residuals(images,planims=None):
```
semaphoreP, line 536: Can we rename this to a more distinctive name? maybe `flatfield_residuals`

corgidrp/detector.py
```diff
@@ -520,7 +528,204 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+    
+    
+def create_onsky_flatfield(filedir):
+	"""Turn this dataset of raster scanned image frames of neptune or uranus into on-sky calibrated flat field 
+
+    	Args:
+        	filedir (str):  Full path to directory of images that are raster scanned
+
+    	Returns:
+    		nept_band1_flatfield (corgidrp.data.FlatField): a master flat for flat calibration using on sky images of neptune in band 1
```
semaphoreP, line 639: The output of this function should be a single `data.FlatField` object. The flat fields of Uranus and Neptune in different bands should be produced through different recipes/function calls to this function.

semaphoreP (changes requested):
Hi Ramya, while you are still figuring out the tests, I started taking a look at the code. I'm just starting very high level right now, but we need to change the API of `create_onsky_flatfield()`. It should really look like (and actually just replace) the `create_flatfield()` function that's at the top of this file. Happy to discuss over slack if this is confusing to you.

maxwellmb:
Note: I had to add the appropriate git-lfs hooks into the CI settings, as previously it was not doing a git-lfs pull to get the new fits files. It was causing the build to fail here, and not before, since the test files introduced here are the first ones that actually triggered the git-lfs system.

corgidrp/detector.py
```diff
@@ -4,6 +4,14 @@
 import corgidrp.data as data
 from scipy import interpolate
 from scipy.ndimage import median_filter
+import glob
```
semaphoreP, line 7: Let's double check these imports to see if all of the are still necessary. I think `os` and `glob` are no longer used?

corgidrp/detector.py
```diff
@@ -506,10 +514,10 @@ def find_plateaus(streak_row, fwc, sat_thresh, plat_thresh, cosm_filter):
         return None
     
 def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
```
semaphoreP, line 516: The indentation level of this whole function seems to have gotten changed to 2 tabs. Can you change it back to 1 tab?

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+        	planet (str):   name of the planet neptune or uranus
+        	band (str):  band of the observation band1 or band4
+        	
+    	Returns:
+        	full_qe (np.array): ideal flat field
+        	full_residuals (np.array): flat field from uranus or neptune images
+        	percens_resel (np.array): percentage error of the flat field
+    """
+    
+    n = 420
```
semaphoreP, line 574: It looks like we have some hard coded variables here: `n`, `rad`, `rad1, and `rad4`. Do we think some of them may need to be tuned or tweaked with the real data? If so, we should expose them as variables we can change via keywords. If not, it's ok to leave them be. I just want to make sure we think about it.

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+
+        bool_innotzero = np.logical_and(nrr<rad,full_residuals[ymin:ymax,xmin:xmax]!=0)
+        bool_iniszero = np.logical_and(nrr<rad,full_residuals[ymin:ymax,xmin:xmax]==0)
+        bool_outisnotzero = np.logical_and(nrr>=rad,full_residuals[ymin:ymax,xmin:xmax]!=0)
+        
+        full_residuals[ymin:ymax,xmin:xmax][bool_innotzero] = (nrr_copy[bool_innotzero] + full_residuals[ymin:ymax,xmin:xmax][bool_innotzero])/2
+        full_residuals[ymin:ymax,xmin:xmax][bool_iniszero] = nrr_copy[bool_iniszero]
+        full_residuals[ymin:ymax,xmin:xmax][bool_outisnotzero] = full_residuals[ymin:ymax,xmin:xmax][bool_outisnotzero]
+
+    if band == '4':
```
semaphoreP, line 608: it looks like the code in the `if`/`elif` statements are the same except for the use of `mask1` vs `mask4`. Can we just use the if statement to set the mask = mask1 or mask4, and then only have these three lines of code written once? Just to limit code duplication.

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+        full_qe_resel = ndimage.convolve(full_qe, mask1)
+        full_residuals_resel = ndimage.convolve(full_residuals,mask1)
+        percens_resel = (((full_qe_resel - full_residuals_resel)/full_qe_resel)*100)
+    
+    full_residuals[full_residuals==0] = None
+    nx = np.arange(0,percens_resel.shape[1])
+    ny = np.arange(0,percens_resel.shape[0])
+    nxx,nyy = np.meshgrid(ny,nx)
+    nrr = np.sqrt((nxx-n/2)**2 + (nyy-n/2)**2)
+    full_residuals[nrr>165]= 1
```
semaphoreP, line 623: More constants here (165 and 302). Do we always expect these to be hard coded?

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+            band (str): 1 or 4
+
+    	Returns:
+    		data.FlatField (corgidrp.data.FlatField): a master flat for flat calibration using on sky images of planet in band specified
+    		
+	"""
+    smooth_images=[];raster_images_cent=[];cent=[]; act_cents=[];frames=[]; files=[];
+    planims=planet+'_band_'+band
+    for i in range(len(dataset)):
+        filename=Path(dataset[i].filename).stem.split('_')[0]+'_band_'+Path(dataset[i].filename).stem.split('_')[2]
```
semaphoreP, line 648: I'm not sure we can assume the filename will look exactly like this. What's the purpose of this block? just to exclude files that aren't flat field images that got mistakenly included? If so, I wouldn't worry too much about that for now.

semaphoreP: The default CGI filenames will look something like CGI_L1_100_0200001001001100001_20270101T120000_000.fits

corgidrp/mocks.py
```diff
@@ -261,6 +264,160 @@ def create_simflat_dataset(filedir=None, numfiles=10):
...
+        dither_sizex (int):(Required) Size of the dither in X axis in pixels
+        dither_sizey (int):(Required) Size of the dither in X axis in pixels
+        row_cent (int): (Required)  X coordinate of the centroid
+        col_cent (int): (Required)  Y coordinate of the centroid
+        n_dith (int): number of dithers required
+        mask_size (int): Size of the mask in pixels
+        snr (int): Required SNR in the planet images
+        planims (str): Planet and band
+        
+	Returns:
```
semaphoreP, line 283: "Returns:" should have one less indent (same level as "Args:")

corgidrp/mocks.py
```diff
@@ -261,6 +264,160 @@ def create_simflat_dataset(filedir=None, numfiles=10):
...
+        mask (int): (Required)  Mask used for the image.
+        data (float):(Required) Data in array npixels*npixels format to be raster scanned
+        dither_sizex (int):(Required) Size of the dither in X axis in pixels
+        dither_sizey (int):(Required) Size of the dither in X axis in pixels
+        row_cent (int): (Required)  X coordinate of the centroid
+        col_cent (int): (Required)  Y coordinate of the centroid
+        n_dith (int): number of dithers required
+        mask_size (int): Size of the mask in pixels
+        snr (int): Required SNR in the planet images
+        planims (str): Planet and band
```
semaphoreP, line 281: please provide the valid options in the docstring, given there are only 4 valid options

tests/test_create_flatfield.py
```diff
@@ -0,0 +1,88 @@
...
+    assert flatdivided_dataset[0].err_hdr["Layer_2"] == "FlatField_error"
+    print("mean of all simulated data",np.mean(simflat_dataset.all_data))
+    print("mean of all simulated data error",np.nanmean(simflat_dataset.all_err) )
+    print("mean of all flat divided data:", np.nanmean(flatdivided_dataset.all_data))
+    print("mean of flatfield:", np.nanmean(onsky_flatfield.data))
+    print("mean of flatfield err:", np.nanmean(onsky_flatfield.err))
+    
+    err_flatdiv=np.nanmean(flatdivided_dataset.all_err)
+    err_estimated=np.sqrt(((np.nanmean(onsky_flatfield.data))**2)*(np.nanmean(simflat_dataset.all_err))**2+((np.nanmean(simflat_dataset.all_data))**2)*(np.nanmean(onsky_flatfield.err))**2)
+    print("mean of all flat divided data errors:",err_flatdiv)
```
semaphoreP, line 80: I'm not sure what the goal of these lines are. Should this be an assert statement that checks that the 2 are approximately the same?

semaphoreP (changes requested):
Looking good now! I have some minor coding comments, and 1 bigger conceptual thing we should think about: we should think about what parameters in the flat field generation should be tunable and exposed to the user as optional keywords in the `create_onsky_flatfield` function. Perhaps @rzellem has some thoughts on this?

maxwellmb:
> Looking good now! I have some minor coding comments, and 1 bigger conceptual thing we should think about: we should think about what parameters in the flat field generation should be tunable and exposed to the user as optional keywords in the `create_onsky_flatfield` function. Perhaps @rzellem has some thoughts on this?

As a general philosophy (not having looked at the code): Given that Flat Fielding is classified as Software Class C, I'd like to maximize the flexibility by putting as much as possible as function keyword arguments.

corgidrp/detector.py
ramya-anche: The 165 pixels come from the radius of the coronagraph FOV for which the flat field is created. Since the corgidrp flatfield is 1024x1024 pixels, I padded 302 pixels on either side of this FOV with 1s. Would you prefer these to be nan's? 

Screenshot 2024-08-14 at 10 21 10 AM

corgidrp/detector.py
ramya-anche: This is intended to pick only those files corresponding to the planet and Filter for which the flatfield has to be created. But I can modify this; I think I can pull the header keywords [Target] and [Filter] and do the same thing.

tests/test_create_flatfield.py
ramya-anche: yeah, I missed adding the asset statement, will fix

corgidrp/detector.py
semaphoreP: I think you can assume that the dataset passed in is for the specific TARGET and FILTER

corgidrp/detector.py
semaphoreP: 1s is fine. I think the main thing is possibly allowing these to be optional keywords that we can change in the future

ramya-anche: okay! I will make all the hard-coded values as keywords

tests/test_create_flatfield.py
```diff
@@ -0,0 +1,88 @@
...
+    
+    ###### create simulated raster scanned data
+    # check that simulated raster scanned data folder exists, and create if not
+    file_dir = os.path.join(os.path.dirname(__file__), "simdata_rasterscan")
+    data_dir = os.path.join(os.path.dirname(__file__),"test_data/")
+    if not os.path.exists(file_dir):
+        os.mkdir(file_dir) 
+    filenames = glob.glob(os.path.join(data_dir, "med*.fits"))
+    data_set = data.Dataset(filenames)
+    mocks.create_onsky_rasterscans(data_set,filedir=file_dir,planet='neptune',band='1')
```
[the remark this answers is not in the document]
semaphoreP: Yes, that would be great! Just to exercise the other options. It can be mostly copied from this test if that's easiest.

corgidrp/mocks.py
```diff
@@ -261,6 +264,160 @@ def create_simflat_dataset(filedir=None, numfiles=10):
...
+    if planims == 'neptune_band_1' or planims=='neptune_band_4':
+        planrad = 54
+        dith_end = n_dith
+    elif planims == 'uranus_band_1' or planims=='uranus_band_4':
+        dith_end = n_dith+1
+        planrad = 90
+    for i in np.arange(-n_dith,dith_end):
+        for j in np.arange(-n_dith,dith_end):
+            mask_data = data.copy()
+            image_data = mask_data[row_min + (dither_sizey * j):row_max + (dither_sizey * j), col_min + (dither_sizex * i):col_max + (dither_sizex * i)]
```
rzellem, line 316: if I'm reading this correctly, it looks like the code requires some beforehand knowledge about the dither locations to find and "lock-in" on the dither positions - one concern is that the planets are moving over the images (because Roman does not have non-sidereal tracking), so the pre-defined locations might not be an exact match to the planet's actual position - do you think your code will be robust against this? or do we need to think up an additional method?

corgidrp/detector.py
```diff
@@ -4,6 +4,14 @@
...
+import glob
+import os
+import astropy.io.fits as fits
+import photutils.centroids as centr
+from pathlib import Path
+from scipy.ndimage import gaussian_filter as gauss
+from photutils.aperture import CircularAperture
+from scipy import ndimage
 
 def create_flatfield(flat_dataset):
```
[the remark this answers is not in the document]
semaphoreP: This was a placeholder function we were using for some other code (in particular, the bad pixel map). We should eventually get rid of it if we're not using it, but this can be for a future PR.

corgidrp/detector.py
ramya-anche: I have made these the keywords for the create_onsky_flatfield function.

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
-    possible_sat_fwcs_arr = np.append((emgain_arr * fwcpp_arr)[:,np.newaxis], fwcem_arr[:,np.newaxis],axis=1)
-    sat_fwcs = sat_thresh * np.min(possible_sat_fwcs_arr,axis=1)
-
-    return sat_fwcs
+	possible_sat_fwcs_arr = np.append((emgain_arr * fwcpp_arr)[:,np.newaxis], fwcem_arr[:,np.newaxis],axis=1)
+	sat_fwcs = sat_thresh * np.min(possible_sat_fwcs_arr,axis=1)
+
+	return sat_fwcs
+
+def flatfield_residuals(images,planims=None):
```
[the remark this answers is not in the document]
ramya-anche: changed the keyword to the planet and also added 'N' to have the flexibility to change the number of images for creating the matching filter

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+    """Turn this dataset of image frames of neptune or uranus and create matched filters and estimate residuals after 
+     dividing from matched filters
+
+     Args:
+    	images (np.array): 2D array of cropped neptune or uranus image frames
```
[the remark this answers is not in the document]
ramya-anche: Now, the `images` contain only the cropped images of Neptune or Uranus. Removed the corresponding masks as we aren't using them to create the matching filter.

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+        	band (str):  band of the observation band1 or band4
+        	
+    	Returns:
+        	full_qe (np.array): ideal flat field
+        	full_residuals (np.array): flat field from uranus or neptune images
+        	percens_resel (np.array): percentage error of the flat field
+    """
+    
+    n = 420
+    full_qe=np.random.normal(1,.03,(n,n))
```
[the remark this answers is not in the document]
ramya-anche: This is used to estimate the error in the flatfield `std_resel = (((full_qe_resel - full_residuals_resel)/full_qe_resel))`
`full_residuals_resel` is the flat field created by combining Neptune or Uranus images
`full_qe_resel` is what the ideal flatfield would be

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+        yc = int(centroid[1])
+        up_radius=55
+        smooth_images.append(planet_image)
+        raster_images_cent.append((planet,band,smooth_images[j][yc-up_radius:yc+up_radius,xc-up_radius:xc+up_radius],qe_prnu_fsm_raster[yc-up_radius:yc+up_radius,xc-up_radius:xc+up_radius]))
+        cent.append((yc-up_radius,yc+up_radius,xc-up_radius,xc+up_radius))
+        frame=data.Image(smooth_images[j][yc-up_radius:yc+up_radius,xc-up_radius:xc+up_radius], pri_hdr=prihdr, ext_hdr=exthdr)
+        frames.append(frame)
+    dataset=data.Dataset(frames)
+    resi=flatfield_residuals(raster_images_cent,planims=planet)
+    onskyflat=combine_rasters(resi,planet=planet,band=band,cent=cent)[1]
```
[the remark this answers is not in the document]
ramya-anche: Now,` flatfield_residuals `are returning an array of 1 dimension.

corgidrp/mocks.py
ramya-anche: I don't remember exactly whether we discussed whether the dither position or some related information would be available in the header. Do we need to query the position of Uranus or Neptune somewhere in the step function?

rzellem: I would assume that we do not get any knowledge from the FITS header as to the location of Uranus/Neptune on the focal plane- I'd recommend that your code automatically finds the planet and centroids on it, to be safe

corgidrp/detector.py
```diff
@@ -520,7 +525,161 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+
+    	Returns:
+    		data.FlatField (corgidrp.data.FlatField): a master flat for flat calibration using on sky images of planet in band specified
+    		
+	"""
+    smooth_images=[];raster_images_cent=[];cent=[]; act_cents=[];frames=[]; files=[];
+    for i in range(len(dataset)):
+        target=dataset[i].pri_hdr['TARGET']
+        filter=dataset[i].pri_hdr['FILTER']
+        if planet==target and band==filter:
```
semaphoreP, line 659: Two things here: 

1. You should expect that all files passed in are indeed for that filter and planet.
2. The behavior of `planet` and `band` should be that if the user doesn't pass in either (i.e., they are None), then we look in the header to determine it. Otherwise, use the planet and band specified by user input. We don't need to compare the input planet/band against what's in the header. This just gives us the flexibility to override the header values, but otherwise, we shouldn't be needing to user planet/band.

corgidrp/detector.py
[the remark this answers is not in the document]
semaphoreP: Is the shape of `images` 2-D (x,y) or 3-D (N_images, x, y)?

corgidrp/detector.py
```diff
@@ -520,7 +525,161 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+def create_onsky_flatfield(dataset, planet=None,band=None,up_radius=None,im_size=None,rad_mask1=None, rad_mask4=None,  planet_rad=None, n_pix=None, n_pad=None,N=None):
+    """Turn this dataset of raster scanned image frames of neptune or uranus into on-sky calibrated flat field 
+    
+        Args:
+            dataset (corgidrp.data.Dataset): a dataset of Image frames (L2a-level)
+            planet (str): neptune or uranus
+            band (str): 1 or 4
+            up_radius (int): Number of pixels on either side of centroided planet images (=55 pixels for Neptune and uranus)
+            im_size (int): x-dimension of the planet image (in pixels= 420 for the HST images)
+            N (int): Number of images to be combined for creating a matched filter (defaults to 3, can be changed)
```
semaphoreP, line 643: The order of the arguments in the docstring should match the order they are defined in the fucntino signature.

tests/test_create_flatfield_uranus.py
semaphoreP, line 1: can you combine this whole file with the neptune test. Since each are separate functions, you can just put them together in a single file.

semaphoreP (commented):
Some more comments about how the keywords in a function should be structured. A good rule is that not specifying keywords should not cause the function to crash.

corgidrp/detector.py
ramya-anche: When I create a `dataset` to be passed to the `create_flatfield` function, it has all the  `.*fits` images from the folder, which might contain both Uranus and Neptune and band 1 and 4. I thought what if the user doesn't have that information ?. The other option is, I do this comparison in `test function` and pass only the dataset that corresponds to the planet and band.

semaphoreP: I think we should always know which files are the ones we want for a given planet/filter. You can do this in the test function if that simplifies your code.

corgidrp/detector.py
```diff
@@ -520,7 +525,155 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+    
+    return (full_qe,full_residuals,std_resel)
+    
+    
+def create_onsky_flatfield(dataset, planet=None,band=None,up_radius=55,im_size=420,N=3,rad_mask=1.25,  planet_rad=54, n_pix=165, n_pad=302):
+    """Turn this dataset of raster scanned image frames of neptune or uranus into on-sky calibrated flat field 
+    
+        Args:
+            dataset (corgidrp.data.Dataset): a dataset of Image frames (L2a-level)
+            planet (str): neptune or uranus
```
semaphoreP, line 639: From my previous comment: If `planet` and `band` are not passed in, the code should be using the header keywords `TARGET` and `FILTER` to determine the planet and band. This makes it so that the user doesn't have to always manually change the `planet` and `band` when calling this function.

corgidrp/detector.py
```diff
@@ -520,7 +525,155 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+        	
+    	Returns:
+        	full_qe (np.array): ideal flat field
+        	full_residuals (np.array): flat field created from uranus or neptune images
+        	std_resel (np.array):  Error of the flat field
+    """
+    n = im_size
+    full_qe=np.random.normal(1,.03,(n,n))
+    full_residuals = np.zeros((n,n))
+    if planet_rad==None:
```
semaphoreP, line 580: shouldn't this value be different for either Uranus or Neptune?

semaphoreP (commented):
I think my previous comments regarding header keywords weren't fully addressed so I'm making more specific comments on them.

tests/test_create_flatfield.py
```diff
@@ -0,0 +1,174 @@
...
+    
+    err_flatdiv=np.nanmean(flatdivided_dataset.all_err)
+    err_estimated=np.sqrt(((np.nanmean(onsky_flatfield.data))**2)*(np.nanmean(simflat_dataset.all_err))**2+((np.nanmean(simflat_dataset.all_data))**2)*(np.nanmean(onsky_flatfield.err))**2)
+    print("mean of all flat divided data errors:",err_flatdiv)
+    print("Error estimated:",err_estimated)
+    assert(err_flatdiv == pytest.approx(err_estimated, abs = 1e-2))
+    
+    print(flatdivided_dataset[0].ext_hdr)
+    
+    #creating flatfield using uranus for band4
```
semaphoreP, line 97: the flat field test for uranus in band4 should be a separate function, since they are independent tests. 1 test per function.

corgidrp/detector.py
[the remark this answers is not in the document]
ramya-anche: 3-D (N_images, x,y)

corgidrp/mocks.py
ramya-anche: I think it does now. This is used to create the raster-scanned images, so the information on the dither location is needed. In the `create_onsky_flatfield` step function, the planet image centroiding is done automatically, and the image is cropped in pixels depending on the planet's radius (keyword).

rzellem: ah- I missed that thanks, so this is to only *simulate* data, not to treat the real data

corgidrp/detector.py
ramya-anche: yes, fixed it

corgidrp/detector.py
```diff
@@ -520,7 +525,178 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+            planet_rad (int): radius of the planet in pixels (planet_rad=50 for neptune, planet_rad=65)
+            n_pix (int): Number of pixels in radius covering the Roman CGI imaging FOV defaults to 165 pixels
+            n_pad (int): Number of pixels padded with '1s'  to generate the image size 1024X1024 pixels around imaging FOV (defaults to 302 pixels)
+
+
+    	Returns:
+    		data.FlatField (corgidrp.data.FlatField): a master flat for flat calibration using on sky images of planet in band specified
+    		
+	"""
+    if planet==None:
```
semaphoreP, line 657: general best practice is to use `is None` and not `== None` for none-checking. Although this isn't a huge deal, but should be a quick change.

semaphoreP (approved):
I just have one small Python style comment, but otherwise looks good to me.

corgidrp/detector.py
semaphoreP: you use this in several places in this module

ramya-anche: Fixed it

corgidrp/detector.py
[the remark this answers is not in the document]
rzellem: I'd be a little careful about assuming that the ideal flat field is a Gaussian distribution around 1 - there *might* be structure in the flats (e.g., due to deposition on a mask or the EMCCD) - but that's fine, as long as we know the structure of the flat field to high enough degree of accuracy, then that's all that matters

so I'm not sure you can use `full_qe` as you have it currently to measure the success of flight data - you might for simulated data (but Erin did include some structure in their code- they included Hubble-like "measles" for example)

rzellem (changes requested):
just one last comment to resolve when you get a moment, @ramya-anche , please and thank you!

corgidrp/detector.py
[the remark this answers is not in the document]
rzellem: so beware that your using `full_qe` as you have it defined might actually *over* estimate the error in the flatfield, since the flat might have structure in it that is not encapsulated by your Gaussian definition of `full_qe`

lastly...I'm not sure we need to actually calculate the error in the flat - we've done that via Erin's simulations - so we know that if we achieve a particular SNR on the images of Uranus/Neptune and apply a similar matched filter method, we should achieve our requirement (with margin)

so I wonder if we should just take this bit out - what do you think, @semaphoreP ?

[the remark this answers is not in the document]
rzellem: hi @ramya-anche and @semaphoreP - @maxwellmb and I quickly chatted on Slack and we agreed that we won't know the "true" flat field, as flats typically have structure in them, so we cannot use a Gaussian like this to estimate the error in our measured flats

rather, the best we can do is to estimate the measurement uncertainty from the measurements themselves, like the standard deviation of the images per pixel - so we'd effectively measure our measurement uncertainty from the data/measurements themselves

so we recommend removing this Gaussian

instead, we will have the error components already for each frame - so we will have a measurement uncertainty at each dither and then we will have a measurement uncertainty for the matched filter (which we can estimate from the matched filter by taking the stddev across each pixel in the dither), and when you divide one by the other, your error should also propagate - and this error is the one we need to track at the end of the day

corgidrp/detector.py
```diff
@@ -520,7 +528,149 @@ def calc_sat_fwc(emgain_arr,fwcpp_arr,fwcem_arr,sat_thresh):
...
+    full_qe[nrr>165]= 1
+    full_qe=np.pad(full_qe, ((302,302),(302,302)), mode='constant',constant_values=(1))
+    full_residuals=np.pad(full_residuals, ((302,302),(302,302)), mode='constant',constant_values=(1))
+    percens_resel=np.pad(percens_resel, ((302,302),(302,302)), mode='constant',constant_values=(1))
+    
+    return (full_qe,full_residuals,percens_resel)
+    
+    
+def create_onsky_flatfield(dataset, planet=None,band=None):
+    """Turn this dataset of raster scanned image frames of neptune or uranus into on-sky calibrated flat field
```
rzellem, line 638: @ramya-anche - can you please also add more detail for this function? in particular, please add the expectations for the input products - what data processing do we expect the images to have already gone through?

flat fielding is part of the data processing of L2a to L2b - and it is essentially the last step of the L2b, so you should write that this function assumes that the input data is a L2b image just prior to flat fielding, which is an image that has been "Divide by Kgain, divide by EM Gain, subtract Dark, desmear"

corgidrp/detector.py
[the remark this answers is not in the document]
ramya-anche: Hi @rzellem I have modified the `flatfield_residuals` to estimate the errors from individual dithers and matched filters. Can you review and see if this is all right?

corgidrp/detector.py
rzellem: ^ @ramya-anche

corgidrp/detector.py
```diff
@@ -726,18 +701,23 @@ def create_onsky_flatfield(dataset, planet=None,band=None,up_radius=55,im_size=4
         yc = int(centroid[1])
         up_radius=up_radius
         smooth_images.append(planet_image)
+        planet_image_err=np.random.normal(np.nanmedian(planet_image),5,(image_size[0],image_size[1]))
```
rzellem, line 704: Can you please explain this line, @ramya-anche ? I'm not sure I'm following, sorry.

corgidrp/detector.py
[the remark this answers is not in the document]
rzellem: @semaphoreP - I think we need to compute an "error" to properly track the additional uncertainty that comes in when dividing by our matched filter - because the matched filter itself will have an uncertainty (that we measure by taking the standard deviation along each pixel of all of the "stacked" images of Uranus or Neptune when making the matched filter)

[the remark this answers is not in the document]
semaphoreP: Ah I misread your comment, you are also referring to the error in the filter itself, not just the data. I think that's a fine way to do it.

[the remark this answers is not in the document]
rzellem: not sure I agree with that error propagation, but happy to punt on it for now and revisit it later :)

@ramya-anche - can you please add a "TODO" in your code to note that we need to update the error propagation so we don't forget later?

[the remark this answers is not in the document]
rzellem: @semaphoreP - in the interest of time, should we punt on this or do we have the time to incorporate it correctly now?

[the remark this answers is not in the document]
semaphoreP: Errors are not a requirement. I think we either put one in that's somewhat accurate, or we don't include one at all (as we aren't anticipating flat field to be the dominant noise source). I think probably the easiest path is to not include it for now, because we don't want one that's significantly incorrect. 

The error I quoted is the statistical error of doing a division by a template matched filter (assuming perfect filter). But I would need to look at the details on how you are doing a matched filter to see if it's appropriate.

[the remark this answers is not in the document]
semaphoreP: Not including it is basically setting the FlatField err field to all zeros. @ramya-anche, hopefully that should be easy to do.

PR #155, opened 2024-08-01
add arguments to make it easier to integrate into larger builds and test scripts

al-niessner:
## Describe your changes

Added arguments keeping original workflow.

## Type of change

Enhancement - kept old behavior but expanded its utility

- [x ] I have linted my code
- [ x] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [ x] I have verified that all docstrings are properly formatted and added new documentation, as needed


semaphoreP (approved):
Looks good to me.

PR #164, opened 2024-08-16
Mocks should not access test_data dir

### LINKED ISSUE #162: Unit tests should run through a pip install without the -e flag
**Describe the bug**
We want corgidrp to run all unit tests without needing to install in place using the -e flag.

**To Reproduce**
`pip install .` (no -e) and then `pytest`

**Expected behavior**
The reproduction above should run without errors

**corgidrp version:**
 v0.1.2

**Additional context**
Need to add some test_data files to MANIFEST.in so they are packaged in pip
Update Github actions to do a `pip install .` instead of a `pip install -e .`.

semaphoreP:
## Describe your changes
mocks.py should not access tests/test_data dir since regular pip install users will not have access to that dir. Instead, to generate mock data that requires data from tests/test_data, the filepath to the appropriate file should be passed in. 

## Type of change

Closes #162 

- [x] I have linted my code

### DIFF
--- a/corgidrp/mocks.py
+++ b/corgidrp/mocks.py
@@ -291,11 +291,12 @@ def create_flatfield_dummy(filedir=None, numfiles=2):
     flatfield = data.Dataset(frames)
     return flatfield
 
-def create_nonlinear_dataset(filedir=None, numfiles=2,em_gain=2000):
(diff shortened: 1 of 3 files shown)


maxwellmb (approved):
look good to me

PR #166, opened 2024-08-21
Use dataset instead of single frame for guessing recipes

semaphoreP:
## Describe your changes

To help @hsergi with nonlinearity, since we will need to check multiple files, and not just the first file

## Type of change

- [x] I have linted my code

### DIFF
--- a/corgidrp/walker.py
+++ b/corgidrp/walker.py
@@ -66,12 +66,13 @@ def autogen_recipe(filelist, outputdir, template=None):
     Returns:
         json: the JSON recipe to process the filelist
     """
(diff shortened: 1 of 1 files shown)


maxwellmb:
Doesn't this still just use the first frame?

semaphoreP:
You now have the option to use the whole dataset.

maxwellmb:
Ah, I see. So the purpose of this change is to enable future developers (Sergi soon) to write their own `guess_template` logic with the whole `dataset`?

semaphoreP:
Yep

maxwellmb (approved):

Issue #128, opened 2024-07-04
.corgidrp directory for config made in __init__.py must be deleted in order for an updated directory to be made


kjl0025:
Perhaps we should add to the README that a user should delete .corgidrp if a new config is desired upon installation.  Or we could have something in __init__.py to see if necessary keys are present and if not to remake the config?

kjl0025:
Sorry I'm just getting to this.  I deleted my .corgidrp to get __init__.py to install a fresh config, so I don't recall the specific error.  But the error was caused by the addition of a key to the config,  and __init__.py didn't make a new config b/c it currently only does that if one doesn't already exist.   I could fix this in another PR soon if you'd like.

semaphoreP:
Sure, although I'd put this at low priority until we start releasing the code for public consumption.

semaphoreP:
Remaining task: add to README that deleting the .corgidrp directory is a possible troubleshooting solution

PR #169, opened 2024-08-23
pipeline setting install fix and additions

semaphoreP:
## Describe your changes
This should partially address the issues in #128 regarding pipeline install. Now in first load, we check specifically that the config file and the default calibrations sub directory exist (and if not, create them). We also check we've loaded all the cals for the default_calib_dir (where DetectorParams file lives) when we import caldb (in case someone completely deleted or wiped their caldb). 

Adds two new pipeline settings (Both default to False currently):
 * `corgidrp.jit_calib_id`: identifies the calibration file at the execution of that particular step rather than the creation of the recipe.
 * `corgidrp.skip_missing_cal_steps`: skips step functions in which we have zero suitable calibration files in the caldb.

## Type of change

- [x] I have linted my code

### DIFF
--- a/tests/test_walker.py
+++ b/tests/test_walker.py
@@ -226,8 +226,69 @@ def test_saving():
     this_caldb.remove_entry(new_nonlinearity)
 
 
+
+def test_skip_missing_calib():
+    """
+    Tests the option of skipping steps with missing calibrations
+    """
+    # turn on skipping
+    old_setting = corgidrp.skip_missing_cal_steps
+    corgidrp.skip_missing_cal_steps = True
+
+    # use an empty test caldb
+    calibdir = os.path.join(os.path.dirname(__file__), "testcalib")
+    if not os.path.exists(calibdir):
+        os.mkdir(calibdir)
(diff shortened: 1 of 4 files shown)


tests/test_walker.py
```diff
@@ -226,8 +226,133 @@ def test_saving():
...
+
+    # clean up
+    mycaldb.remove_entry(new_nonlinearity)
+
+    corgidrp.jit_calib_id = old_setting
+
+
 if __name__ == "__main__":#
-    test_saving()
+    test_jit_calibs()
```
maxwellmb, line 356: Shouldn't all the test functions be listed down here?

maxwellmb (changes requested):

tests/test_walker.py
semaphoreP: Doesn't matter. that's just for debugging purposes (if you run `python test_walker.py` from command line. Pytest runs tests by identifying all the functions that follow the `test_*` format.

maxwellmb (approved):

Issue #108, opened 2024-06-18
Use the ‘applied gain’ header keyword instead of the ‘commanded gain’ header keyword wherever EM gain is used.


mygouf:
There should be a small % difference between the commanded and applied EM gain. Thus it would be preferable to use the applied EM gain instead of the commanded gain. 
Please note that the applied EM gain header doesn't exist yet so this issue cannot be implemented yet, but I'm created it to make sure it doesn't fall under the cracks. Marie will work with SSC to make that happen when Jim comes back from vacation.

maxwellmb:
@mygouf, can you comment on whether or not this is needed for R3.0.1?

mygouf:
Yes, let's plan on having this implemented for R3.0.1.

kjl0025:
The CMDGAIN is *supposed* to be the actual applied gain, assuming a good fit to the parametrization function.  In the future, we may implement a methodology of reading off the EM gain from a given frame and assign some new header like EMGAIN_M for that read-off value.  For low-gain frames, the methodology is only possible if a set of illuminated flat-field frames at the same commanded gain and a set at unity gain accompany the observation. For high-gain, the methodology requires a set of dark frames at the same commanded gain accompanying the observation.  

Recommended practice:  We could do something like what is currently done in the `corgidrp.darks.calibrate_darks_lsq` function with a `try` and `except` statement so that even if "EMGAIN_M" for measured EM gain is never implemented, the code doesn't fail:
```
try: # if EM gain measured directly from frame TODO change hdr name if necessary
     EMgain_arr = np.append(EMgain_arr, datasets[i].frames[0].ext_hdr['EMGAIN_M'])
except: # use commanded gain otherwise TODO change hdr name if necessary
     EMgain_arr = np.append(EMgain_arr, datasets[i].frames[0].ext_hdr['CMDGAIN'])
```

hsergi:
This has been implemented in calibrate_kgain.py, as well. https://github.com/roman-corgi/corgidrp/pull/116. 

See also https://collaboration.ipac.caltech.edu/display/romancoronagraph/EXCAM+headers+wishlist

maxwellmb:
Note: The `EMGAIN_M` keyword that you and Kevin have used should be for the _measured_ EM Gain, not the "actual" EMGAIN, so we still have yet to implement that everywhere.

maxwellmb:
[...]

PR #170, opened 2024-08-24
accommodation of EMGAIN_M, EMGAIN_A, and CMDGAIN in that order of priority

kjl0025:
## Describe your changes
Where absolute EM gain is desired, this PR tries to retrieve the header key measured EM gain (EMGAIN_M), and if applied EM gain (EMGAIN_A) if that fails, and commanded EM gain (CMDGAIN) if that fails.  For sorting by EM gain, commanded EM gain or applied EM gain is used instead of measured EM gain (since measured EM gain will vary frame to frame even if the commanded EM gain is the same for them all).

## Type of change

#108 

- [X ] I have linted my code
- [X ] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [X ] I have verified that all docstrings are properly formatted and added new documentation, as needed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,4 @@ tests/e2e_tests/*/*.json
 # Output figures
 figures/
 tests/figures/
+tests/test_data/simastrom/guesses.csv
--- a/corgidrp/calibrate_kgain.py
+++ b/corgidrp/calibrate_kgain.py
@@ -79,21 +79,21 @@ class CalKgainException(Exception):
     """Exception class for calibrate_kgain."""
 
 ################### function defs #####################
-    
+
 def ptc_bin2(frame_in, mean_frame, binwidth, max_DN):
-    """ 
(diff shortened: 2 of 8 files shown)


kjl0025:
And I apologize; the whitespace changes did not appear in my view of the diff in GItHub Desktop, but it shows up.  But you can click the gear option icon and select the option to hide whitespace changes, and it looks normal then.

maxwellmb (changes requested):
I'm sorry to say, but we've now reached the point where the whitespace issue must be addressed. It's not just a visual thing: it causes merge conflicts with other branches currently under development, which ends up just making the problem someone else's. It's gotta be a setting in your IDE or something...

kjl0025:
Ah, true, I didn't think of that before.  Okay, I'll fix this.  I'll probably do a new PR with a new branch with the same changes but without the whitespace changes.

PR #174, opened 2024-08-28
accommodation of EMGAIN_M, EMGAIN_A, and CMDGAIN in that order of priority

kjl0025:
## Describe your changes
Where absolute EM gain is desired, this PR tries to retrieve the header key measured EM gain (EMGAIN_M), and if applied EM gain (EMGAIN_A) if that fails, and commanded EM gain (CMDGAIN) if that fails. For sorting by EM gain, commanded EM gain or applied EM gain is used instead of measured EM gain (since measured EM gain will vary frame to frame even if the commanded EM gain is the same for them all).
(same changes as in deleted emgain_a branch but w/o whitespace differences)

## Type of change
New feature (non-breaking change which adds functionality)

#108 

- [ X] I have linted my code
- [ X] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [ X] I have verified that all docstrings are properly formatted and added new documentation, as needed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,4 @@ tests/e2e_tests/*/*.json
 # Output figures
 figures/
 tests/figures/
+tests/test_data/simastrom/guesses.csv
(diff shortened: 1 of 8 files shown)


kjl0025:
I had to turn off the VS Code option that enforced the deletion of trailing whitespaces.

maxwellmb (approved):

Issue #111, opened 2024-06-21
L2a -> L2b Step Function: CTI Correction 


maxwellmb:
This is a break-up from issue #20 which currently includes the step function and the calibration generation. 

We need a CTI correction step function for L2a -> L2b. 

As of right now it is a requirement to work, but @nasavbailey and @mygouf are trying to make it a "should". Should it become a "should" it will be ok to be a no-op code.

maxwellmb:
Closing with PR #180

PR #180, opened 2024-08-31
Porting over of the II&T PumpTrap Calibration Routines

maxwellmb:
## Describe your changes

The PR predominantly is a port of the II&T PumpTrap Calibration Routines: 

- It defines the TrapCalibration datatype
- It adjust the step_function to accept this kind of calibration
- It includes the code to analyze pump-trap data
- As part of the tests we genereate pumptrap data on the fly using emccd-detect (a new required package - non pip-able thanks to Kevin!)
- There are a few changes to the prescan_biassub function to accommodate a slightly different way of slices the arrays that seems to be common in some of the II&T Calibration functions

There are other minor changes along the way. 

I'm currently still working on finalizing some of the unit test behavior. 

## Type of change

This PR will resolve Issues #112 and #111

- [ ] I have linted my code


maxwellmb:
Apologies for the size of the PR, but these calibration functions are beasts.

corgidrp/data.py
```diff
@@ -763,21 +773,27 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
         # if this is a new calibration file, we need to bookkeep it in the header
-        # b/c of logic in the super.__init__, we just need to check this to see if it is a new NonLinearityCalibration file
+        # b/c of logic in the super.__init__, we just need to check this to see if 
+        # it is a new NonLinearityCalibration file
         if ext_hdr is not None:
             if input_dataset is None:
                 # error check. this is required in this case
-                raise ValueError("This appears to be a new Non Linearity Correction. The dataset of input files needs to be passed in to the input_dataset keyword to record history of this calibration file.")
-            self.ext_hdr['DATATYPE'] = 'NonLinearityCalibration' # corgidrp specific keyword for saving to disk
+                raise ValueError("This appears to be a new Non Linearity "
```
kjl0025, line 791: Just FYI, looks like you have an IDE linter at work on your computer (an example here, imposing a line length limit).  I don't think it affected too many things on this particular PR.

corgidrp/data.py
```diff
@@ -1317,6 +1333,53 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+    The format will be [n,10], where each entry will have: 
+    [row, column, sub-electrode location, index numnber of trap at this pixel/electrode, 
+    capture time constant, maximum amplitude of the dipole, energy level of hole, 
+    cross section for holes, R^2 value of fit, release time constant]
+
+     Args:
+        data_or_filepath (str or np.array): either the filepath to the FITS file to read in OR the 2D image data
+        pri_hdr (astropy.io.fits.Header): the primary header (required only if raw 2D data is passed in)
+        ext_hdr (astropy.io.fits.Header): the image extension header (required only if raw 2D data is passed in)
+        input_dataset (corgidrp.data.Dataset): the Image files combined together to make this bad pixel map (required only if raw 2D data is passed in)
```
kjl0025, line 1351: "bad pixel map" --> "trap calibration"

corgidrp/data.py
```diff
@@ -1317,6 +1333,53 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
         if 'DATATYPE' not in self.ext_hdr or self.ext_hdr['DATATYPE'] != 'AstrometricCalibration':
             raise ValueError("File that was loaded was not an AstrometricCalibration file.")    
         
+class TrapCalibration(Image):
+    """
+
+    Class for data related to charge traps that cause charge transfer inefficiency. 
+    The calibrationi is generated by trap-pumped data.
```
kjl0025, line 1340: "calibration"

corgidrp/data.py
kjl0025: And I see no `err_hdr`, and I think that's fine for this particular calibration.  It would amount to running an iterative trap correction algorithm for the central tau and cap values, and then multiple other times for the upper and lower bounds of these, and in practice, we could probably get an err on the observation frames we CTI-correct from the correction process itself apart from considering the error in tau and cap.

kjl0025: On second thought, I think in general those uncertainties the II&T function returns would be nice to have in an `err_hdr`, but as the 9/9 deadline is approaching, I'm okay keeping that as an enhancement in an issue.

corgidrp/detector.py
```diff
@@ -274,6 +280,69 @@ def slice_section(frame, obstype, key, detector_regions=None):
         raise Exception('Corners invalid. Tried to slice shape of {0} from {1} to {2} rows and {3} columns'.format(frame.shape, r0c0, rows, cols))
     return section
 
+
+def imaging_area_geometry(obstype='SCI',detector_regions=None):
```
kjl0025, line 284: This is the same as the already existing `imaging_area_geom` in detector.py.

corgidrp/detector.py
```diff
@@ -274,6 +280,69 @@ def slice_section(frame, obstype, key, detector_regions=None):
...
+
+    rows_im = detector_areas[obstype]['frame_rows'] - rows_parallel_ovr
+    cols_im = detector_areas[obstype]['frame_cols'] - cols_pre - cols_serial_ovr
+    r0c0_im = r0c0_image.copy()
+    r0c0_im[0] = r0c0_im[0] - (rows_im - fluxmap_rows)
+
+    return rows_im, cols_im, r0c0_im
+
+
+def imaging_slice(frame,obstype='SCI',detector_regions=None):
```
kjl0025, line 318: This is the same as `imaging_slice` further below in this file (which is the one your script must be calling since this comes before it).

corgidrp/l1_to_l2a.py
```diff
@@ -49,24 +51,39 @@ def prescan_biassub(input_dataset, noise_maps=None, return_full_frame=False, det
                 raise Exception(f"Observation type of frame {i} is not 'SCI' or 'ENG' or 'ENG_EM' or 'EMG_CONV'")
 
         # Get the reliable prescan area
-        prescan = slice_section(frame_data, obstype, 'prescan_reliable', detector_regions)
+        prescan = slice_section(frame_data, obstype, 'prescan', detector_regions=detector_regions)
```
kjl0025, line 54: I think this original version of this line should remain.  The argument `detector_regions` of `prescan_biassub` already handles `detector_regions`.

corgidrp/l1_to_l2a.py
```diff
@@ -1,9 +1,10 @@
 # A file that holds the functions that transmogrify l1 data to l2a data
-from corgidrp.detector import get_relgains, slice_section, detector_areas, flag_cosmics, calc_sat_fwc
+from corgidrp.detector import get_relgains, slice_section, detector_areas, flag_cosmics, calc_sat_fwc, imaging_slice, imaging_area_geom
 import numpy as np
 import corgidrp.data as data
 
-def prescan_biassub(input_dataset, noise_maps=None, return_full_frame=False, detector_regions=None):
+def prescan_biassub(input_dataset, noise_maps=None, return_full_frame=False, 
+                    detector_regions=None, use_imaging_area = False):
```
kjl0025, line 7: Instead of adding in `use_imaging_slice`, which would need to be tested in the unit test for `prescan_biassub`, you could always `return_full_frame` here and take a subframe of the result in the trap calibration function right after it calls `prescan_biassub`.

corgidrp/l1_to_l2a.py
kjl0025: Oh, never mind on this particular comment.  That's valid if we go this route where the reliable prescan can be accessed 2 different ways.

corgidrp/pump_trap_calibration.py
```diff
@@ -0,0 +1,3281 @@
...
+- Kevin Ludwick, UAH, 6/2022
+
+- MMB Changes for corgi-drp: 
+    - Removed everything meta and replaced with functions from detector.py
+    - Replaced II&T get_relgains function with corgidrp function -> Now 
+    requires a NonLinearityCorrection object to work. Defaults to None
+    - Removed image slicing
+    - Removed non-linearity correction
+    - Input dataset is now a Dataset object from corgidrp.data
+    - output is not a corgi.drp.TrapCalibration object
```
kjl0025, line 17: "is not" --> "is now"

corgidrp/pump_trap_calibration.py
```diff
@@ -0,0 +1,3281 @@
...
+                                int(split_electrode_str[1])
+
+        #Which trap is this at this location? 
+        trap_cal_array[i,3] = key_list[2]
+        
+        #Now extract the dictionary entry: 
+        dict_entry = trap_dict[key]
+
+        #Grab the first time constant 
+        #TODO: It seems like there could be three of these. How do we choose which one?
```
kjl0025, line 3254: True, when I suggested keeping the capture time constant, I kind of forgot there is in general more than one!  But I think this is fine given the upcoming deadline.  I can write up an issue about an enhancement to this step function thaedt includes a slot for the 2nd and 3rd capture time constant which is 0 if non-existent (and a trap is identified by at most 3 of these).  Same goes for max amplitude in line 3258.  And I'll include info about the inclusion of an `err_hdr` in the future as well.

corgidrp/pump_trap_calibration.py
```diff
@@ -0,0 +1,3281 @@
...
+        Number of times traps were identified that did not get used in
+        identifying release time constant values across all temperatures
+    two_or_less_count : int
+        Number of traps that only appeared at 2 or fewer temperatures.
+    noncontinuous_count : int
+        Number of traps that appeared at a noncontinuous series of
+        temperatures.
+
+    Args:
+        input_dataset (corgi.drp.Dataset): The input dataset to be analyzed. The dataset should be a stack of trap-pumped frames, sorted according to temperature and pumping scheme.
```
kjl0025, line 2179: I suppose they don't have to be sorted since the dataset is sorted in the code according to temp and scheme?

corgidrp/pump_trap_calibration.py
```diff
@@ -0,0 +1,3281 @@
...
+
+                #Get the data and the phase time - convert from us to seconds
+                phase_time = float(frame.ext_hdr[time_head])/10**6
+
+                timings.append(phase_time)   
+                frames.append(frame.data)
+
+            # no need for cosmic ray removal since we do ill. correction
+            # plus cosmics wouldn't look same as it would on regular frame,
+            # and they affect the detection of dipoles (or very low chance)
```
kjl0025, line 2341: Should be "they wouldn't affect".  I messed up that comment originally!

tests/test_pumptrap_calibration.py
```diff
@@ -0,0 +1,274 @@
...
+
+    # Set the seed - II&T ut tests don't work everytime, so let's fix it. 
+    np.random.seed(39)
+
+    #Generate the mock data:
+    test_data_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'test_data', "pump_trap_data_test")
+    metadata_file = os.path.join(os.path.abspath(os.path.dirname(__file__)), 'test_data', "metadata_test.yaml")
+    print("Generating mock data")
+    generate_mock_pump_trap_data(test_data_dir, metadata_file)
+    print("Done generating mock data")
```
kjl0025, line 70: Not sure if we're okay with prints in pytest outputs?  If we are, that's fine.

corgidrp/l1_to_l2a.py
kjl0025: After looking at everything else and coming back to this, `use_imaging_slice` is implicitly tested by the trap-pump test, and I suppose it's okay to have redundant mechanisms which don't break anything else.

kjl0025 (changes requested):
Overall, very nice!  I will make an issue for a future enhancement covering the err_hdr, the addition of any tests needed for that addition, and the ability to document up to 3 capture time constants. I request just a few things for this PR:
-- Delete the first `imaging_slice` function found in detector.py
-- Delete `imaging_area_geometry` and change the call to it in pump_trap_calibration.py and test_pumptrap_calibration.py to a call to `imaging_area_geom` instead.
-- Fix the other minor things I mentioned (e.g., concerning typos in comments)

tests/test_pumptrap_calibration.py
```diff
@@ -0,0 +1,274 @@
...
+from corgidrp.data import Dataset
+from corgidrp.l1_to_l2a import prescan_biassub
+from corgidrp.l2a_to_l2b import em_gain_division
+from corgidrp.pump_trap_calibration import tpump_analysis, tau_temp
+
+
+
+# Adjust the system's limit of open files. We need to load 2000 files at once. 
+# some systems don't like that. 
+import resource
```
semaphoreP, line 15: do you need a `def teardown_module():` at the end that sets the system open file limit back to its original value?

tests/test_pumptrap_calibration.py
semaphoreP: It's fine. They're neither warnings or errors, and will only be displayed if the test fails.

tests/test_pumptrap_calibration.py
```diff
@@ -0,0 +1,274 @@
...
+                'cols': 107,
+                'r0c0': [106, 108]
+            }
+            },
+        }
+
+    #Subtract the prescane Bias
+    bias_subbed_dataset = prescan_biassub(pump_trap_dataset, detector_regions=detector_regions,use_imaging_area=True)
+
+    bias_subbed_dataset[0].save(filedir = "./", filename="bias_subbed_dataset_file0.fits")
```
semaphoreP, line 146: does this file and the emgain save file need to happen? It's not clear they are ever used.

If they are used, the outputdir should not be `./`. It should be defined relative to this file, so that it is always saved to the same location.

semaphoreP (commented):
a few additional comments, to supplement Kevin's.

tests/test_pumptrap_calibration.py
```diff
@@ -0,0 +1,274 @@
...
+        #Get the phase time from the filename: its between the string "phasetime" and the ".fits" extension at the end
+        # phase_time = os.path.basename(data_filenames[j]).split("phasetime")[1].split(".fits")[0]
+        # frame.ext_hdr['TPTAU'] = float(phase_time)
+
+    #Run the bias subtraction
+    # #TODO Figure out which detector regions to pass in here. 
+    # arrtype = pump_trap_dataset[0].ext_hdr['ARRTYPE']
+
+    #Detector regions for the smaller pump_trap_data - taken from metadata_test.yaml
+    detector_regions = {
```
kjl0025, line 107: One other thing:  This is in mocks.py, so you could do `from corgidrp.mocks import detector_areas_test` to get this instead of copying it here.

corgidrp/pump_trap_calibration.py
```diff
@@ -0,0 +1,3281 @@
...
+
+            #Find the TPSCHEM keyword that is non-zero
+            this_num_pumps = [header0[x] for x in scheme_header_keywords]
+            this_num_pumps = np.array(scheme_list[i] )
+            this_scheme = int(np.where(this_num_pumps != 0)[0]) + 1
+            sch_list.append(this_scheme)
+
+            #Grab the number of pumps from the first dataset. 
+            if i == 0: 
+                num_pumps = this_num_pumps[this_scheme-1]
```
kjl0025, line 2306: Sorry I didn't catch this before.  As I work on the e2e based on this branch, I go a bit deeper into it.  In theory, the number of pumps could be different for each scheme, we shouldn't assume `num_pumps` is the same number for all schemes.

corgidrp/pump_trap_calibration.py
maxwellmb: Issue made!

tests/test_pumptrap_calibration.py
maxwellmb: good call!

corgidrp/pump_trap_calibration.py
```diff
@@ -3128,7 +3128,7 @@ def _el_loc_3(sch1, or1, prob1, sch2, or2, prob2, sch3, or3, prob3,
             range = [[0, max(1, E_vals.max()) + 0.5/bins[0]],
                 [0, max(1e-14, cs_vals.max()) + 1e-14*0.5/bins[1]]])
 
-        nrows, ncols, _ = imaging_area_geometry()
+        nrows, ncols, _ = imaging_area_geom('SCI')
```
kjl0025, line 3131: Actually, trap-pumped frames are usually 'ENG', even though the test data was 'SCI'.  For this, I think you should use `ext_hdr['ARRTYPE]`.

corgidrp/l1_to_l2a.py
```diff
@@ -49,24 +51,39 @@ def prescan_biassub(input_dataset, noise_maps=None, return_full_frame=False, det
...
+            if use_imaging_area: 
+                image_data = imaging_slice(obstype, frame_data, detector_regions=detector_regions)
+                image_dq = imaging_slice(obstype, frame_dq, detector_regions=detector_regions)
+
+                image_err = []
+                for err_slice in frame_err:
+                    image_err.append(imaging_slice(obstype, err_slice, detector_regions=detector_regions))
+                image_err = np.array(image_err)
+
+                prows, _, r0c0 = imaging_area_geom('SCI',detector_regions=detector_regions)
```
kjl0025, line 67: Instead of 'SCI', use `obstype`.  This function gets the area right even if using 'ENG'.

tests/test_pumptrap_calibration.py
maxwellmb: I don't really know why I was doing this...

tests/test_pumptrap_calibration.py
maxwellmb: good call!

corgidrp/data.py
maxwellmb: very weird, but good for me to keep an eye on in the future. It's possible it was me...

corgidrp/mocks.py
```diff
@@ -980,4 +985,836 @@ def create_astrom_data(field_path, filedir=None, subfield_radius=0.02):
     frames.append(frame)
     dataset = data.Dataset(frames)
 
-    return dataset
\ No newline at end of file
+    return dataset
+
+def P1(time_data, offset, tauc, tau, num_pumps = 10000):
```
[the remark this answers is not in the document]
maxwellmb: Yes, good thinking. This is done now. I refactored the functions from `trap_fit` to be callable. I'll note that there's actually a second set of these functions that live in `trap_fit_const`, but they take slightly different inputs, so they remain defined within that function (they are not used by `mocks.py`)

corgidrp/pump_trap_calibration.py
maxwellmb: makes sense! Fixed now!

corgidrp/pump_trap_calibration.py
```diff
@@ -0,0 +1,3281 @@
...
+            raise KeyError('both_a must contain \'amp\' and \'t\' keys.')
+        if x != y:
+            raise ValueError('both_a[\'amp\'] and both_a[\'t\'] must have the '
+            'same number of elements')
+    # if input for both_a doesn't conform to expected dictionary structure,
+    # exceptions will be raised
+
+    # define all these inside this function because they depend also on
+    # num_pumps, and I can't specify an unfitted parameter in the function
+    # definition if I want to use curve_fit
```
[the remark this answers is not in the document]
maxwellmb: done!

tests/test_pumptrap_calibration.py
maxwellmb: my understanding is that this change is only for the current process and its children, so we shouldn't have to change it back. Is it fine as is then?

corgidrp/l1_to_l2a.py
maxwellmb: done!

corgidrp/pump_trap_calibration.py
maxwellmb: good call! done!

maxwellmb:
@kjl0025, @semaphoreP. I think I've addressed all the comments, please let me know if there's anything else.

kjl0025:
I'm good to go with this!

kjl0025:
And my e2e PR for trap-pumping shouldn't be far behind this merge.

tests/test_pumptrap_calibration.py
semaphoreP: just checked and this is correct, so all good.

semaphoreP (approved):
Looks good! @kjl0025 can you approve as well?

kjl0025 (approved):

PR #184, opened 2024-09-06
frame combination step function and test

semaphoreP:
## Describe your changes

New step function to combine subsets of frames together. Made currently for the intent of the flat field e2e to combine subexposures of a raster scan. I started a new combine.py since I anticipate we want to combine frames in different frames (e.g., collapse via keywords, sigma clipping, etc). 

I also updated the recording of parent filenames to include grand-parents. This way, a file (E.g., flat field) would record its parents, but also the frames that were combined together from its parents. In the current logic, great grand parents of arbitrary depth are also tracked. 

## Type of change

- [x] I have linted my code

### DIFF
--- a/corgidrp/combine.py
+++ b/corgidrp/combine.py
@@ -1,0 +1,79 @@
+"""
+Module to support frame combination
+"""
+import numpy as np
+import corgidrp.data as data
+
+
+def combine_images(data_subset, err_subset, dq_subset, collapse="lower"):
+    """
+    Combines several images
+
+    Args:
+        data_subset (np.array): 3-D array of N 2-D images
+        err_subset (np.array): 4-D array of N 2-D error maps
+        dq_subset (np.array): 3-D array of N 2-D DQ maps
+    """
+    tot_frames = data_subset.shape[0]
+    # mask bad pixels
+    bad = np.where(dq_subset > 0)
+    data_subset[bad] = np.nan
+    err_subset[bad[0],:,bad[1],bad[2]] = np.nan
+    # track the number of good values that go into the combination
+    n_samples = np.ones(data_subset.shape)
+    n_samples[bad] = 0
+    n_samples = np.sum(n_samples, axis=0)
+    if collapse.lower() == "mean":
+        data_collapse = np.nanmean(data_subset, axis=0) * tot_frames
+        err_collapse = np.sqrt(np.nanmean(err_subset**2, axis=0)) * tot_frames/np.sqrt(n_samples) # not sure if this is correct, but good enough for now
+    elif collapse.lower() == "median":
+        data_collapse = np.nanmedian(data_subset, axis=0) * tot_frames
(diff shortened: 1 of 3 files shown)


corgidrp/combine.py
```diff
@@ -0,0 +1,91 @@
+"""
+Module to support frame combination
+"""
+import numpy as np
+import corgidrp.data as data
+
+
+def combine_images(data_subset, err_subset, dq_subset, collapse="lower"):
+    """
+    Combines several images together
```
maxwellmb, line 10: It looks to me like the operations below try to maintain the total number of counts, which isn't the intuitive mean or median behavior that you'd expect. I think we might want the option to multiply by tot_frames or not and to describe that more in the docstring.

corgidrp/combine.py
```diff
@@ -0,0 +1,91 @@
...
+        err_subset = np.copy(input_dataset.all_err[num_frames_per_group*i:num_frames_per_group*(i+1)])
+        dq_subset = input_dataset.all_dq[num_frames_per_group*i:num_frames_per_group*(i+1)]
+
+        data_collapse, err_collapse, dq_collapse = combine_images(data_subset, err_subset, dq_subset, collapse=collapse)
+
+        # grab the headers from the first frame in this sub sequence
+        pri_hdr = input_dataset[num_frames_per_group*i].pri_hdr.copy()
+        ext_hdr = input_dataset[num_frames_per_group*i].ext_hdr.copy()
+        err_hdr = input_dataset[num_frames_per_group*i].err_hdr.copy()
+        dq_hdr = input_dataset[num_frames_per_group*i].err_hdr.copy()
```
maxwellmb, line 82: Should we also be grabbing the hdu_list if it exists?

maxwellmb (changes requested):
Overall this looks fine, the biggest issue is that the combination is much closer to a sum than it is a mean or median and this needs to be clearer, either with different keyword values to `collapse`, and/or a clearer docstring, in both the combine images function and the step function.

corgidrp/combine.py
```diff
@@ -0,0 +1,91 @@
+"""
+Module to support frame combination
+"""
+import numpy as np
+import corgidrp.data as data
+
+
+def combine_images(data_subset, err_subset, dq_subset, collapse="lower"):
```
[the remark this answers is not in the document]
semaphoreP: oops typo. fixed

corgidrp/combine.py
semaphoreP: I added a second keyword to toggle the scaling by tot_frames

corgidrp/combine.py
semaphoreP: We were doing it right below, but I've moved it up here to be more explicit.

semaphoreP:
@maxwellmb ready for re-review

corgidrp/combine.py
maxwellmb: Whoops, I missed it, my bad!

maxwellmb (approved):

Issue #153, opened 2024-07-31
Create recipe template for boresight calibration


semaphoreP:
Is your feature request related to a problem? Please describe.

No recipe template exists for creating a boresight correction file for astrometric calibration

Describe the solution you'd like

A recipe template needs to be written to generate an Astrometric Calibration file from L1 data.

Additional context

In order to test this you will need to get your hands on the appropriate input data from II&T or TVAC and  make a simulated observation from it (this way, we include reaslitic noise). In addition please describe in your own words (~several sentences to a paragraph) what the recipe does and what the inputs and outputs are. Finally, please make sure the appropriate table entry is complete within the Confluence DRP Implementation Document table in Section 2.

Here's a checklist:

- [ ] Recipe Written
- [ ] Test Written and passing
- [ ] Table Entry Complete

PR #185, opened 2024-09-06
Boresight Calibration End to End Testing

manduhmia:
## Describe your changes
Creating recipe template for boresight calibration end-to-end testing.

## Type of change

Issue #153  

- [ y] I have linted my code
- [ y] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [ y] I have verified that all docstrings are properly formatted and added new documentation, as needed


manduhmia:
Currently the platescale calculation does not pass-- working on fixing this, but I wanted to create the PR

tests/test_astrom.py
```diff
@@ -36,17 +36,17 @@ def test_astrom():
...
     expected_northangle = 45
 
-    # check orientation is correct within 0.3 [deg]
+    # check orientation is correct within 0.05 [deg]
     # and plate scale is correct within 0.5 [mas] (arbitrary)
     assert astrom_cal.platescale == pytest.approx(expected_platescale, abs=0.5)
-    assert astrom_cal.northangle == pytest.approx(expected_northangle, abs=0.3)
+    assert astrom_cal.northangle == pytest.approx(expected_northangle, abs=0.05)
 
     # check that the center is correct within 30 [mas]
```
ajeldorado, line 44: This line should be changed to the following to match the change to the code:
`# check that the center is correct within 3 [mas]`

corgidrp/astrom.py
```diff
@@ -676,33 +683,49 @@ def boresight_calibration(input_dataset, field_path='JWST_CALFIELD2020.csv', thr
...
-    image = dataset[0].data
-
-    # call the target coordinates from the image header
-    target_coordinate = (dataset[0].pri_hdr['RA'], dataset[0].pri_hdr['DEC'])
 
     # find sources in image and match to field
     if field_path == 'JWST_CALFIELD2020.csv':
-        dir_name = os.path.dirname(__file__)
-        full_field_path = os.path.join(dir_name, "tests/test_data", field_path)
+        full_field_path = os.path.join(os.path.dirname(os.path.dirname(__file__)), "tests/test_data", field_path)
```
semaphoreP, line 691: because this file is required for this function, and not just for tests, we need to move this function into corgidrp/corgidrp (same directory as this file). Can you make a folder called like "data" inside of corgidrp/corgidrp, and put JWST_CALFIELD2020.csv there instead? Also update MANIFEST.in so that this file is in the MANIFEST.

corgidrp/recipe_templates/l1_to_boresight.json
```diff
@@ -0,0 +1,92 @@
...
+    "template" : true,
+    "drpconfig" : {
+        "track_individual_errors" : false
+    },
+    "inputs" : [],
+    "outputdir" : "",
+    "steps" : [
+        {
+            "name" : "prescan_biassub",
+            "keywords" : {
```
semaphoreP, line 12: I just made a change to this step in PR #192, so we should adopt it here eventually as well.

tests/e2e_tests/astrom_e2e.py
```diff
@@ -0,0 +1,185 @@
...
+    # create a directory in the output dir to hold the simulated data files
+    rawdata_dir = os.path.join(astrom_cal_outputdir, 'data')
+    if not os.path.exists(rawdata_dir):
+        os.mkdir(rawdata_dir)
+
+    for dark in os.listdir(noise_characterization_path)[:5]:  ## only using the first five frames to save comp time
+        with fits.open(os.path.join(noise_characterization_path, dark)) as hdulist:
+            dark_dat = hdulist[1].data
+            hdulist[0].header['OBSTYPE'] = "AST"
+            rms_noise = np.sqrt(np.mean(dark_dat.flatten())**2)
```
semaphoreP, line 52: I'm not sure this is the rms_noise in the darks (note that taking the sqrt of something that was just squared just cancels out). Probably the best way to get the rms noise is to look at the standard deviation of the dark pixels over time (across multiple files).

tests/e2e_tests/astrom_e2e.py
```diff
@@ -0,0 +1,185 @@
...
+            scaled_image = ((10 * rms_noise) / np.max(amplitudes)) * image_sources[0].data
```
semaphoreP, line 53: so this just sets the brightest source to be at an SNR = 10? If so, we should make a note of that assumption we're making here. 

How did you pick this level?

tests/e2e_tests/astrom_e2e.py
```diff
@@ -0,0 +1,185 @@
...
+    expected_platescale = 21.8
+    expected_northangle = 45
+    target = (80.553428801, -69.514096821)
+
+    # currently checks each file indiidually
+    ## boresight now takes all frames and averages platescale, northangle, ra, dec
+    ## this now need to check only the one astrometric calibration file
+    astrom_cal = data.AstrometricCalibration(os.path.join(astrom_cal_outputdir, 'AstrometricCalibration.fits'))
+
+    assert astrom_cal.platescale == pytest.approx(expected_platescale, abs=0.5)
```
semaphoreP, line 164: useful to add comments like in the unit test that spell out in words what the threshold is for each value

semaphoreP (commented):
A few suggestions and comments. Looking good! Hopefully we get the tests to pass!

maxwellmb (approved):
This all looks good to me. I've run the e2e test on my machine and it passes. The only thing that came to my attention is that the OBSTYPE keyword used, "AST" isn't the one on the confluence header page: "ASTROM". I'm going to approve the PR, but make an issue about that.

Issue #146, opened 2024-07-26
Create recipe template for generating a K-Gain Calibration file


maxwellmb:
**Is your feature request related to a problem? Please describe.**

No recipe template exists for creating a K-Gain calibration file from L1 data. 

**Describe the solution you'd like**

A recipe template needs to be written to generate a K-gain Calibration file from L1 data. 

**Additional context**

In order to test this you will need to get your hands on the appropriate input data from II&T or TVAC and you will quantitatively compare the calibration file to the outputs of the II&T data pipeline. In addition please describe in your own words (~several sentences to a paragraph) what the recipe does and what the inputs and outputs are. Finally, please make sure the appropriate table entry is complete within the Confluence DRP Implementation Document table in Section 2. 

Here's a checklist: 
- [ ] Recipe Written
- [ ] Test Written and passing
- [ ] Table Entry Complete

JuergenSchreiber:
found a bug in calibrate_kgain: OBSTYPE keyword is in pri_hdr (not ext_hdr)

semaphoreP:
Yeah, it looks like OBSTYPE should be in pri_hdr, and we are inconsistent in our step functions so far on which one we assumed it to be in. Good to correct it as we go! 

If you think those variables are useful, should be expose them as tunable parameters in the step function arguments?

semaphoreP:
Gotcha. I think it's ok to make a copy of the data with the right OBSTYPE keyword that we are requesting (assume that the request is successful).

JuergenSchreiber:
I am running into a division by zero error, which comes from frames with a roi with 0 columns. That comes from the kgain_params: 
'colroi1': 1199,
'colroi2': 2000
which means 1 column?
I wonder why that ever worked.

 File "/home/schreiber/workspaceroman/corgidrp/corgidrp/calibrate_kgain.py", line 530, in calibrate_kgain
    local_mean_arrays = [ptc_bin2(frame[rowroi, colroi], 
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/schreiber/workspaceroman/corgidrp/corgidrp/calibrate_kgain.py", line 116, in ptc_bin2
    DN_bin = max_DN / tot_bins
             ~~~~~~~^~~~~~~~~~
ZeroDivisionError: division by zero

semaphoreP:
[...]

PR #191, opened 2024-09-09
Kgain e2e

JuergenSchreiber:
## Describe your changes
- added an end-to-end test for kgain determination inclusive the corresponding walker recipe
- compare the determined kgain and readnoise with the ones determined from TVAC

## Type of change

- [x] I have linted my code


JuergenSchreiber:
For efficiency reasons I have only used a subset of the available TVAC noise data, so I had to reduce the n_cal and n_mean parameters to 3, I think I should change that back to default for the real e2e tests?

semaphoreP:
What is the difference in efficiency? We ideally want to test against a realistic dataset, which includes assessing the computing resources we need. Note that we don't intend to run the e2e tests against every PR like we do for the unit tests.

corgidrp/recipe_templates/l1_to_kgain.json
```diff
@@ -0,0 +1,38 @@
...
+        },
+        {
+            "name" : "correct_nonlinearity",
+            "calibs" : {
+                "NonLinearityCalibration" : "AUTOMATIC"
+            }
+        },
+		{   "name" : "calibrate_kgain",
+		    "keywords" : {
+		        "n_cal" : 3,
```
semaphoreP, line 30: this is going to be the default recipe for the real data, so we should use the parameters we think are best for the real data.

tests/e2e_tests/kgain_e2e.py
```diff
@@ -0,0 +1,125 @@
...
+
+    for file in l1_data_filelist_same_exp:
+        image = data.Image(file)
+        # This should not be necessary anymore after the updates of the OBSTYPE keyword, up to now it is only "SCI"
+        if image.pri_hdr['OBSTYPE'] != 'MNFRAME':
+            image.pri_hdr['OBSTYPE'] = 'MNFRAME'
+            image.save(filename = file)
+        l1_data_filelist_range_exp.append(file)
+
+    walker.walk_corgidrp(l1_data_filelist_range_exp, "", kgain_outputdir, template="l1_to_kgain.json")
```
semaphoreP, line 74: Should we edit `walker.guess_template()` to handle how to process kgain without needing to pass into the template. Or can we not differentiate this data from the nonlinearity data?

tests/e2e_tests/kgain_e2e.py
```diff
@@ -0,0 +1,125 @@
...
+    print("determined read noise", new_readnoise)    
+    
+    diff_kgain = new_kgain - tvac_kgain
+    diff_readnoise = new_readnoise - tvac_readnoise
+    print ("difference to TVAC kgain:", diff_kgain)
+    print ("difference to TVAC read noise:", diff_readnoise)
+    print ("error of kgain:", kgain.error)
+    print ("error of  readnoise:", kgain.ext_hdr["RN_ERR"])
+
+    assert np.abs(diff_kgain) < 1
```
semaphoreP, line 104: are these differences small enough to be tolerated? We should try running on the full dataset and circle back to this.

semaphoreP (commented):
Thanks for doing this Juergen. Fairly straightforward. We should definitely run the e2e on the full set of real data, and assess its performance. We need to also think about if there's a way to automatically identify this set of data (or is it degenerate with the nonlinearity recipe)?

corgidrp/recipe_templates/l1_to_kgain.json
JuergenSchreiber: I leave it to the default values

tests/e2e_tests/kgain_e2e.py
JuergenSchreiber: We could use the determined errors to define a tolerance.

JuergenSchreiber: Understood, so then just a very small number

corgidrp/recipe_templates/l1_to_kgain.json
JuergenSchreiber: These values will be changed to default, that means they do not appear here anymore

semaphoreP:
We recently learned that the Kgain and Nonlin algorithms were updated after the TVAC data products were reduced, and the version we ported over is the updated version. This means that the TVAC data products were generated with the old code unfortunately (this is what Max is referring to in his recent comments in #146). So we have to unfortunately do an extra step: run the TVAC data through the updated engineering pipeline and compare it with our ported version (for nonlin, Sergi confirmed this made the differences disappear). Can you do that? Hopefully, it won't take too long to get running. We can then store the output products on Box for this e2e test to compare against. 

Do you have a copy of the original Kgain code from the updated engineering pipeline? If not, @mygouf is working on releasing it, so you should contact her for the latest status.

mygouf:
> We recently learned that the Kgain and Nonlin algorithms were updated after the TVAC data products were reduced, and the version we ported over is the updated version. This means that the TVAC data products were generated with the old code unfortunately (this is what Max is referring to in his recent comments in #146). So we have to unfortunately do an extra step: run the TVAC data through the updated engineering pipeline and compare it with our ported version (for nonlin, Sergi confirmed this made the differences disappear). Can you do that? Hopefully, it won't take too long to get running. We can then store the output products on Box for this e2e test to compare against.
> 
> Do you have a copy of the original Kgain code from the updated engineering pipeline? If not, @mygouf is working on releasing it, so you should contact her for the latest status.

kgain code is now released and available here: https://github.com/roman-corgi/cgi_iit_drp/tree/main/Calibration_NTR/cal/kgain

JuergenSchreiber:
@mygouf, thanks. Do I have to take the same ROI as @hsergi did for nonlin e2e?

maxwellmb:
> @mygouf, thanks. Do I have to take the same ROI as @hsergi did for nonlin e2e? OK, the ROI is defined in the kgain_params, but the offset_"xxxroi" is not applied.

I think you just want to reproduce as much as possible what the KGain TVAC code is doing, not necessarily what nonlin is doing, so you want to the right ROI from the TVAC KGain code.

JuergenSchreiber:
OK, now it can be reviewed. The test results and performance I have put on confluence

maxwellmb (approved):
I'm going to approve the PR, but there are a couple oustanding issues for which I'm going to make issues: 

- As of right now, `walk_corgidrp` can't guess the template recipe, it must be passed in. We'd like it to be able to guess. 
- The results are not a perfect port to the engineering code. They are very close, but not perfect. We'd like to resolve this.

Issue #203, opened 2024-09-18
detect_cosmic_rays draws k gain from DetectorParams instead of KGain class


kjl0025:
**Is your feature request related to a problem? Please describe.**
Currently, `detect_cosmic_rays` draws the k gain value from `DetectorParams`, but k gain is a calibrated value coming from the corgidrp.data.KGain class.  

**Describe the solution you'd like**
Change the function to draw k gain from corgidrp.data.KGain instead, and adjust tests accordingly.

**Additional context**

JuergenSchreiber:
Should we make it optional and leave the default value in DetectorParams, if KGain cal file not available. Or should we delete the kgain value in DetectorParams?

kjl0025:
Sorry I didn't get back to you sooner on this!  I shared my thoughts on your PR.

semaphoreP:
Added in PR #211

Issue #202, opened 2024-09-17
nonlin and k gain calibrations to utilize the DQ values added from `detect_cosmic_rays`


kjl0025:
**Is your feature request related to a problem? Please describe.**
Nonlin and k gain calibrations should perform cosmic ray correction (and the corresponding II&T code does not.)  However, `detect_cosmic_rays` merely identifies them and modifies the DQ accordingly.  To use this info in these 2 calibrations, a code change would be required. 

**Describe the solution you'd like**
Adjust DRP code to implement the DQ masks due to cosmic rays.  Then the frames would have to be combined with `mean_combine` to appropriately take the mean which takes into account masked pixels in each frame.

**Additional context**
Relatively low priority for now.  There may be other such calibrations this is applicable to?

semaphoreP:
Implemented in PR #439

PR #211, opened 2024-09-24
Cosmics kgain

JuergenSchreiber:
Added KGain calibration file as mandatory input parameter in detect_cosmic_rays

## Type of change

#203

- [ ] I have linted my code


JuergenSchreiber:
I guess I still have to update all the e2e tests with a kgain caldb entry where detect_cosmic_rays is used...

semaphoreP:
Can this be an optional calibration? since detect cosmic rays is a L2a level step, while Kgain is a L2b level calibration product

kjl0025:
Thanks for this, Juergen, but now that I think about this, to Jason's point, we do have an issue (#202 ) about enhancing the k gain and nonlin code to take into account flagged cosmics to improve k gain and nonlinearity calibration.  (The II&T code did not account for cosmic rays, but ideally, that should be done.) So the recipe for making the k gain and nonlin calibrations will eventually use `detect_cosmic_rays`, which would be circular.  So I'm sorry to say this given all your work on this, Juergen, but I think we should actually cancel this PR and leave the code as it is (drawing k gain from `DetectorParams`), and we would just update the `DetectorParams` default k gain value whenever a new k gain calibration is done.  Jason can chime in if he has other thoughts.

maxwellmb:
Do I understand correctly that detect_cosmic_rays _needs_ a KGain value, and its just that right now we're pulling it from DetectorParams, rather than a KGain file?

semaphoreP:
Can we handle it like how we handle Bias offset subtraction with NoiseMaps in prescan_biassub? We leave it as an optional keyword (it needs to be the first keyword for this to work), and the recipe species Kgain to be "AUTOMATIC,OPTIONAL". Basically, the walker will pass in a Kgain if we have one, or pass in None. If None gets passed in, let's just use the Kgain in the DetectorParams.

kjl0025:
> Do I understand correctly that detect_cosmic_rays _needs_ a KGain value, and its just that right now we're pulling it from DetectorParams, rather than a KGain file?

Yes, that is correct.  

And I agree with Jason's proposal.  That's probably better than having to update DetectorParams after a calibration.

corgidrp/l1_to_l2a.py
```diff
@@ -207,7 +208,7 @@ def detect_cosmic_rays(input_dataset, detector_params, sat_thresh=0.7,
 
 
     # Calculate the full well capacity for every frame in the dataset
-    kgain = np.array([detector_params.params['kgain'] for frame in crmasked_dataset])
```
kjl0025, line 210: If kgain is None, use the value from detector_params.  And k_gain would have to be listed before detector_params in the list of arguments so that the "AUTOMATIC, OPTIONAL" works.

maxwellmb:
I don't think I understand why we have a KGain in DetectorParams. What is its relation to the KGain calibration. A backup?

kjl0025:
I think we put it in there initially, before we had started on the calibration script for k gain.

corgidrp/l1_to_l2a.py
JuergenSchreiber: By the way: is there a reason why an array of equal kgain values is made? We can divide an array by a scalar.

JuergenSchreiber:
@maxwellmb: yes, kgain in detetor_params works like a backup if I implement it like desired.

JuergenSchreiber:
To make this work as first parameter, I have to specify detector_params also as an optional parameter

corgidrp/l1_to_l2a.py
```diff
@@ -154,7 +154,7 @@ def prescan_biassub(input_dataset, noise_maps=None, return_full_frame=False,
 
     return output_dataset
 
-def detect_cosmic_rays(input_dataset, detector_params, k_gain, sat_thresh=0.7,
+def detect_cosmic_rays(input_dataset, k_gain = None, detector_params = None, sat_thresh=0.7,
```
semaphoreP, line 157: sorry, I was a bit unclear here: I meant that kgain should be the first optional keyword (i.e., `input_dataset, detector_params, kgain=None, sat_thresh=0.7, ...`)

JuergenSchreiber: Understood, changed back again. I think now it should be ready for review.

corgidrp/recipe_templates/build_trad_dark_full_frame.json
```diff
@@ -19,7 +19,8 @@
         {
             "name" : "detect_cosmic_rays",
             "calibs" : {
-                "DetectorParams" : "AUTOMATIC"
+                "DetectorParams" : "AUTOMATIC",
+				"KGain" : "AUTOMATIC, OPTIONAL"
```
semaphoreP, line 23: for this and other recipes below, can the indentation be at the same level?

corgidrp/l1_to_l2a.py
```diff
@@ -205,9 +206,12 @@ def detect_cosmic_rays(input_dataset, detector_params, sat_thresh=0.7,
 
     crmasked_cube = crmasked_dataset.all_data
 
-
     # Calculate the full well capacity for every frame in the dataset
-    kgain = np.array([detector_params.params['kgain'] for frame in crmasked_dataset])
+    if k_gain == None:
```
semaphoreP, line 210: should use `if k_gain is None` (forget the exact reason, but `is` is preferred to `==`)

corgidrp/l1_to_l2a.py
semaphoreP: agree, not really needed to have an array since we are not assuming to deal with different kgain values right now

semaphoreP (commented):
Looks good. Only a few very very minor comments before we merge this.

corgidrp/recipe_templates/l1_to_kgain.json
```diff
@@ -31,6 +31,13 @@
 				}
             }
         },
+		{
+		     "name" : "detect_cosmic_rays",
+		     "calibs" : {
+		         "DetectorParams" : "AUTOMATIC",
+                 "KGain" : "AUTOMATIC, OPTIONAL"
```
semaphoreP, line 38: one more indent that's off

tests/test_cr_detection.py
semaphoreP, line 1: Sorry, just realized, but is there no test that tests the step function with kgain=None?  If not, we should have a test that does this.

semaphoreP (commented):
Sorry, just realized we may not have a test that tests `kgain=None`, which we should definitely have.

tests/test_cr_detection.py
JuergenSchreiber: We have kgain = None in line 260 in test_crs_zeros_frame(), isn't it?

corgidrp/recipe_templates/l1_to_kgain.json
semaphoreP: might be a tab vs spaces thing?

semaphoreP (approved):

Issue #207, opened 2024-09-20
The kgain e2e tests don't perfectly reproduce the engineering code. 


maxwellmb:
**Describe the bug**
The kgain e2e tests don't perfectly reproduce the engineering code. 

**To Reproduce**
Run the kgain_e2e.py. It gives the following outputs: 
```
determined kgain: 8.81447236619661
determined read noise 132.11032575822475
difference to TVAC kgain: -2.7633803391324818e-05
difference to TVAC read noise: 1.990325758224742
error of kgain: [1.15369296]
error of readnoise: 18.29121579886837
```

**Expected behavior**
We expect the difference to the TVAC kgain and TVAC read nosie to be much smaller. 

**corgidrp version:**
 - main

Issue #204, opened 2024-09-20
discrepancy in CTI e2e when run on different computers


kjl0025:
**Describe the bug**
The CTI e2e test consistently shows exact agreement between the DRP and II&T results on Kevin's computer but consistent disagreement at the ~< 1e-5 level on Max's.  Perhaps their Python versions differ?  OS?  Box reads back the .npy file for the dictionary slightly differently per computer??

**To Reproduce**
The error or lack thereof.

**Expected behavior**
Expected exact agreement.

**corgidrp version:**
(soon-to-be) main branch

**Additional context**

PR #215, opened 2024-10-04
better e2e test agreement with same file order, cosmic detection fix from latest II&T

kjl0025:
## Describe your changes
I updated the trap-pump, synthesized master dark, traditional master dark, and k gain e2e tests to call the public-access repos from cgi_iit_drp to run the data in the script a la II&T, run the data in the script a la DRP, and then compare directly, all using the same file input order.  I also updated the cosmic ray head flagging to only occur in the image area (in keeping with the latest from II&T).  There is a pending PR on cgi_iit_drp with this cosmic flagging update.  

These changes give exact agreement for k gain and trap-pump e2e tests and full-frame agreement to the < 1e-11 level for synthesized and traditional master dark e2e tests. The full-frame agreement is mainly due to the cosmic ray flagging fix.  I would like to have a reviewer run these tests locally to ensure reproducibility, though.  These e2e tests now require importing the II&T code (from cgi_iit_drp), but the casual user of the DRP shouldn't have those as mandatory libraries (i.e., I don't think they need to be listed in requirements.txt).  

The other e2e tests appear to have already a high level of agreement between II&T and DRP.

## Type of change

#207 , #204 

- [ X] I have linted my code
- [ X] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [ X] I have verified that all docstrings are properly formatted and added new documentation, as needed


JuergenSchreiber:
I ran the file_order branch and indeed get no difference:
determined kgain: 9.484170888255141
determined read noise 139.82050344644298
difference to TVAC kgain: 0.0
difference to TVAC read noise: 0.0
error of kgain: [1.06210368]
error of readnoise: 21.151983297402197

But I wonder why we get such a different kgain value now, we had 8.8 in the first run?

kjl0025:
I just pushed a change to the k gain e2e test.  I figured Guillermo ran the data in the order according to the numbers at the end of each file name (i.e., the order in which the data was actually taken), and when I do this here, I get the 8.8145 e-/DN number from before.  Since the script takes frame differences according to the order of the frames in the stacks, changing the order of files input can change the result some.

JuergenSchreiber:
OK, I also assumed to read the list according to the number in the file name and that they have a rising exposure time. Do we have a sorting problem of exposure times? I recognized that the files ....51836 to 51840 have a reduced exposure time of 1.9 sec, while the 5 files before have 19 sec exposure time. That means that the last 5 files would not fit into a scheme of rising exposure times with rising file number.

kjl0025:
No, I think we're good.  And I don't think having increasing exposure time order of stacks matters.  The DRP at least agrees with the ENG results.  I think the result heavily depends on the order of files within a single-exposure-time sub-stack because there are only 5 per sub-stack for this particular dataset, and so there can be non-negligible variance in the frame differences for a single sub-stack, which can affect what k gain value is returned.

semaphoreP:
Hi @kjl0025, maybe we should start a requirements_e2etests.txt? It would be good to list the dependencies for testing in a systematic way. 

I will try to install cgi_iit_drp and run the tests now.

corgidrp/detector.py
```diff
@@ -377,7 +377,7 @@ def imaging_slice(obstype, frame, detector_regions=None):
     return sl
 
 def flag_cosmics(cube, fwc, sat_thresh, plat_thresh, cosm_filter, cosm_box,
-                   cosm_tail, mode='image'):
+                   cosm_tail, mode='image', detector_regions=None, obstype='SCI'):
```
semaphoreP, line 380: doesn't we call it `arrtype` instead of `obstype` since it checks against `arrtype`? to avoid future confusion

corgidrp/l1_to_l2a.py
```diff
@@ -246,14 +251,17 @@ def detect_cosmic_rays(input_dataset, detector_params, sat_thresh=0.7,
     m2 = np.zeros_like(crmasked_cube)
 
     for i in range(len(crmasked_cube)):
+        obstype = crmasked_dataset.frames[i].ext_hdr['ARRTYPE']
```
semaphoreP, line 254: same here, shouldn't be call it `arrtype`?

tests/e2e_tests/kgain_e2e.py
```diff
@@ -10,13 +10,86 @@
 import corgidrp.walker as walker
 import corgidrp.caldb as caldb
 
+try:
+    from cal.kgain.calibrate_kgain import calibrate_kgain
+    import cal
+except:
```
semaphoreP, line 16: just to check, this is to prevent unit tests from failing?

tests/e2e_tests/kgain_e2e.py
```diff
@@ -10,13 +10,86 @@
...
+    pass
+
 thisfile_dir = os.path.dirname(__file__) # this file's folder
 
-tvac_kgain = 8.8145 #e/DN, result from new iit drp code
-tvac_readnoise = 130.12 #e, result from new iit drp code
+# tvac_kgain = 8.49404981510777 #8.8145 #e/DN, result from new iit code with specified file input order
```
semaphoreP, line 21: can these lines be deleted?

tests/e2e_tests/kgain_e2e.py
```diff
@@ -10,13 +10,86 @@
...
+            continue
+        stack_arr.append(dset.all_data)
+    stack_arr = np.stack(stack_arr)
+    pass
+    ordered_filelist = stack_arr_f+stack_arr2_f
+
+    (tvac_kgain, tvac_readnoise, mean_rn_std_e, ptc) = calibrate_kgain(stack_arr, stack_arr2, emgain=1, min_val=800, max_val=3000, 
+                    binwidth=68, config_file=default_config_file, 
+                    mkplot=None, verbose=None)
+
```
semaphoreP, line 92: I would either indicate with comments that we're done with II&T calculation or stick the II&T code into a helper function in this file that you call. The test is a bit long and complicated, so such comments or refactoring would make it a bit easier to follow

tests/e2e_tests/kgain_e2e.py
```diff
@@ -26,29 +99,11 @@ def test_l1_to_kgain(tvacdata_path, e2eoutput_path):
...
-        # This should not be necessary anymore after the updates of the OBSTYPE keyword, up to now it is only "SCI"
-        if image.pri_hdr['OBSTYPE'] != 'MNFRAME':
-            image.pri_hdr['OBSTYPE'] = 'MNFRAME'
-            image.save(filename = file)
-        l1_data_filelist_range_exp.append(file)
-
-    walker.walk_corgidrp(l1_data_filelist_range_exp, "", kgain_outputdir)
-    kgain_file = os.path.join(kgain_outputdir, "CGI_EXCAM_L1_0000051731_kgain.fits")
+
+    walker.walk_corgidrp(ordered_filelist, "", kgain_outputdir, template="l1_to_kgain.json")
```
semaphoreP, line 104: it would be nice if the variables from the II&T section and the variables from the corgidrp section don't intersect except for clear input/output products. It's not clear what has been done to create `ordered_filelist`. I would actually rather do it again here, rather than reusing the code from II&T.

tests/e2e_tests/noisemap_cal_e2e.py
```diff
@@ -41,68 +50,138 @@ def test_noisemap_calibration_from_l1(tvacdata_path, e2eoutput_path):
...
                     input_dataset=mock_input_dataset)
     kgain.save(filedir=noisemap_outputdir, filename="mock_kgain.fits")
     this_caldb.create_entry(kgain)
 
 
     ####### Run the walker on some test_data
-
     template = "l1_to_l2a_noisemap.json"
-    walker.walk_corgidrp(l1_data_filelist, "", noisemap_outputdir,template=template)
+    walker.walk_corgidrp(stack_arr_files, "", noisemap_outputdir,template=template)
```
semaphoreP, line 159: same as for the other e2e tests. It would be good to keep intermediate variables separate between the two. This is to improve test clarity.

tests/e2e_tests/noisemap_cal_e2e.py
```diff
@@ -192,30 +353,37 @@ def test_noisemap_calibration_from_l2a(tvacdata_path, e2eoutput_path):
 
 
     ####### Run the walker on some test_data
-
     template = "l2a_to_l2a_noisemap.json"
-    walker.walk_corgidrp(l2a_data_filelist, "", noisemap_outputdir,template=template)
+    walker.walk_corgidrp(l2a_filepaths, "", noisemap_outputdir,template=template)
```
semaphoreP, line 357: same comment here as for other intermediate variables.

tests/e2e_tests/noisemap_cal_e2e.py
```diff
@@ -249,24 +417,27 @@ def test_noisemap_calibration_from_l2a(tvacdata_path, e2eoutput_path):
...
+    tvacdata_dir = "/Users/kevinludwick/Library/CloudStorage/Box-Box/CGI_TVAC_Data/Working_Folder/"
+    outputdir = thisfile_dir
 
     ap = argparse.ArgumentParser(description="run the l2a->l2a_noisemap end-to-end test")
     ap.add_argument("-tvac", "--tvacdata_dir", default=tvacdata_dir,
                     help="Path to CGI_TVAC_Data Folder [%(default)s]")
     ap.add_argument("-o", "--outputdir", default=outputdir,
                     help="directory to write results to [%(default)s]")
     args = ap.parse_args()
+    args_here = ['--tvacdata_dir', tvacdata_dir, '--outputdir', outputdir]#, '--e2e_flag',False]
```
semaphoreP, line 437: are these just for debugging? should they be removed?

tests/e2e_tests/nonlin_e2e.py
```diff
@@ -140,6 +140,8 @@ def test_nonlin_cal_e2e(
     # We are going to make a new nonlinear calibration file using
     # a combination of the II&T nonlinearty file and the mock headers from
     # our unit test version of the NonLinearityCalibration
+
```
semaphoreP, line 143: minor, but can we undo these two new lines if this PR isn't focused on this file?

semaphoreP (commented):
Ran the e2e tests myself and they seem to be passing. Here's some minor comments in addition to my suggestion above of creating a new requirements file that has the new repos we need for e2e testing.

semaphoreP:
@mygouf Just to confirm with you, running the II&T pipeline and corgidrp on the same data and checking it produces the same results works for you? Previously we were comparing corgidrp against previously ran II&T results (which was harder to reproduce), whereas here we are running it with the II&T pipeline directly.

tests/e2e_tests/kgain_e2e.py
kjl0025: Well, I kind of like having a record of the current results just for comparison when another user tries the test, and I like having the numerical results before the change as well (i.e., the second set of numbers) in case some future change makes a reversion back to those results somehow.  I just made them into more readable comments that don't seem like they are meant to be uncommented.  If you disagree, I can delete the lines.

semaphoreP: ok, as long as they are purposeful, that's fine

tests/e2e_tests/noisemap_cal_e2e.py
kjl0025: Yeah, I used that for debugging so that I could do breakpoints (instead of running from the command line).

kjl0025: But it only works if one uncomments `args = ap.parse_args(args_here)`.

kjl0025:
@semaphoreP , I think I addressed all your comments except for the requirements_e2etests.txt one.  The repos require downloading folders that are not accessible via `repo @git+ web_address.git` in a requirements.txt file.  However, we could point a potential user (via the corgidrp README) to cgi_iit_drp's README for installation instructions of the needed repos?

semaphoreP:
I don't think I needed to download any folders to get this code working. I just did the pip installs of the various subpackages. It seems to me the downloaded data is needed to run tests/scripts, but we're just calling the functions on our own data.

semaphoreP:
Is issue #223 fixed in this PR, or are you planning on doing so later? (Wasn't clear to me)

kjl0025:
I just meant that the code from cgi_iit_drp is not pip-installable from the command line or directly via a web address, so I'm not sure how useful requirements_e2etests.txt would be, so I left it off.

And issue #223 is fixed with this PR.

kjl0025:
Hmm, that command doesn't work for me.  The installation fails.

And good thinking; I added tests for 1024x1024 synthesized and traditional darks.  I also realized the units for k gain and noisemaps were wrong.  I changed "detected EM electrons" to "detected electrons".  I also added a header key for "RN_UNIT" for read noise units in the k gain calibration product, similar to "B_O_UNIT" for the bias offset units in the header of the noisemaps calibration product.

kjl0025:
I also fixed Dark.copy() for the case of input_dataset=None, and I fixed the doc string of calibrate_darks_lsq() in darks.py.

kjl0025:
Okay, requirements_e2etests.txt works for me.  Should be all good to go now.

semaphoreP (approved):

PR #224, opened 2024-10-22
E2e L1 to L2b from scratch

semaphoreP:
## Describe your changes
An e2e test demonstrating we can produce L2b data (including all necessary calibrations from "scratch". 

## Type of change

- [ ] I have linted my code


semaphoreP:
seems like this is not needed right now, so I'm going to close this PR. Can always include this test later if we want.

PR #230, opened 2024-10-25
Fix pickling for dataset

### LINKED ISSUE #229: corgidrp.data class instances must be pickle-able for the docker image to work
**Describe the bug**
Most functions and methods take as the first argument a class instance from corgidrp.data such as Dataset, Image, etc. For them to be input arguments, they have to be instantiated before the function/method call, which means that the docker image must be able to create instances of those class objects. The data objects are created with either "data_or_filepath" inputs. I get no error with direct data passed in, but I always do for the input list of filepaths when using the docker image. The error I get is `TypeError: cannot pickle 'BufferedReader' instances` when the docker image or its wrapper tries to pickle the class instance.

**To Reproduce**
This can be reproduced without the docker image by creating a class instance of say corgidrp.data.Dataset using a list of filenames as the input and then trying to pickle the object.

**Expected behavior**
The data class objects need to be pickle-able in order to use the docker image of corgidrp.

**corgidrp version:**
v1.0

**Additional context**
As Al Niessner put it, "The original agreement/pact was that all object that are passed via function arguments need to be pickle-able for dorba to work." 

I tried adding the keyword argument `memmap=False` as described [here](https://docs.astropy.org/en/latest/io/fits/appendix/faq.html#id11) in corgidrp/data.py whenever astropy.io.fits.open() is called, but that didn't fix it. I thought it might because it makes the data be loaded rather than waiting to load it, which of course can't work when you try to pickle it. I also tried setting `os.environ['MKL_NUM_THREADS'] = "1"` at the top of corgidrp/data.py to see if that would disable multithreading, but it did not change the issue with BufferedReader. So I don't know what else can be done to avoid or disable BufferedReader, but we have to figure out a way for this to work.

semaphoreP:
## Describe your changes

Copy hdulist after read, so file resources are released. Added unit test to check datasets read in from disk can be pickled. 

## Type of change

Closes #229 

- [x] I have linted my code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -81,4 +81,5 @@ tests/e2e_tests/*/*.json
 figures/
 tests/figures/
 tests/e2e_tests/*/*.png
+tests/e2e_tests/*_output
 tests/test_data/simastrom/guesses.csv
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -312,10 +312,12 @@ class Image():
 
 
                 if input_hdulist is not None:
-                    self.hdu_list = input_hdulist
+                    this_hdu_list = [hdu.copy() for hdu in input_hdulist]
                 else: 
                     #After the data, err and dqs are popped out, the rest of the hdulist is stored in hdu_list
-                    self.hdu_list = hdulist.copy()
+                    this_hdu_list = [hdu.copy() for hdu in hdulist]
+                self.hdu_list = fits.HDUList(this_hdu_list)
+                
 
             # parse the filepath to store the filedir and filename
             filepath_args = data_or_filepath.split(os.path.sep)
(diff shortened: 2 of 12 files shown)


semaphoreP:
@ajeldorado can you check this fix works for you? If true, it ended up being a very simple fix.

ajeldorado (changes requested):
Could you also add tests for pickling all the data.py class instances to make sure they all work?

semaphoreP:
Yup, just added pickling tests for all the calibration files (the dataset and image class are both tested with the original test).

ajeldorado (approved):
I can't do a complete check while on vacation, but from what I can see this looks good.

maxwellmb (approved):

Issue #81, opened 2024-05-09
when using the Image.copy() function for sub-classes of Image (e.g. Darks) the sub-class type is not copied


maxwellmb:
This was brought to my attention by the fact that @JuergenSchreiber created a new copy function for the KGain class in PR #80. 
You can see this like this: 
`import corgidrp.mocks as mocks
import corgidrp.detector as detector

dark_dataset = mocks.create_dark_calib_files()
dark_frame = detector.create_dark_calib(dark_dataset)
dark_frame_copy = dark_frame.copy()
print(type(dark_frame))
print(type(dark_frame_copy))`

Gives the output: 
<img width="215" alt="image" src="https://github.com/roman-corgi/corgidrp/assets/18196772/0c7f278e-3ea9-47ae-b9e9-01f73f560093">

Hopefully there's a way we can generalize this without the need for new copy methods for each new calibration datatype.

semaphoreP:
unfortunately the best solution is to write a new version of the copy() class, unless we can guarantee that all constructors have the same signature (all `__init__()` functions take the same constructor). There is a hack we can do, but I can't guarantee it'll always work

semaphoreP:
Only if that function actually gets used.

maxwellmb:
I suppose. It's low priority, but I still think something we want eventually.

maxwellmb:
I made a new low priority tag

JuergenSchreiber:
If we want to use the copy module, we could save writing copy functions.
I tested it on a kgain instance and it seems to work:
import copy
#reference copy
kgain_copy = copy.copy(kgain)
#deep data copy
kgain_copy = copy.deepcopy(kgain)

What do you think?

semaphoreP:
Good idea!

semaphoreP:
Sounds good!

maxwellmb:
Closed with Merge PR #217

maxwellmb:
So that I understand, @JuergenSchreiber, your PR #217, updated the readme, but it doesn't affect existing implementations, right?

JuergenSchreiber:
@maxwellmb yes, correct

maxwellmb:
Should we just replace the innards of the existing `copy` functions to use `copy.deepcopy()` instead?

semaphoreP:
sounds like a good idea to me

JuergenSchreiber:
[...]

PR #248, opened 2024-11-06
Copy replace

JuergenSchreiber:
## Describe your changes
In the copy function of Image I replaced the single copies by copying the whole class using the copy.copy and copy.deepcopy functions of the python copy module. The copy functions of the subclasses of Image could be deleted.

## Type of change
- bug fix
- writing of copy functions for calibration subclasses is not necessary anymore. 

#81

- [ ] I have linted my code


semaphoreP:
Does this need to go into `main` (v1.1?), or should this be saved for `develop` (v2.0)?

JuergenSchreiber:
> Does this need to go into `main` (v1.1?), or should this be saved for `develop` (v2.0)?

Hmm, I have no strong opinion, maybe a question for Max? It is low priority.

maxwellmb:
I think develop is fine

semaphoreP (approved):

PR #249, opened 2024-11-06
Obstype switch

semaphoreP:
## Describe your changes

 * use VISTYPE instead of OBSTYPE in walker.py
 * walker.py can also now queue multiple recipes that process the same input data 
 * the above is used to run both kgain and nonlin recipes from the same L1 PUPILIMG dataset (this is not yet reflected in the e2e tests)

## Type of change

- [x] I have linted my code


corgidrp/mocks.py
```diff
@@ -988,7 +990,7 @@ def create_astrom_data(field_path, filedir=None, subfield_radius=0.02, platescal
     # load as an image object
     frames = []
     prihdr, exthdr = create_default_headers()
-    prihdr['OBSTYPE'] = 'ASTROM'
+    prihdr['OBSTYPE'] = 'BORESITE'
```
maxwellmb, line 993: Should this be VISTYPE?

corgidrp/walker.py
```diff
@@ -53,36 +53,51 @@ def walk_corgidrp(filelist, CPGS_XML_filepath, outputdir, template=None):
...
+    if isinstance(recipes, list):
+        # if multiple recipes
+        for recipe in recipes:
+            run_recipe(recipe)
+    else:
+        # process single recipe
+        run_recipe(recipes)
 
-    return recipe
+    return recipes
```
maxwellmb, line 86: I think we maybe want a unit test for when there's two recipes, no?

tests/test_kgain_cal.py
```diff
@@ -122,7 +122,7 @@ def count_contiguous_repeats(arr):
                                divide_em=True)
         image_sim.ext_hdr['DATETIME'] = time_stack_arr0[t+j*exp_repeat_counts[j]]
         # OBSTYPE has no KGAIN value, but NONLIN
-        image_sim.ext_hdr['OBSTYPE'] = 'NONLIN'
+        image_sim.pri_hdr['OBSTYPE'] = 'NONLIN'
```
maxwellmb, line 125: This should be VISTYPE, right?

tests/e2e_tests/nonlin_e2e.py
```diff
@@ -164,7 +164,7 @@ def test_nonlin_cal_e2e(
 
     # Run the walker on some test_data
     print('Running walker')
-    walker.walk_corgidrp(nonlin_l1_list, '', e2eoutput_path)
+    walker.walk_corgidrp(nonlin_l1_list, '', e2eoutput_path, "l1_to_l2a_nonlin.json")
```
maxwellmb, line 167: I assume this is a temporary fix, since a number of things will change once Sergi finishes with the new filesorting thing, right?

maxwellmb (commented):
A few minor things. The main thing is that I think we want a test for the new dual recipe processing.

tests/e2e_tests/nonlin_e2e.py
semaphoreP: we need a new e2e test eventually to test walk_corgidrp() producing both a kgain and nonlin at the same time. I don't think it should be in this test, because the current kgain and nonlin e2e tests focus on checking consistency with TVAC. And I don't think mixing the two tests together is a good idea.

tests/test_kgain_cal.py
semaphoreP: No, this is a unit test. We assume that Sergi's step has already been run and populated OBSTYPE into the header. (the step function still needs OBSTYPE, the DRP now produces it for the step function).

tests/e2e_tests/nonlin_e2e.py
maxwellmb: Ok, sounds good. Should we create an issue for the new e2e test that we'll need, or is it implied that it should be included in Sergi's PR?

semaphoreP: new issue. we probably won't get to it today. I'm thinking it's less necessary since the functionality is there, but we don't need to have all the automation done this time around.

maxwellmb: works for me as long as the issue is created

corgidrp/walker.py
semaphoreP: I added a unit test to a the autogen_recipe() function, as it wasn't straigthforward to test this function, and I didn't want to figure it out right now.

maxwellmb: I made a follow Issue #255

maxwellmb (approved):

PR #258, opened 2024-11-22
Northup task

taichiuyama:
## Describe your changes
added northup function in l3_to_l4.py and tests/test_northup.py with a mock data 'tests/test_data/mock_northup.fits'

## Type of change

- [ ] I have linted my code
- [] I have verified that all docstrings are properly formatted and added new documentation, as needed


tests/test_northup.py
```diff
@@ -0,0 +1,72 @@
...
+    ax1.set_title(f'Derotated Mock Data\n by {-roll_angle}deg counterclockwise')
+    ax1.imshow(im_derot,origin='lower')
+
+    outdir = 'test_data/'
+    outfilename = 'compare_northup.png'
+
+    if not os.path.isdir(outdir):
+        os.mkdir(outdir)
+    
+    plt.savefig(outdir+outfilename)
```
semaphoreP, line 66: please use os.path.join instead of your own path concatenation

tests/test_northup.py
```diff
@@ -0,0 +1,72 @@
...
+
+    outdir = 'test_data/'
+    outfilename = 'compare_northup.png'
+
+    if not os.path.isdir(outdir):
+        os.mkdir(outdir)
+    
+    plt.savefig(outdir+outfilename) 
+    
+    print(f"Comparison figure saved at {outdir+outfilename}")
```
semaphoreP, line 68: It would be good to do an `assert` to check that the image was rotated as you specified, rather than just a a plot, since we want this test to be automatic, and not requiring manual inspection.

semaphoreP (commented):
Nice work putting this together! I have a few comments below. I think the main thing is to figure out how to rotate the DQ mask for effectively. Maybe we can chat on slack on this.

taichiuyama:
@semaphoreP do you know why the CI test got filenot found error? I uploaded the mock file at test_data but the unit test doesn't seem to find it.

semaphoreP:
@taichiuyama the unit tests can be run from an arbitrary directory (arbitrary current working directroy), so we have to use the location of the file itself to figure out the path to test data. E.g., here's what's in the test_astrom.py code: `    field_path = os.path.join(os.path.dirname(__file__), "test_data", "JWST_CALFIELD2020.csv")`

taichiuyama:
@semaphoreP thank you Jason, now it's solved.

semaphoreP:
Great! I think there's one comment from my last PR that's not addressed, which is to use assert statements in the test to quantitatively verify the function is correct. We want something more automated than just looking at a plot (Which is very manual).

semaphoreP:
For example, you can assert that pixel values after rotation are now a different expected value.

taichiuyama:
@semaphoreP oh sorry I missed that, OK will do

semaphoreP:
Great! It would be great if you can check the DQ maps post rotation as well, since we spent a good amount of work on that.

taichiuyama:
@semaphoreP I added a pix cluster with value=1 in DQ and used that location to test rotation

semaphoreP (approved):

Issue #237, opened 2024-10-29
Compute apparent magnitude of source in CGI band


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Given the magnitude of the star in a standard filter band and the star's spectra type, compute the apparent magnitude of the source in the CGI band and store it in the header

**Describe the solution you'd like**
The above in a step function. 

**Additional context**
Need to figure out where the CGI filter profiles are, but they should already be documented

JuergenSchreiber:
I guess you mean something like a color correction factor?

semaphoreP:
Yep, exactly it.

semaphoreP:
@JuergenSchreiber, I was just looking over this issue more closely, and I'm not sure #259 fully solves this issue, because we computed a color correction, but we didn't go all the way and compute the apparent magnitude in the CGI band. Is that correct?

JuergenSchreiber:
Well, we will compute the flux in the CGI bands, and I think we could convert the units to magnitudes. But that might not be trivial since we do not have the standard UBV filters. Since the goal is the calibration of the ND filter, so relative fluxes is good enough, I think the final unit is not that important as long as we stay at the same all the time. Is this reasonable?

semaphoreP:
I think we want vega magnitudes, so given a filter profile and a vega spectrum, it should be doable.

I only emphasized magnitudes, because that's what's listed in the design document. I think for the science data, we do want to be able to calculate the apparent magnitude of both the star and any point sources around it (not just the flux ratio). So we do need some way to convert to fluxes (not just relative fluxes to the star).

JuergenSchreiber:
Ok, so please keep this issue open, I will write it in a header keyword

PR #259, opened 2024-11-22
Color correction of a source sed vs a reference sed

JuergenSchreiber:
## Describe your changes
A pipeline step that adds a reference wavelength (pivot wavelength, independent of source) of the used color filter and the color correction factor to the ext_hdr headers of the frames in a dataset (keyword names tbd and approved).
I added a fluxcal module, that contains the needed functions to calculate the color correction, which is based on  Gorden et al. 2022, Astronomical Journal. For absolute flux calibration the calibration factor has to be divided by the dimensionless color correction factor of a defined filter band. The color correction is need to account for different shapes of the SEDs within a band pass. 
I assume that the file of the CALSPEC standard star and the SED model of the observed star as well is given by the user in the CALSPEC format and unit. The csv files of the transmission curves of the color filters I have put in the data/filter_curves directory.  We could make a cal file out of them, but I think that is not necessary since they won't be re-measured in space and do not change anymore. To test the code I have put a "small" calspec file in the test_data directory. I have put the pipeline step in l2b_to_l3.py which is tbc.

## Type of change

#237

- [x] I have linted my code


corgidrp/l2b_to_l3.py
```diff
@@ -30,6 +32,44 @@ def divide_by_exptime(input_dataset):
...
+def determine_color_cor(input_dataset, calspec_filepath, source_filepath):
+    """
+    determine the color correction factor of the observed source
+    at the reference wavelength of the used filter band and put it into the header.
+    We assume that each frame in the dataset was observed with the same color filter.
+    
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (L2b-level)
+        calspec_filepath (str): file name of the known reference flux (usually CALSPEC)
+        source_filepath (str): file name of the flux model of the observed source in
```
semaphoreP, line 44: Do you envision that we'll have a library of possible star spectra that we download alongside the pipeline, and we manually pick the one we should use in the future? 

As a follow-up, is it useful to allow the user to specify a blackbody temperature to instead use a simple blackbody model for this?

semaphoreP (commented):
Thanks for putting this together. A few quick comments

corgidrp/l2b_to_l3.py
```diff
@@ -30,6 +32,44 @@ def divide_by_exptime(input_dataset):
...
 
+def determine_color_cor(input_dataset, calspec_filepath, source_filepath):
+    """
+    determine the color correction factor of the observed source
+    at the reference wavelength of the used filter band and put it into the header.
+    We assume that each frame in the dataset was observed with the same color filter.
+    
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (L2b-level)
+        calspec_filepath (str): file name of the known reference flux (usually CALSPEC)
```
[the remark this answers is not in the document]
JuergenSchreiber: Well, my imagination was, that we need to observe at least 2 stars (I guess these will be mostly cal standards with available calspec files, or at least known stellar models in calspec style format) one with and one without the ND filter in different color filter bands. I think there might well be different reference stars needed for the different bands, visibility times etc, so this should be flexible. I think the star observed without ND filter is the reference, and the observation of a star with ND filter will be calibrated on that, but it could of course also be vice versa.

corgidrp/l2b_to_l3.py
JuergenSchreiber: I guess we should ask this the calibration group, we could also try to access the corresponding calspec data via automatic download. I doubt that a simple BB model is good enough for the flux calibration on sky?

corgidrp/l2b_to_l3.py
```diff
@@ -30,6 +32,44 @@ def divide_by_exptime(input_dataset):
 
     return input_dataset.copy()
 
+def determine_color_cor(input_dataset, calspec_filepath, source_filepath):
```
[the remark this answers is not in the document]
JuergenSchreiber: You are right, I just had a look at the FDD

corgidrp/l2b_to_l3.py
semaphoreP: I like the idea of automatic download of calspec files as needed for this. Probably don't need to do it for this PR, but we can chalk it up to future work.

corgidrp/l4_to_tda.py
```diff
@@ -0,0 +1,73 @@
...
+    # calculate the flux from the user given CALSPEC file binned on the wavelength grid of the filter
+    flux_ref = fluxcal.read_cal_spec(calspec_filepath, wave)
+    # we assume that the source spectrum is a calspec standard or its 
+    # model data is in a file with the same format and unit as the calspec data
+    source_sed = fluxcal.read_cal_spec(source_filepath, wave)
+    #Calculate the color correction factor
+    k = fluxcal.compute_color_cor(filter_trans, wave, flux_ref, lambda_ref, source_sed)
+    
+    # write the reference wavelength and the color correction factor to the header (keyword names tbd)
+    history_msg = "the color correction is calculated and added to the header {0}".format(str(k))
```
rzellem, line 36: @JuergenSchreiber and @semaphoreP - so the color correction might change slightly with different star spectral types - is it worth putting in the cal star's spectral type into the header as well or leave that as an exercise for the user?

Also: just want to be abundantly clear on this - the true purpose of these absolute flux calibrations is to calibrate out the ND filter. CGI will be delivering the flux of the planet relative to the flux of the star - this greatly simplifies our science calibration error budget as it puts everything into relative terms, instead of absolute.

We need the ND filter to observe the unocculted flux of the exoplanet host star. We observe our planet/disk without the ND filter. So, the difference between observations of the star vs. those of the planet/disk is that ND filter. We need to effectively measure the ND filter's response in order to "convert" our unocculted star measurements with the ND filter to remove the ND filter's response. Also, the ND filter is in a focal plane, so depositions can occur on the ND filter and change the spectral response- and this can vary by the position on the ND filter.

The way we came up with removing/correcting for the ND filter is to perform absolute flux calibrations with and without the ND filter in place. At the end of the day, we just need to remove that silly ND filter from our measurements of the unocculted host star.

corgidrp/fluxcal.py
```diff
@@ -0,0 +1,164 @@
...
+    lambda_nm = tab['lambda_nm'].data #unit nm
+    transmission = tab['%T'].data
+    return lambda_nm * 10 , transmission/100.
+
+def read_cal_spec(calspec_filename, filter_wavelength):
+    """
+    read the calspec flux density data interpolated on the wavelength grid of the transmission curve
+    
+    Args:
+        calspec_filename (str): file name of the CALSPEC fits file
```
rzellem, line 56: For the current target list, please see Tables 1 and 2 here: https://www.spiedigitallibrary.org/conference-proceedings-of-spie/12188/2628627/Absolute-flux-calibrations-for-the-Nancy-Grace-Roman-Space-Telescope/10.1117/12.2628627.full?tab=ArticleLink

We are working on getting the final target tables approved by JPL for sharing - stay tuned! But in the meantime, you can grab those CALSPEC files for those targets.

rzellem (changes requested):

corgidrp/l4_to_tda.py
```diff
@@ -0,0 +1,73 @@
+# A file that holds the functions that transmogrify l4 data to TDA (Technical Demo Analysis) data 
+import corgidrp.fluxcal as fluxcal
+
+def determine_color_cor(input_dataset, calspec_filepath, source_filepath):
+    """
+    determine the color correction factor of the observed source
+    at the reference wavelength of the used filter band and put it into the header.
+    We assume that each frame in the dataset was observed with the same color filter.
```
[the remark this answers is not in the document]
JuergenSchreiber: @rzellem - up to now I assume that all frames in a dataset use the same color filter and get it from the header keyword "CFAMNAME"...otherwise we must sort the dataset accordingly.

corgidrp/l4_to_tda.py
JuergenSchreiber: @rzellem - thanks for further clarifying. I am aware of that and that simplifies things, e.g. we do not have to care about units, unit conversions etc. My way forward would be to determine the calibration factor from a flux standard observed without ND, then observe a suitable other (ND) star with known flux in the same band with the ND. Applying the color corrected calibration factor to the DNs of that ND star would yield the measured ND flux. Dividing the measured ND flux by the known ND star flux would yield the ND transmission (assuming a sufficient constant ND transmission within the color filter band).
The spectral type of the e.g. calspec stars is listed in the calspec table on stsci.edu, so I think the target name of the observed stars in the header would be sufficient enough.

corgidrp/l4_to_tda.py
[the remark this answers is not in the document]
rzellem: @JuergenSchreiber - thanks! I'm just being a bit paranoid that we might have occasionally user or pipeline error and data will get mixed in...but maybe I'm just being too paranoid here....

corgidrp/l4_to_tda.py
rzellem: @JuergenSchreiber - apologies for not making this clearer sooner! There's certainly nothing wrong with what you currently have - and honestly, it will probably be the best thing in the long term in case someone who does an observation of a system (maybe a disk?) does want absolute flux units - so I think your code is ultimately for the better, thank you!

We found that in order to achieve our allocation in the Science Calibration error budget, that we had to observe multiple stars to increase our SNR - a single CALSPEC standard did not offer enough precision. If you take a look at Linsey's poster (and we're working on getting you her code - but it needs to be blessed by JPL first), she had a few dim stars (~4) to observe without the ND filter and then another ~6-10 bright ones to observe with the ND filter. By combining all of these measurements together, she was able to determine the correction factor with the required precision.

corgidrp/fluxcal.py
JuergenSchreiber: @rzellem: I implemented the possibility in the pipeline step to either input the name of a calspec star or the file path to a fits file. Note: up to now the names are taken from the list of your poster and put in a dictionary associating the according current calspec fits names, and must match exactly. Can/should be made more flexible in future.

corgidrp/fluxcal.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    datadir = os.path.join(os.path.dirname(__file__), "data", "filter_curves")
+    filters = os.path.join(datadir, "*.csv")
+    filter = dataset[0].ext_hdr['CFAMNAME']
+    filter_names = os.listdir(datadir)
+    for name in filter_names:
+        if filter in name:
+            return os.path.join(datadir, name)
+        else:
+            pass
```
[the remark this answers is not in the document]
semaphoreP: Hi @rzellem I don't think we should pause this PR in the interest is continuing to move along.

corgidrp/l4_to_tda.py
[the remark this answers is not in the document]
semaphoreP: instead of updating via the history entry, you can just update the headers in a for loop. that way (I assume it's not computationally intensive to compute the color correction, so it's ok if we repeat it a bunch of times).

corgidrp/l4_to_tda.py
semaphoreP: @rzellem, I think we can just re-run the abs phot calib and/or ND filter calib on multiple reduced datasets after we process all the individual datasets first (e.g., the color corrections are already in the headers). So I don't think this step precludes doing multiple stars like you mentioned.

semaphoreP: However, I think if the code in this function is quick to run, we might as well run it on a per-file basis (see my above comment).

corgidrp/fluxcal.py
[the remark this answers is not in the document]
semaphoreP: @JuergenSchreiber I think you still should change the error throwing ehavior

corgidrp/fluxcal.py
```diff
@@ -0,0 +1,229 @@
...
+    Returns:
+        str: file path
+    """
+    if star_name not in calspec_names:
+        raise ValueError('{0} is not in list of anticipated standard stars {1}, please check naming'.format(star_name, calspec_names.keys()) )
+    fits_name = calspec_names.get(star_name)
+    # TODO: be flexible with the version of the calspec fits file, so essentially, the number in the name should not matter
+    fits_url = calspec_url + fits_name
+    try:
+        calspec_dir = os.path.join(os.path.dirname(__file__), "data", "calspec_data")
```
semaphoreP, line 49: because we can't assume we always have write access to the folder in which this file resides on a user's computer, we should download new files into the .corgidrp folder that stores the caldb and config files: `os.path.dirname(corgidrp.config_filepath)` (I need to update this so we can directly access the path to the .corgidrp folder instead of having to do dirname()). 

You are free to make a folder called calspec_data in ~/.corgidrp though!

corgidrp/fluxcal.py
semaphoreP: let's make it a future work to double check this list once we have that.

semaphoreP (commented):
One more comment to change the directory where you are downloading the calspec files, and also you should still fix the error catching comment from last time.

corgidrp/l4_to_tda.py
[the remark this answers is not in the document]
JuergenSchreiber: My naive assumption was, that we have an input dataset of the source star in a defined color filter band, and update each frame header with the same K-factor.  If we would have different star observations e.g. in different filters in one dataset, we would indeed need to run this step for each single frame. But I think we only have one observation in one dataset, isn't it?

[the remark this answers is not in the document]
rzellem: For images taken without the ND filter, we have only one exposure per star. However, for images taken *with* the ND filter, we will do a pointing grid pattern of multiple exposures - we need to do this to ensure that we align the star with the ND filter's sweet spot (as there's some pointing instabilities that will prevent us from directly aligning with the sweet spot) - so we will select only 1 of the bunch that best aligns with the sweet spot.

corgidrp/l4_to_tda.py
JuergenSchreiber: @semaphoreP, I agree. I assume that we run the complete pipeline to abs flux cal for each observation of one star in one filter band. And for the color correction I assume, that we know already what the reference star is (should be in the abs flux cal file). For the reference star observation itself and determination of the calibration factor which is a cal file e2e processing no color correction is needed.  In the cal file there should be in the end  calibration factors of reference stars with and without ND filter. Well, I think we will see how to organize this in the course of the development. 
For the ND filter calibration, it is getting more complex, in the end we will have some combinations of bright with dim cal standard stars reduced, on which we determine the filter transmission. We have to see how we organize this within the pipeline. Inside a dataset, or set up a dictionary, maybe something like the associations in JWST pipeline at this higher level stage of processing, just brainstorming.

corgidrp/l4_to_tda.py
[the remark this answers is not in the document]
semaphoreP: I'm not sure we know all the possible ways we want to use the pipeline for flux calibration (e.g., processing multiple stars together seems unlikely, but you never know). I'm fine punting this until later. I think putting this in a for loop later is not a problem if we determine that's the workflow we want to do.

corgidrp/l4_to_tda.py
semaphoreP: This makes sense to me. I agree that it'll be best to write the abs flux cal (and then the ND calib), and see if there's anything missing in this function as we do that. Shall we wrap up this PR and move on to that?

corgidrp/fluxcal.py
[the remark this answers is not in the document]
JuergenSchreiber: Is this ok now?

corgidrp/fluxcal.py
JuergenSchreiber: Ok, I did that

corgidrp/fluxcal.py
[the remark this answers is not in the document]
semaphoreP: I don't think this is valid python: `"cannot access CALSPEC archive web page and/or download {0}".fits_name`. You probably need to use the format() function.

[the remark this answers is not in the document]
JuergenSchreiber: Indeed, good catch

semaphoreP (approved):
I think this looks good. Once the full target list released, it would be good to re-check the target list against it (@rzellem, can you make an issue once that full target list is released?).

rzellem (approved):

PR #265, opened 2024-12-09
calculate apparent magnitude, issue #237

JuergenSchreiber:
## Describe your changes
added a l4_to_tda pipeline step to calculate the apparent Vega magnitude in a color filter and add it to the header

## Type of change

#237

- [x] I have linted my code

### DIFF
--- a/corgidrp/l4_to_tda.py
+++ b/corgidrp/l4_to_tda.py
@@ -2,7 +2,7 @@
 import corgidrp.fluxcal as fluxcal
 import numpy as np
 
-def determine_app_mag(input_dataset, source_star):
+def determine_app_mag(input_dataset, source_star, scale_factor = 1.):
     """
     determine the apparent Vega magnitude of the observed source
     in the used filter band and put it into the header.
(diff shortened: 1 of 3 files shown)


corgidrp/l4_to_tda.py
```diff
@@ -1,5 +1,50 @@
 # A file that holds the functions that transmogrify l4 data to TDA (Technical Demo Analysis) data 
 import corgidrp.fluxcal as fluxcal
+import numpy as np
+
+def determine_app_mag(input_dataset, source_star):
```
semaphoreP, line 5: would it be useful to offer the ability to apply a scale factor to the source_star spectrum? I could see cases where a science target doesn't have a nicely created CALSPEC file, but we know it's spectral type is basically the same as a CALJSPEC source, but the fact that it's X times fainter. This could allow for a quick calibration without needing to generate a brand new CALSPEC file.

semaphoreP (approved):
Looks good generally. Nice work! I just have one question that I thought could make the function a bit more flexible.

corgidrp/l4_to_tda.py
JuergenSchreiber: Well, I could add scale_factor as optional parameter. At least it doesn't hurt.

semaphoreP: Looks good. thanks!

PR #268, opened 2024-12-10
Make sure to set the DQ when dividing by a zero in a flat-field 

maxwellmb:
## Describe your changes

If a different flatfield is used for flat field division than was used to create the bad pixel map and the new one has 0s in it, we might introduce infs when we divide. This fix makes sure that the DQ gets set when this happens. 

## Type of change

- [x] I have linted my code

### DIFF
--- a/corgidrp/l2a_to_l2b.py
+++ b/corgidrp/l2a_to_l2b.py
@@ -133,7 +133,7 @@ def flat_division(input_dataset, flat_field):
     where_zero = np.where(flat_field.data == 0)
     flatdiv_dq = copy.deepcopy(flatdiv_dataset.all_dq)
     for i in range(len(flatdiv_dataset)):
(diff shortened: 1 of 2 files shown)


corgidrp/l2a_to_l2b.py
```diff
@@ -128,6 +129,12 @@ def flat_division(input_dataset, flat_field):
     #Divide by the master flat
     flatdiv_cube = flatdiv_dataset.all_data /  flat_field.data
 
+    #Find where the flat_field is 0 and set a DQ flag: 
+    where_zero = np.where(flat_field.data == 0)
+    flatdiv_dq = copy.deepcopy(flatdiv_dataset.all_dq)
```
semaphoreP, line 134: do you need to do this, when the dataset is already copied above?

semaphoreP (approved):
Looks good. just one question on memory efficiency.

mygouf:
Looks good on my side.

Issue #236, opened 2024-10-29
Combine images


semaphoreP:
**Is your feature request related to a problem? Please describe.**
We need a step function to combine images. 

**Describe the solution you'd like**
Given an input dataset, combine images to output a single Image. Error and DQ should also be combined. Combination can be either mean or median. Can look at combine.py which already has some code to do this for subexposures.

Issue #124, opened 2024-07-04
potential improvements in accuracy of err 


kjl0025:
Here are some ways to improve the err estimation for various pipeline steps (which isn't already covered by another issue to my knowledge).  Perhaps not all of these enhancements are feasible and some may be overly rigorous, but I wanted at least to have things documented:
1. Bias subtraction from prescan:  For each row's err, take the median absolute deviation (MAD=median( |x- median(x)| ), where median(x) is the bias value that was subtracted for the row.  Currently, the standard deviation is instead taken, but the MAD is more accurate for the error in a median vs a mean. The MAD can combine in quadrature just like a standard deviation, too.
2. Nonlinearity correction:  The err would the error propagation involving the uncertainty of the corrective multiplicative factor.  I believe the uncertainty in the multiplicative factors would not be too hard to obtain in the calibration function for the nonlinearity correction map.
3. DN to e-:  Err would be the error propagation for multiplication by k gain involving the uncertainty of the k gain factor, and this can be obtained from the k gain calibration function.
4. Division by EM gain:  Err would be the error propagation for multiplication by EM gain involving the uncertainty of the EM gain, but this would have to be obtained from the function that does the EM gain calibration, which is not open to CPP and probably wouldn't be modified to return the uncertainties unless we argued it was necessary.
5. Flat division: Err would be the error propagation for division involving the uncertainty for each pixel in the flat (which could include the initial error bars from the raw frames that made the flat as well).

semaphoreP:
I'm classifying this as low priority, as errors aren't actually a main deliverable. 

Btw, I don't think MAD is more accurate for the error of the median than the standard deviation, especially when you are assuming 2nd-order statistics.

kjl0025:
We could also pretty easily get the err for the PTC curve (error bars in x (signal) and y (noise) directions) that comes from the k gain calibration function.

JuergenSchreiber:
[...]

PR #271, opened 2024-12-17
Combine images

bensutlieff:
## Describe your changes
Tasks related to issue Combine images #236 :
- Added additional tests to test_combine.py to confirm that combine.py behaves as expected
- Verified the error propagation calculation in combine.py
- Fixed bugs and typos in combine.py
- Fixed a couple of typos in readme.md

## Type of change

Combine images #236

- [x] I have linted my code


kjl0025:
These are all supposed to be "identical" frames, right?  My suggestion would be to implement the methodology for computing the combined `err` and combined `dq` from corgidrp.darks.build_trad_dark()`, which uses `corgidrp.darks.mean_combine()`.  `build_trad_dark()` in lines 215-234 masks pixels that have a non-zero DQ value in the computation of the error, and if a pixel has more than half of its values flagged in the stack, the combined DQ value for that pixel is flagged as unreliable (with a value of 256).  The error includes "systematic error" (coming from the `err` from each frame in the stack) as well as the statistical error (coming from the variation from frame to frame).  Combining the "systematic" error with the statistical error in this way assumes `corgidrp.l2a_to_l2b.add_photon_noise()` has not been applied to the frames' `err` since that adds in essentially an estimate of the statistical error per frame, whereas computing it as in `build_trad_dark()` computes that statistical error over the stack directly.  A discussion on that point about `add_photon_noise()` can be found here:  https://github.com/roman-corgi/corgidrp/issues/124

kjl0025:
Yes, issue #124 refers to our previous discussion about assigning a more accurate statistical noise compared to what is assigned per frame in `add_photon_noise()` once a dataset is combined into a single frame.  One option is to leave `add_photon_noise()` out of a recipe if the intended end product will be a combined frame.  In the rare case that only 1 frame is taken for a dataset, `add_photon_`noise()` would be a useful estimate, though.

semaphoreP:
I'm not sure why we would leave `add_photon_noise()` out. My thinking is that the `err` array should be an accurate representation of the noise. If we need to account for other noise terms (e.g., by taking the stddev from a stack), we should do that in a separate step function and add it to the existing `err` array. By the time we're at L3 data products, I want all sources of uncertainty in the `err` array already, and not have to further compute additional error terms from the data.

semaphoreP (approved):
Looks good to me!

semaphoreP:
@kjl0025 I looked through that discussion, and I'm not sure if there's anything in there that's missing in this PR.

kjl0025:
Oh, my bad, I see that this code indeed already covers the other points I mentioned (about masking for calculating err, etc).  Sorry, I had overlooked some lines and didn't actually check the unit tests!  

However, one issue:  test_combine.py only calls one of the tests to run after `if __name__ == '__main__'`.

semaphoreP:
the `pytest` framework ignores code in `if __name__ == '__main__'`. That chunk of code is only used for developer debugging (i.e., running it in a debugger). All pytest evocations will run every defined function that starts with the "test_" in its name.

kjl0025:
Ah, okay, thanks!

Issue #244, opened 2024-10-30
nonlinearity, trap-pump, and astrometric calibration products have unnecessary DQ header


kjl0025:
**Describe the bug**
A DQ header is created for the nonlinearity and trap-pump calibration products.  However, the array values in the nonlinearity array are relative gain values for specific EM gain-DN combinations (not pixel data), and the array values for trap-pump are likewise not pixel data.  So I don't think we need a DQ array here.  Its presence doesn't hurt anything, but a user new to corgidrp may be confused to find this HDU data in the calibration product that doesn't have any meaning.

**To Reproduce**
For example, run an e2e test or unit test that creates a the calibration product.

**Expected behavior**
An array of zeros should be produced for the 3rd extension header (the default DQ header).

**corgidrp version:**
Main branch, v1.0

**Additional context**
Should be a simple fix to remove the DQ header from the NonlinCalibration and TrapCalibration classes.

kjl0025:
Same applies for the astrometric calibration product.

semaphoreP:
Are you saying the current DQ frame isn't an array of all 0's, or are you requesting the DQ extension to be removed completely.

Removing it may make it harder to subclass these data classes from the Image subclass, so I'm inclined to keep the DQ extension for simplicity.

kjl0025:
I was suggesting removing the DQ extension completely to avoid confusion.  But I agree with you that it would be simpler to keep it, which would ensure good compatibility with `Image`.  I don't have strong feelings on this one, just making folks aware of it.  Maybe we could put a comment in the headers for these calibration classes that the DQ is not meaningful?

semaphoreP:
Sure, sounds good to me.

kjl0025:
Cool, I'll add a comment to the relevant headers for my next PR.

kjl0025:
The comment about the irrelevance of the DQ header in these cases has been implemented, so we can close this issue.

Issue #200, opened 2024-09-17
nonlin_params and kgain_params should be made an optional input to calibrate_nonlin() and calibrate_kgain(), respectively


kjl0025:
**Is your feature request related to a problem? Please describe.**
`calibrate_nonlin` is hardcoded to use the dictionary `nonlin_params` at the top of the Python file.  The dictionary uses `nonlin_params` to specify the region of the frame where the pupil image is located, but the pupil image may be located in a different frame region in any given future dataset.

**Describe the solution you'd like**
Make an optional input that defaults to this dictionary so that a user can specify a keyword argument in a custom .json file.

**Additional context**

JuergenSchreiber:
I guess the same is valid for calibrate_kgain? Then we might leave the checks in and delete issue #195?

kjl0025:
Yes, good catch, that too!  We can delete issue #195 then.

kjl0025:
As for `kgain_params`, I think what is there is fine b/c it defines a region for getting data at different DN values, and the region is defined to get the illuminated pupil and surrounding pixels.  And that region is different from the prescan region obtained from DetectorParams.

maxwellmb:
@kjl0025, did PR #277 close this issue?

kjl0025:
Yes, it did!  Sorry, I forgot to link this one to it.

maxwellmb:
> Yes, it did! Sorry, I forgot to link this one to it.

Great, thanks!

PR #277, opened 2024-12-26
Photon counting

kjl0025:
## Describe your changes
Added a step function for photon counting frames.  Also addressed some other minor issues. (This is the same as the PR I had into the main branch.)

## Type of change

#2 , #200 , #244 

- [X] I have linted my code
- [X] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [X] I have verified that all docstrings are properly formatted and added new documentation, as needed


semaphoreP:
@kjl0025 can you finish filling out the DRP Test Definition row with your test description. It'll actually be helpful for my review to be able to read through the tests at a high level. (Rather than waiting to fill it out after PR review).

kjl0025:
Sure, just finished doing that.

tests/e2e_tests/photon_count_e2e.py
```diff
@@ -0,0 +1,151 @@
...
+
+    # more frames gets a better agreement; agreement to 1% for ~160 darks and illuminated
+    assert np.isclose(np.nanmean(pc_frame), ill_mean - dark_mean, rtol=0.01) 
+    assert np.isclose(np.nanmean(pc_dark_frame), dark_mean, rtol=0.01) 
+    assert pc_frame_err.min() >= 0
+    assert pc_dark_frame_err.min() >= 0
+
+    this_caldb.remove_entry(kgain)
+    this_caldb.remove_entry(noise_map)
+    this_caldb.remove_entry(new_nonlinearity)
```
semaphoreP, line 134: we should also remove the PC Dark from the caldb as well.

tests/test_photon_counting.py
```diff
@@ -0,0 +1,119 @@
...
+    # exposure time too long to get reasonable photon-counted result (after photometric correction)
+    np.random.seed(555)
+    dataset_err, dark_dataset_err, ill_mean, dark_mean = mocks.create_photon_countable_frames(Nbrights=160, Ndarks=160, cosmic_rate=0, full_frame=False) 
+    # instead of running through walker, just do the pre-processing steps simply
+    # using EM gain=5000 and kgain=7 and bias=20000 and read noise = 100 and QE=0.9 (quantum efficiency), from mocks.create_photon_countable_frames()
+    for f in dataset_err.frames:
+        f.data = f.data.astype(float)*7 - 20000.
+    for f in dark_dataset_err.frames:
+        f.data = f.data.astype(float)*7 - 20000.
+    dataset_err.all_data = dataset_err.all_data.astype(float)*7 - 20000.
```
semaphoreP, line 37: I think if you do `dataset_err.all_data[:] = dataset_err.all_data.astype(float)*7 - 20000.` with the `[:]` (and equivalent for line 38), then you don't need lines 33-36.

tests/test_photon_counting.py
```diff
@@ -0,0 +1,119 @@
...
+def test_negative():
+    """Values at or below the 
+    threshold are photon-counted as 0, including negative values."""
+    fr = np.array([-3,2,0])
+    pc_fr = photon_count(fr, thresh=2)
+    assert (pc_fr == np.array([0,0,0])).all()
+    
+
+def test_pc():
+    '''Test that a pixel that is masked heavily (reducing the number of usable frames for that pixel) has a bigger err than the average pixel.
```
semaphoreP, line 23: The docstring is a bit hard to read. Can we put each thing that is being tested in a new line?

We should also mention that we test the dark subtraction from illuminated frames gives the expected value.

corgidrp/recipe_templates/l1_to_l2b_pc_dark.json
```diff
@@ -0,0 +1,52 @@
...
+            "name" : "frame_select"
+        },
+        {
+            "name" : "convert_to_electrons",
+            "calibs" : {
+                "KGain" : "AUTOMATIC"
+            }
+        },
+        {
+            "name": "get_pc_mean"
```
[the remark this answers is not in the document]
kjl0025: Okay, sounds good.  I'll make a flag input, and I'll make it raise an exception if the flag value is incompatible with the VISTYPE.

tests/e2e_tests/photon_count_e2e.py
```diff
@@ -0,0 +1,151 @@
...
+    tvac_nonlin_data = np.genfromtxt(input_non_linearity_path, delimiter=",")
+
+    pri_hdr, ext_hdr = mocks.create_default_headers()
+    new_nonlinearity = data.NonLinearityCalibration(tvac_nonlin_data,pri_hdr=pri_hdr,ext_hdr=ext_hdr,input_dataset = dummy_dataset)
+    new_nonlinearity.filename = nonlin_fits_filepath
+    new_nonlinearity.pri_hdr = pri_hdr
+    new_nonlinearity.ext_hdr = ext_hdr
+    this_caldb.create_entry(new_nonlinearity)
+    # make PC dark
+    walker.walk_corgidrp(l1_data_dark_filelist, '', output_dir, template="l1_to_l2b_pc_dark.json")
```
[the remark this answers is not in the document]
kjl0025: I think it would be good for the user to have the freedom to override and run either the PC or analog pipeline, but the walker could be informed by the 'ISPC' header listed here (https://collaboration.ipac.caltech.edu/display/romancoronagraph/L1+image+headers+SSC+release+3.2)?

tests/e2e_tests/photon_count_e2e.py
```diff
@@ -0,0 +1,151 @@
...
+    for f in os.listdir(output_dir):
+        if f.endswith('_pc_dark.fits'):
+            pc_dark_filename = f
+    # add the Dark to the Caldb for processing the illuminated
+    pc_dark_file = os.path.join(output_dir, pc_dark_filename)
+    dark_entry = data.Dark(pc_dark_file)
```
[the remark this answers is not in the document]
kjl0025: It does add it in.  You're right, I don't actually need to add it in.  I just copied lines from a previous formatting.

tests/test_photon_counting.py
kjl0025: I had tried that kind of approach at first, but it didn't work.  The `data` for each frame and `all_data` aren't linked in such a way to update if one of them changes.

kjl0025:
Alright, I think I've addressed all your comments.  Should be good to go.

corgidrp/data.py
```diff
@@ -808,6 +819,13 @@ def __init__(self, data_or_filepath, err = None, ptc = None, pri_hdr=None, ext_h
        # run the image class contructor
         super().__init__(data_or_filepath, err=err, pri_hdr=pri_hdr, ext_hdr=ext_hdr, err_hdr=err_hdr)
 
+        # initialize these headers that have been recently added so that older calib files still contain this keyword when initialized and allow for tests that don't require 
+        # these values to run smoothly; if these values are actually required for 
+        # a particular process, the user would be alerted since these values below would result in an error as they aren't numerical
+        if 'RN' not in self.ext_hdr:
+            self.ext_hdr['RN'] = ''
```
semaphoreP, line 826: should we be initializing them to 0? feels awkward to initalize to a different data type

semaphoreP (approved):
Hi Kevin, two small comments remaining (weird that only one got attached to this comment.. there's another one above). I also need to run the e2e tests myself to make sure they run now, but I will assume you fixed those issues.

tests/e2e_tests/photon_count_e2e.py
```diff
@@ -0,0 +1,179 @@
...
+    assert pc_dark_frame_err.min() >= 0
+
+    # load in CalDB again to reflect the PC Dark that was implicitly added in (but not found in this_caldb, which was loaded before the Dark was created)
+    post_caldb = caldb.CalDB()
+    post_caldb.remove_entry(kgain)
+    post_caldb.remove_entry(noise_map)
+    post_caldb.remove_entry(new_nonlinearity)
+    post_caldb.remove_entry(flat)
+    post_caldb.remove_entry(bp_map)
+    for i in range(len(post_caldb._db['Type'])):
```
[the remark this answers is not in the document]
kjl0025: Ah, true.  It does update it for me, but I do agree that your suggestion is better since `remove_entry` does more than just `drop` (and it does it by finding the corresponding filename and not just by type).

tests/e2e_tests/photon_count_e2e.py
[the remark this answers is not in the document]
kjl0025: I think the way to do the trigger would be to have the walker choose the path based on the 'ISPC' header listed here (https://collaboration.ipac.caltech.edu/display/romancoronagraph/L1+image+headers+SSC+release+3.2).  I can edit the walker accordingly on this PR if you'd like?

corgidrp/data.py
kjl0025: I chose to do a different data type so that we would know for sure if the cal files were legitimate when an error is thrown further down the line.  In other words, I wouldn't want a fake value to be mistakenly used as the real calibrated value when it is called for in a step function.

tests/e2e_tests/photon_count_e2e.py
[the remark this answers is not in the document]
semaphoreP: Yes, that would be great if you can add it. that way, it doesn't force us to manually figure out the PC recipes each time (PC Dark, and PC L1 -> L2b)

corgidrp/data.py
semaphoreP: ok. as long as it doesn't break things that use this attribute.

tests/e2e_tests/photon_count_e2e.py
```diff
@@ -0,0 +1,178 @@
...
+
+import corgidrp
+import corgidrp.data as data
+import corgidrp.mocks as mocks
+import corgidrp.walker as walker
+import corgidrp.caldb as caldb
+import corgidrp.detector as detector
+
+@pytest.mark.e2e
+def test_expected_results_e2e(tvacdata_dir, file_dir):
```
semaphoreP, line 18: The arguments need to be named `tvacdata_path, e2eoutput_path` exactly for the pytest framework to run correctly. Can you rename these two argument variables?

semaphoreP (changes requested):
Just ran the e2e tests and the input argument naming needs to be fixed in order for the batch pytest framework to run the e2e tests. 

Can you also fix the conflict with mock?

kjl0025:
Should be good to go now.

kjl0025:
Oops, not yet.  Fixing

corgidrp/calibrate_kgain.py
```diff
@@ -281,7 +277,7 @@ def calibrate_kgain(dataset_kgain,
                     n_cal=10, n_mean=30, min_val=800, max_val=3000, binwidth=68,
                     make_plot=True,plot_outdir='figures', show_plot=False,
                     logspace_start=-1, logspace_stop=4, logspace_num=200,
-                    verbose=False, detector_regions=None):
+                    verbose=False, detector_regions=None, kgain_params=kgain_params_default):
```
semaphoreP, line 280: the python "best practices" on this is to set `kgain_params=None` and later in the code, `if kgain_params is None: kgain_params = default`. This is to avoid bugs that can occur when setting python objects (which are mutable) as default arguments.

corgidrp/calibrate_nonlin.py
```diff
@@ -124,7 +127,7 @@ def calibrate_nonlin(dataset_nl,
                      pfit_upp_cutoff1 = -2, pfit_upp_cutoff2 = -3,
                      pfit_low_cutoff1 = 2, pfit_low_cutoff2 = 1,
                      make_plot=True, plot_outdir='figures', show_plot=False,
-                     verbose=False):
+                     verbose=False, nonlin_params=nonlin_params_default):
```
semaphoreP, line 130: same comment here as kgain params

semaphoreP (commented):
Hi Kevin, just a few small changes in the code, to conform to some python best practices and hopefully avoid bugs in future development. 

Additionally, it seems like header keywords for CMDGAIN and KGAIN have changed. Can you make a new issue to track that we should update these header keywords? I would strongly advocate not doing it in this PR so we can wrap up the last few things and merge!

corgidrp/walker.py
```diff
@@ -235,14 +237,45 @@ def guess_template(dataset):
...
+                    recipe_filename = "l1_to_l2a_noisemap.json"
+                elif len(unique_vals) == 1: # traditional darks
+                    recipe_filename = "build_trad_dark_image.json"
+
         elif image.pri_hdr['VISTYPE'] == "PUPILIMG":
             recipe_filename = ["l1_to_l2a_nonlin.json", "l1_to_kgain.json"]
         else:
             recipe_filename = "l1_to_l2b.json"
+        if 'ISPC' in image.ext_hdr:
+            if image.ext_hdr['ISPC'] and image.pri_hdr['VISTYPE'] != "DARK":
```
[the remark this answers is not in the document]
kjl0025: I did it that way b/c I wasn't sure what the 'VISTYPE' would be for PC images, but I suppose it wouldn't be any of the ones checked above this line, so I can put in this inside the `else` statement above it and take out the 'DARK' check.  

And I saw Julia's message about how she would make an issue about CMDGAIN and KGAIN.

corgidrp/walker.py
```diff
@@ -235,16 +237,49 @@ def guess_template(dataset):
...
+                if len(unique_vals) > 1: # darks for noisemap creation
+                    recipe_filename = "l1_to_l2a_noisemap.json"
+                elif len(unique_vals) == 1: # traditional darks
+                    recipe_filename = "build_trad_dark_image.json"
         elif image.pri_hdr['VISTYPE'] == "PUPILIMG":
             recipe_filename = ["l1_to_l2a_nonlin.json", "l1_to_kgain.json"]
         else:
-            recipe_filename = "l1_to_l2b.json"
+            if 'ISPC' in image.ext_hdr:
+                if image.ext_hdr['ISPC']:
```
semaphoreP, line 257: can this be done as `if 'ISPC' in image.ext_hdr & image.ext_hdr['ISPC']`? The second have of the conditional will only be run if the first half is True. 

Right now, if "ISPC" is in the header and it is False, the walker does nothing, which is not what we want (I assume we just want it to run "l1_to_2b.json". Same later on below.

corgidrp/walker.py
```diff
@@ -235,16 +237,49 @@ def guess_template(dataset):
...
             recipe_filename = "l1_flat_and_bp.json"
         elif image.pri_hdr['VISTYPE'] == "DARK":
-            recipe_filename = "l1_to_l2a_noisemap.json"
+            _, unique_vals = dataset.split_dataset(exthdr_keywords=['EXPTIME', 'CMDGAIN', 'KGAIN'])
+            if 'ISPC' in image.ext_hdr: # to be added eventually, so check if it's present first
+                if image.ext_hdr['ISPC']:
+                    recipe_filename = "l1_to_l2b_pc_dark.json"
+                elif len(unique_vals) > 1: # darks for noisemap creation
+                    recipe_filename = "l1_to_l2a_noisemap.json"
+                elif len(unique_vals) == 1: # traditional darks
```
semaphoreP, line 246: We shouldn't end on an `if`. We should turn this into an `else` or have an `else` below that raises an Exception. 

Same for the 3 other places you have this structure.

semaphoreP (commented):
Sorry to keep dragging this out, but I want to make sure our logic on the walker has no gaps, to avoid bugs in the future.

semaphoreP (approved):
Looks good!

Issue #261, opened 2024-11-26
Need a step function to crop data


ell-bogat:
**Is your feature request related to a problem? Please describe.**
We need to crop the science data before PSF subtraction so we're not wasting computation/memory. This would be best done as a separate step function.

**Describe the solution you'd like**
This step should crop the data (and error/dq arrays) to the new desired shape and update the pixel locations of the array center, mask location, star location (if computed), etc in the header. 

**Additional context**
- We can assume a square image shape if only one integer is passed. Consider if we should assert that the image shape evenness/oddness is preserved to avoid half-pixel shifts for centering.
- We can also assume that we are cropping to the outer working angle of the instrument mode (plus a 10% or so border) if no specific shape is passed. We should have a dictionary of the OWA for each mode stored somewhere in the DRP code. The imaging mode can be determined based on the coronagraph focal plane mask name (at least for the modes we're supporting so far).

maxwellmb:
Closed with #278

PR #278, opened 2025-01-02
Crop Step Function

ell-bogat:
## Describe your changes
Adds a new step function `l3_to_l4.crop()` to crop data and update relevant header values.
- Can accept any data level.
- Can handle 2D or 3D Image data, but if data is 3D we assume it is a stack of images and only crop the last two indices.
- kwarg `sizexy` is the desired data shape in xy order. Currently required to be even. If a single value is provided we assume a square data shape. Defaults to 60 (a little bigger than the dark hole region).
- kwarg `centerxy` is the pixel location to center the cropping on, should be the intersection of 4 pixels since we want an even data shape. Raises a warning if the provided center is not a half-integer, and centers on the nearest pixel intersection. Defaults to the star location in the extension header (`"STARLOCX"`/`"STARLOCY"`) and will raise an exception if no value is provided but the location is missing from the header.
- Updates axis sizes and various pixel locations that may be in the headers, but we'll need to make sure this function gets updated if new relevant header keywords get added.

**Possible future additions (low priority):**
- Add option to crop to an odd data shape centered on a pixel
- Add capability to handle regions outside the FOV by padding with nans.

**Example:**
Input dataset with a square "PSF" at the position of "STARLOCX/Y" in the header:
image
`cropped_dataset = crop(input_dataset)` automatically crops to 60 pixels around "STARLOCX/Y": 
image

## Type of change

Issue #261: Need a step function to crop data 

- [x] I have linted my code

### DIFF
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -151,10 +151,6 @@ def crop(input_dataset,sizexy=60,centerxy=None):
             prihdr["CRPIX1"] -= x1
             prihdr["CRPIX2"] -= y1
             updated_hdrs.append('CRPIX1/2')
-        if ("PSFCENTX" in prihdr.keys()):
-            prihdr["PSFCENTX"] -= x1
-            prihdr["PSFCENTY"] -= y1
-            updated_hdrs.append('PSFCENTX/Y')
         new_frame = data.Image(cropped_frame_data,prihdr,exthdr,cropped_frame_err,cropped_frame_dq,frame.err_hdr,frame.dq_hdr)
         frames_out.append(new_frame)
 
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -4,6 +4,7 @@ from pyklip.klip import rotate
 from corgidrp import data
 from scipy.ndimage import rotate as rotate_scipy # to avoid duplicated name
 from scipy.ndimage import shift
+import warnings
 import numpy as np
 import glob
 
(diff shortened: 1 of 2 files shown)


corgidrp/l3_to_l4.py
```diff
@@ -36,6 +37,125 @@ def find_star(input_dataset):
...
+    TODO: 
+        - Pad with nans if you try to crop outside the array (handle err & DQ too)
+        - Option to crop to an odd data array and center on a pixel?
+
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (any level)
+        sizexy (int or array of int): desired frame size, if only one number is provided the 
+            desired shape is assumed to be square, otherwise xy order. Defaults to 60.
+        centerxy (float or array of float): desired center (xy order), should be a pixel intersection (a.k.a 
+            half-integer). Defaults to the "STARLOCX/Y" header values.
```
maxwellmb, line 57: Can you add this keyword to [this](https://collaboration.ipac.caltech.edu/display/romancoronagraph/L2-L4+image+header+wishlist) page. It'll help us keep track of keywords being used in the step functions, even if they're placeholders for the moment.

corgidrp/l3_to_l4.py
```diff
@@ -36,6 +37,125 @@ def find_star(input_dataset):
...
+    to the star location. Assumes 3D Image data is a stack of 2D data arrays, so only 
+    crops the last two indices.
+
+    TODO: 
+        - Pad with nans if you try to crop outside the array (handle err & DQ too)
+        - Option to crop to an odd data array and center on a pixel?
+
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (any level)
+        sizexy (int or array of int): desired frame size, if only one number is provided the
```
maxwellmb, line 54: In the issue (which you wrote ;P ) you suggest it would be good to crop based on the instrument mode. I like this idea, but don't want to hold up the current function. Can you make a new issue as a desired upgrade to do that?

corgidrp/l3_to_l4.py
maxwellmb, line 57: It looks like the code rounds to the nearest half-pixel for the center, even if that's not what's input. Do you mind putting that behaviour in the docstring?

maxwellmb: (I like the warning!)

corgidrp/l3_to_l4.py
```diff
@@ -36,6 +37,125 @@ def find_star(input_dataset):
...
+            cropped_frame_dq = frame.dq[y1:y2,x1:x2]
+        elif frame.data.ndim == 3:
+            cropped_frame_data = frame.data[:,y1:y2,x1:x2]
+            cropped_frame_err = frame.err[:,:,y1:y2,x1:x2]
+            cropped_frame_dq = frame.dq[:,y1:y2,x1:x2]
+        else:
+            raise ValueError('Crop function only supports 2D or 3D frame data.')
+
+        # Update headers
+        exthdr["NAXIS1"] = sizexy[0]
```
maxwellmb, line 123: Does this properly represent the case if the data is 3D? I think probably not, right? (this is good to be thinking about since I think polarimetry data will eventually be 3D).

corgidrp/l3_to_l4.py
```diff
@@ -36,6 +37,125 @@ def find_star(input_dataset):
...
+            updated_hdrs.append('STARLOCX/Y')
+        if ("MASKLOCX" in exthdr.keys()):
+            exthdr["MASKLOCX"] -= x1
+            exthdr["MASKLOCY"] -= y1
+            updated_hdrs.append('MASKLOCX/Y')
+        if ("CRPIX1" in prihdr.keys()):
+            prihdr["CRPIX1"] -= x1
+            prihdr["CRPIX2"] -= y1
+            updated_hdrs.append('CRPIX1/2')
+        if ("PSFCENTX" in prihdr.keys()):
```
maxwellmb, line 143: Do PSFCENT and STARLOC have different meanings?

maxwellmb (changes requested):
Some quick change requests/comments

corgidrp/l3_to_l4.py
ell-bogat: They are the same, I think we can remove PSFCENT from the headers. It's just there because it's the default name for that parameter that comes out of pyklip so we don't need both.

corgidrp/l3_to_l4.py
ell-bogat: Mostly! The fits standard definition of the NAXIS keywords is that NAXIS1 corresponds to the fastest changing axis (the x axis), NAXIS 2 is the y axis, etc. I've just added the correct handling of the last axis in the 3D case and fixed the error NAXIS 3 header as well.

corgidrp/l3_to_l4.py
ell-bogat: Done!

corgidrp/l3_to_l4.py
ell-bogat: Done!

corgidrp/l3_to_l4.py
ell-bogat: Done!

corgidrp/l3_to_l4.py
```diff
@@ -36,6 +37,132 @@ def find_star(input_dataset):
...
+
+    for frame in dataset:
+        prihdr = frame.pri_hdr
+        exthdr = frame.ext_hdr
+        dqhdr = frame.dq_hdr
+        errhdr = frame.err_hdr
+
+        # Require that mode is HLC for now
+        if not prihdr['MODE'] == 'HLC':
+            raise UserWarning('Crop function is currently only configured for mode HLC.')
```
maxwellmb, line 82: I think we should make use the FSAMNAME keyword to define the crop region. I'm not sure what the header keyword value we want is, but the definition is "easily understand PAM configuration.", as seen on [this](https://collaboration.ipac.caltech.edu/display/romancoronagraph/L1+image+headers+SSC+release+3.2) page.

maxwellmb (approved):
Thanks for the changes!

PR #280, opened 2025-01-06
PSF Subtraction Step Function

ell-bogat:
## Describe your changes
* New step function  to perform PSF subtraction via pyKLIP
* New subclass of pyKLIP_data in data.py
* Tests for do_psf_subtraction() and PyKLIP_data in tests/test_psfsub.py

**Results using OS11 simulations:**
image
image
image

**Testing using gaussian stellar & planetary PSFs:**
Note: I'm only comparing to the "analytical" result outside of the IWA, as there is increased deviation as you approach the mask center. Also, there is offset between the corgidrp and analytical result equal to the median of the corgidrp output, so if you subtract the median (on the order of -1e-4) then they match much more precisely (order 1e-6 or less).
image
image
image

Output header keywords:
* We keep by default all the header keywords from the first science frame in the dataset
* We then add all the output pyklip header keywords to the ext_hdr.
 
Still to do:
- [x] Handle DQ arrays properly (may need to update what value we assign to the nans pixels after pyklip runs)
- [x] Double check that roll angle direction convention is correct!

Future PRs:
* Calibrate for algorithm throughput in next PR (use OS11 mock calibration for now, eventually use a calibration input file that contains off-axis psfs)
  * Account for core throughput transmission for point sources near the detection limit
* Incorporate WCS header when ready
* How to track mask position after PSF subtraction? If star-mask alignment is possibly different in each input image?
* Handle error arrays properly

## Type of change

Issue #8 : Produce Calibrated PSF Subtracted Images with pyKLIP

- [x] I have linted my code


ell-bogat:
I'm trying to confirm the convention for which direction is a positive roll angle. I was assuming the names of the OS11 files referred to the roll of the telescope, but when I do a straight subtraction of the negative direction roll from the positive direction roll, I get this:
image

If the positive roll direction is from North towards East (left), then the planet should appear to move in the opposite direction, aka clockwise, right? So is it possible the filenames are really referring to the planet position angle, or possibly mislabeled? Or is the convention different? If I don't swap the sign of the rolls when I input them into pyKLIP, I get a negative planet, and I want to make sure the convention is correct.

semaphoreP:
Hi Ell, I frequently have a negative sign uncertainty on the roll angles. I think it's ok if, for OS11, we need to apply a negative sign to get the roll angles right. Hopefully we can get this right after we figure out the ROLL keywords in real CGI headers, but for now, I think having to apply a negative sign for OS11 is fine.

ell-bogat:
Thanks @semaphoreP ! 
The main issue I'm running into now is in the WCS info. My current understanding is there are two ways to store the coordinate transformation, either using a CD-matrix or a PC-matrix + a CDELT matrix (still figuring out exactly what those are). And the fits headers default to storing the PC version, while pyKLIP expects the CD version. For some reason it was working anyway before I updated to the same software versions that the github testing environment uses, but now I'm getting a "no cd present" angry message from pyKLIP causing the tests to fail. Also if I just overwrite the WCS header to None so that pyKLIP ignores it, we don't get the correct psf subtraction result anymore. Is there a simple way to convert from PC to CD? Do we need to tackle generating the correct WCS header info first?

ell-bogat:
Basically any time you use the method to convert a WCS object to a fits header, it is saved using the PC formalism.
e.g. even if I generate an object using the pyKLIP `generate_wcs` function which uses the CD matrix:
```
center = (30,30)
roll = 13.
wcs_generated = generate_wcs(roll,center) # WCS object
display(wcs_generated)

> WCS Keywords
> 
> Number of WCS axes: 2
> CTYPE : 'RA---TAN' 'DEC--TAN' 
> CRVAL : 0.0 0.0 
> CRPIX : 30.0 30.0 
> CD1_1 CD1_2  : -2.7065835132923e-06 6.24864039844069e-07 
> CD2_1 CD2_2  : 6.24864039844069e-07 2.70658351329232e-06 
> NAXIS : 0  0
```
If I then call the wcs.to_header() function to store it in the extension header it produces this:
```
wcs_header = wcs_generated.to_header()
display(wcs_header)

> WCSAXES =                    2 [/](https://file+.vscode-resource.vscode-cdn.net/) Number of coordinate axes                      
> CRPIX1  =                 30.0 [/](https://file+.vscode-resource.vscode-cdn.net/) Pixel coordinate of reference point            
> CRPIX2  =                 30.0 [/](https://file+.vscode-resource.vscode-cdn.net/) Pixel coordinate of reference point            
> PC1_1   = -2.7065835132923E-06 [/](https://file+.vscode-resource.vscode-cdn.net/) Coordinate transformation matrix element       
> PC1_2   =  6.2486403984407E-07 [/](https://file+.vscode-resource.vscode-cdn.net/) Coordinate transformation matrix element       
> PC2_1   =  6.2486403984407E-07 [/](https://file+.vscode-resource.vscode-cdn.net/) Coordinate transformation matrix element       
> PC2_2   =  2.7065835132923E-06 [/](https://file+.vscode-resource.vscode-cdn.net/) Coordinate transformation matrix element       
> CDELT1  =                  1.0 [/](https://file+.vscode-resource.vscode-cdn.net/) [deg] Coordinate increment at reference point  
> CDELT2  =                  1.0 [/](https://file+.vscode-resource.vscode-cdn.net/) [deg] Coordinate increment at reference point  
> CUNIT1  = 'deg'                [/](https://file+.vscode-resource.vscode-cdn.net/) Units of coordinate increment and value        
> CUNIT2  = 'deg'                [/](https://file+.vscode-resource.vscode-cdn.net/) Units of coordinate increment and value        
> CTYPE1  = 'RA---TAN'           [/](https://file+.vscode-resource.vscode-cdn.net/) Right ascension, gnomonic projection           
> CTYPE2  = 'DEC--TAN'           [/](https://file+.vscode-resource.vscode-cdn.net/) Declination, gnomonic projection               
> CRVAL1  =                  0.0 [/](https://file+.vscode-resource.vscode-cdn.net/) [deg] Coordinate value at reference point      
> CRVAL2  =                  0.0 [/](https://file+.vscode-resource.vscode-cdn.net/) [deg] Coordinate value at reference point      
> LONPOLE =                180.0 [/](https://file+.vscode-resource.vscode-cdn.net/) [deg] Native longitude of celestial pole       
> LATPOLE =                  0.0 [/](https://file+.vscode-resource.vscode-cdn.net/) [deg] Native latitude of celestial pole        
> MJDREF  =                  0.0 [/](https://file+.vscode-resource.vscode-cdn.net/) [d] MJD of fiducial time                       
> RADESYS = 'ICRS'               [/](https://file+.vscode-resource.vscode-cdn.net/) Equatorial coordinate system

```

tests/test_psfsub.py
```diff
@@ -0,0 +1,449 @@
...
+    """Tests that pyklip dataset centers frame, assigns rolls, and initializes PSF library properly for ADI data. 
+    """
+
+    rolls = [0,90]
+    # Init with center shifted by 1 pixel
+    mock_sci,mock_ref = create_psfsub_dataset(2,0,rolls,
+                                              centerxy=(50.5,50.5))
+
+    pyklip_dataset = PyKLIPDataset(mock_sci,psflib_dataset=mock_ref)
+    pass
```
maxwellmb, line 67: why pass?

tests/test_psfsub.py
```diff
@@ -0,0 +1,449 @@
...
+## pyKLIP data class tests
+
+def test_pyklipdata_ADI():
+    """Tests that pyklip dataset centers frame, assigns rolls, and initializes PSF library properly for ADI data. 
+    """
+
+    rolls = [0,90]
+    # Init with center shifted by 1 pixel
+    mock_sci,mock_ref = create_psfsub_dataset(2,0,rolls,
+                                              centerxy=(50.5,50.5))
```
maxwellmb, line 64: would it make sense to have the tests move the two dimensions by different amounts so that we can verify we've got x and y correct? (full credit: idea stolen from Vanessa in another circumstance).

tests/test_psfsub.py
```diff
@@ -0,0 +1,449 @@
...
+        # Overall counts should decrease        
+        if not np.nansum(mock_sci[0].data) > np.nansum(frame.data):
+            raise Exception(f"PSF subtraction resulted in increased counts for frame {i}.")
+
+    # Check expected data shape
+    expected_data_shape = (len(numbasis),60,60)
+    if not result.all_data.shape == expected_data_shape:
+        raise Exception(f"Result data shape was {result.all_data.shape} instead of expected {expected_data_shape} after ADI subtraction.")
+
+if __name__ == '__main__':
```
maxwellmb, line 435: should we add tests about manually setting the "mode", both it working correctly, but also catching the error if we put in "SDI" or something incorrect (since you have the error catching)?

corgidrp/l3_to_l4.py
```diff
@@ -160,27 +166,161 @@ def crop(input_dataset,sizexy=None,centerxy=None):
...
+                              calibrate_flux=False, mode=mode,psf_library=pyklip_dataset._psflib,
+                              fileprefix=fileprefix)
+    
+    # Construct corgiDRP dataset from pyKLIP result
+    result_fpath = os.path.join(outdir,f'{fileprefix}-KLmodes-all.fits')   
+    pyklip_data = fits.getdata(result_fpath)
+    pyklip_hdr = fits.getheader(result_fpath)
+
+    frames = []
+    for i,frame_data in enumerate(pyklip_data):
```
maxwellmb, line 260: Here, unless I'm missing something it looks like we'd be setting things up so that there's a new file per KL modes. My guess is that we probably want to keep all the KL modes in just a single file, in which case we wouldn't need to iterate over this, right? We'd just have a 3D image cube (which _I think_ is acceptable-but worth checking)

corgidrp/l3_to_l4.py
```diff
@@ -160,27 +166,161 @@ def crop(input_dataset,sizexy=None,centerxy=None):
...
-             Updated header kws: {", ".join(updated_hdrs)}."""
-    
-    output_dataset.update_after_processing_step(history_msg)
+    history_msg1 = f"""Frames cropped to new shape {list(output_dataset[0].data.shape)} on center {list(centerxy)}. Updated header kws: {", ".join(updated_hdrs)}."""
+    output_dataset.update_after_processing_step(history_msg1)
     
     return output_dataset
 
-def do_psf_subtraction(input_dataset, reference_star_dataset=None):
+def do_psf_subtraction(input_dataset, reference_star_dataset=None,
```
maxwellmb, line 174: I think one option here is to pass in a full dataset and use the IS_PSF flag to distinguish between science and PSF images. It would make things smoother for the walker. @semaphoreP, any thoughts on the preferred way to handle this for the walker?

maxwellmb (changes requested):
Hi all, 
This is looking overall nice to me. Please see the comments attached here and let me know if anything is unclear or if you want to discuss any of them further. Always happy to iterate. 
Cheers, 
Max

corgidrp/l3_to_l4.py
semaphoreP: Max, I thought we chatted about this and discussed the possibility of having the walker split up the input dataset and reference star dataset using the IS_PSF flag, rather than doing it in the step function.

We could also do it here, so it's done in both places for flexibility.

maxwellmb: My brain doesn't work anymore, so it could very well be that we discussed this. 

There's a similar situation for @m-samland's PR where we may have to split up a dataset one place or another, either with the walker or in the step function (w and w/o sat spots). 

Is there a friendly way to do this in the walker, where we could imagine starting with L1 data and taking it all the way through to PSF-subtracted data? 

This did just make me think of the broader dataset: We might have a reference star with and w/o sat spots, as well as science data w and w/o sat spots. Are we going to support inputting it all in at the same time from L1 data all the way through to PSF-subtracted data or should we plan to have manual intervention, or queued recipes?

semaphoreP: We haven't decided, and there's no code in the walker to do this right now, but like you said, this would be 1-2 lines for these special cases. 

I think we have to decide how we want to do this. Perhaps we can discuss next week with Al and Cynthia. Obviously, having flexibility is best. 

Regardless, we can always edit this interface in the future as a small change.

maxwellmb: Ok, @ell-bogat, to maximize flexibility let's do both: Keep the keyword for now and if its None, then check to see if the input dataset has any IS_PSF flags. If so, then split the dataset internally. It should be a line or two of extra code.

corgidrp/data.py
```diff
@@ -1370,6 +1367,433 @@ def __init__(self, data_or_filepath, err = None, pri_hdr=None, ext_hdr=None, err
...
+                raise UserWarning('Requires 2D/3D data cube')
+            NINTS = data.shape[0]
+            pix_scale = shead['PIXSCALE'] # arcsec
+            PIXSCALE += [pix_scale] 
+
+            # Get centers.
+            if self.center_include_offset == True:
+                # Use the offset values from the header to adjust the center
+                centers = np.array([shead['STARLOCX'] + phead['XOFFSET'] / pix_scale, 
+                    shead['STARLOCY'] + phead['YOFFSET'] / pix_scale] * NINTS)
```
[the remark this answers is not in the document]
ell-bogat: I think we won't actually use these, it's a leftover behavior since pyklip doesn't internally track the mask and star location separately.

corgidrp/data.py
```diff
@@ -1370,6 +1367,433 @@ def __init__(self, data_or_filepath, err = None, pri_hdr=None, ext_hdr=None, err
...
+            TELESCOP = phead['TELESCOP']
+            INSTRUME = phead['INSTRUME']
+            CFAMNAME = shead['CFAMNAME']
+            data = frame.data
+            if data.ndim == 2:
+                data = data[np.newaxis, :]
+            if data.ndim != 3:
+                raise UserWarning('Requires 2D/3D data cube')
+            NINTS = data.shape[0]
+            pix_scale = shead['PIXSCALE'] # arcsec
```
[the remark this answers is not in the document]
ell-bogat: This will be written in the WCS header! I confirmed with Amanda/Jason :)

corgidrp/mocks.py
```diff
@@ -2113,3 +2140,226 @@ def create_flux_image(star_flux, fwhm, cal_factor, filedir=None, color_cor = 1.,
...
+CRPIX2  =                  0.0 / Pixel coordinate of reference point            
+CDELT1  =                  1.0 / Coordinate increment at reference point        
+CDELT2  =                  1.0 / Coordinate increment at reference point        
+CRVAL1  =                  0.0 / Coordinate value at reference point            
+CRVAL2  =                  0.0 / Coordinate value at reference point            
+LATPOLE =                 90.0 / [deg] Native latitude of celestial pole        
+MJDREF  =                  0.0 / [d] MJD of fiducial time
+"""
+
+def gaussian_array(array_shape=[50,50],sigma=2.5,amp=100.,xoffset=0.,yoffset=0.):
```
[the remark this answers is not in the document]
ell-bogat: Done!

ell-bogat:
Ready for re-review! Right now here are the tasks that I think should be follow-up issues:
* Calibrate for algorithm throughput in next PR (use OS11 mock calibration for now, eventually use a calibration input file that contains off-axis psfs)
  * Account for core throughput transmission for point sources near the detection limit
* Incorporate WCS header when ready
* Track mask position in each frame after PSF subtraction?
* Handle error arrays properly

maxwellmb (approved):

Issue #160, opened 2024-08-13
Fine astrometric solution + distortion solution


semaphoreP:
**Is your feature request related to a problem? Please describe.**
A step function to:
 * measure average platescale and north angle
 * measure distortion solution (ask Jason if you need code)

This should offer a finer-precision astrometric calibration than the boresight calibration, and also include a distortion map. 

**Describe the solution you'd like**
A new step function, or adopt the existing boresight calibration step function to include a distortion map (former is preferred). 

**Additional context**

semaphoreP:
Implemented in PR #284

PR #284, opened 2025-01-13
Distortion Solution

manduhmia:
## Describe your changes
Adding distortion solution function

## Type of change

#160 

- [ ] I have linted my code


corgidrp/data.py
```diff
@@ -1311,27 +1311,29 @@ class AstrometricCalibration(Image):
     Class for astrometric calibration file. 
     
     Args:
-        data_or_filepath (str or np.array); either the filepath to the FITS file to read in OR the 1D array of calibration measurements
+        data_or_filepath (str or np.array): either the filepath to the FITS file to read in OR an array of calibration measurements (boresight, plate scale, north angle, distortion coeffs)
```
semaphoreP, line 1314: can you be a bit more specific on the format of the array if you pass it in? It's ok to use multiple lines to specify this. e.g., which of those values are single floats (platescale, northangle), and which are multiple entries (and what should be the length of them).

corgidrp/data.py
```diff
@@ -1311,27 +1311,29 @@ class AstrometricCalibration(Image):
...
         # File format checks
-        if self.data.shape != (4,):
-            raise ValueError("The AstrometricCalibration data should be a 1D array of four values")
+        if type(self.data) != np.ndarray:
+            raise ValueError("The AstrometricCalibration data should be an array of calibration measurements")
         else:
             self.boresight = self.data[:2]
             self.platescale = self.data[2]
             self.northangle = self.data[3]
+            self.distortion_coeffs = (self.data[4:-1], self.data[-1])
```
semaphoreP, line 1336: what is the purpose of self.data[-1]?

corgidrp/data.py
```diff
@@ -1311,27 +1311,29 @@ class AstrometricCalibration(Image):
...
-    def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=None):
+    def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err=None, input_dataset=None):
         # run the image class constructor
-        super().__init__(data_or_filepath, pri_hdr=pri_hdr, ext_hdr=ext_hdr)
+        super().__init__(data_or_filepath, pri_hdr=pri_hdr, ext_hdr=ext_hdr, err=err)
 
         # File format checks
-        if self.data.shape != (4,):
-            raise ValueError("The AstrometricCalibration data should be a 1D array of four values")
+        if type(self.data) != np.ndarray:
```
semaphoreP, line 1330: you could still be checking for a minimum length (appears to be 6)?

corgidrp/mocks.py
```diff
@@ -984,9 +987,69 @@ def create_astrom_data(field_path, filedir=None, subfield_radius=0.02, platescal
...
+        coeff_data = np.genfromtxt(distortion_coeffs_path, delimiter=',', dtype=None)
+        fitorder = int(coeff_data[-1])
+
+        # convert legendre polynomials into distortin maps in x and y 
+        yorig, xorig = np.indices(image_shape)
+        y0, x0 = image_shape[0]//2, image_shape[1]//2
+        yorig -= y0
+        xorig -= x0
+
+        # get the number of fitting params from the order
```
semaphoreP, line 1005: doesn't have to be this PR, but these 20 lines (taking an array of distortion params, and turning it into a 2-D distortion map) would probably be useful to have as a separate function, since I imagine you might be calling it in multiple places.

semaphoreP: e.g., correction distortion will also use it. probably can replace some code in the distortion cost function as well.

tests/test_distortion_map_creation.py
```diff
@@ -0,0 +1,119 @@
...
+    y_corr = np.polynomial.legendre.legval2d(xorig.ravel(), yorig.ravel(), best_params_y)
+    y_corr = y_corr.reshape(yorig.shape)
+    y_diff = y_corr - yorig
+
+    true_y_corr = np.polynomial.legendre.legval2d(xorig.ravel(), yorig.ravel(), true_params_y)
+    true_y_corr = true_y_corr.reshape(yorig.shape)
+    true_y_diff = true_y_corr - yorig
+
+    # check the distortion maps are less than the maximum injected distortion (~3 pixels)
+    assert np.all(np.abs(x_diff)) < np.max(np.abs(true_x_diff))
```
semaphoreP, line 101: Perhaps we should follow this statement from the Zellem paper: "Current model distortion maps indicate that distortion is less than 4 mas in the central 1 arcsec × 1 arcsec (TTR5) region of the detector". So check that the error in the distortion within the central 1" x 1" region is < 4 mas.

tests/test_distortion_map_creation.py
semaphoreP, line 1: shoudl this file just be combined with test_astrom.py? I don't think it needs to live separately.

semaphoreP (changes requested):
Overall looks good. I have some minor comments to clean things up and add a bit for documentation. 

In the future, we should think about perhaps re-framing the distortion correction so we can use the linear solver for solving legendre polynomial coefficients: https://numpy.org/doc/stable/reference/generated/numpy.polynomial.legendre.legfit.html#numpy.polynomial.legendre.legfit. I think this will make it easier to converge.

semaphoreP:
Please also add a row to the table on the DRP Test Definition page, before we finish the PR (hopefully quick).

corgidrp/data.py
manduhmia: the last entry of the distortion_coeffs is the fit order of the polynomial so this was intended as a way to separate those, but isn't necessary

corgidrp/data.py
manduhmia: I think the length now depends on the order of leg polynomial being fit

corgidrp/mocks.py
manduhmia: okay, I made an issue for this

corgidrp/l2b_to_l3.py
```diff
@@ -1,18 +1,69 @@
...
+        wcs_info['CRPIX1'] = center_pixel[0]
+        wcs_info['CRPIX2'] = center_pixel[1]
+
+        wcs_info['CTYPE1'] = 'RA---TAN'
+        wcs_info['CTYPE2'] = 'DEC--TAN'
+
+        wcs_info['CDELT1'] = (platescale * 0.001) / 3600  ## converting to degrees
+        wcs_info['CDELT2'] = (platescale * 0.001) / 3600
+
+        wcs_info['CRVAL1'] = center_coord[0]
```
semaphoreP, line 52: The center coordinate value should be an offset from the commanded RA/Dec of the telescope pointing. Basically, the RA/Dec the telescope tried to point at, and the offset that we think it missed it by based on the boresight calibration.

semaphoreP (commented):
Another comment based on an issue raised by Max.

semaphoreP (approved):

PR #304, opened 2025-02-21
Measure star center behind coronagraph

m-samland:
## Describe your changes
This pull request implements the IIT version of measuring the star position behind the coronagraph into corgidrp. This is a direct port, with minor changes.

New files:
- star_center.py: contains the IIT functions changed to fit corgidrp. It includes a dictionary that contains parameters that were originally in YAML files.
- mocks.py: contains the functions 1) "create_synthetic_satellite_spot_image" that implements a simple synthetic image with noise and Gaussian satellite spots that can be configured in different positions and orientations (no stellar PSF is included at the moment). 2) "create_satellite_spot_observing_sequence" that creates the 3 data sets needed for measuring the star position: science data without satellite spots, and positive, negative DM modulation spots. Positive and negative images are the same, since no speckle field is generated.
- test_find_star.py: Currently generates a mock "NFOV" observing sequence and tests if the measured center is within 0.1 pixel in x and y direction of the expected position.

Modified:
- l3_to_l4.py: Contains the "find_star" function that wraps everything

I have marked some places in the code with "TODO" tags, that I think may need revisiting and improvements in the future.

## Type of change

#1 

- [x] I have linted my code


corgidrp/star_center.py
```diff
@@ -0,0 +1,871 @@
...
+    Number of points used along each direction for the grid search.
+    Odd numbers are better to provide symmetry of values when the array is
+    truly centered.
+stepSize : float
+    The step size used in the grid search. Units of pixels.
+nIter : int
+    Number of iterations in the loop that hones in on the radial separation
+    of the satellite spots.
+"""
+satellite_spot_parameters = {
```
maxwellmb, line 35: So in the end I think we're going to have to put many if not all of these in find_star. In the "production environment" version of this, it will be very difficult to manually adjust parameters such as these on the fly because anything in a .py file is essentially frozen. The best way we'll have to access them is from keywords in the step function itself. This does seem like a lot of keywords though, so if we can fix any of them, that'd be great.

maxwellmb: Perhaps one way to do it is to have find_star accept some kind of dictionary that overwrites the default values, so that the input for find_star is just a dictionary. It defaults to None, and when its present it must have the structure {"offset": {}, "separation":{}}. thoughts?

corgidrp/star_center.py
```diff
@@ -0,0 +1,871 @@
...
+def star_center_from_satellite_spots(
+    img_ref,
+    img_plus,
+    img_minus,
+    xOffsetGuess,
+    yOffsetGuess,
+    thetaOffsetGuess,
+    satellite_spot_parameters,
+    observing_mode='NFOV',
+):
```
maxwellmb, line 791: if you implement my above suggestion of allowing a user to overwrite `satellite_spot_parameters` then we might want to have the observing mode of `satellite_spot_parameters` selected in the find_star function just before things get overwritten

corgidrp/l3_to_l4.py
```diff
@@ -23,19 +24,83 @@ def distortion_correction(input_dataset, distortion_calibration):
...
+    """
+    # Copy input dataset
+    dataset = input_dataset.copy()
+    observing_mode = []
+
+    # Check that all frames for each input have the correct satellite spot setting
+    # TODO: There may be a lot more we want to check for here
+    # TODO: Add check on negative or positive DM offset setting once we know what the keyword is
+    for frame in dataset.frames:
+        assert frame.pri_hdr["VISTYPE"] == "SCI", "Input has to be science frames."
```
maxwellmb, line 62: It doesn't look like this is a valid VISTYPE: https://collaboration.ipac.caltech.edu/display/romancoronagraph/VISTYPE

I also don't think you really need to check the VISTYPE here necessarily.

corgidrp/mocks.py
```diff
@@ -2113,3 +2115,186 @@ def create_flux_image(star_flux, fwhm, cal_factor, filedir=None, color_cor = 1.,
...
+    minus_frames = []
+    
+    prihdr, exthdr = create_default_headers()
+    prihdr['TELESCOP'] = 'ROMAN'
+    prihdr['INSTRUME'] = 'CGI'
+    prihdr['FILENAME'] = ''
+    prihdr['NAXIS1'] = image_shape[1]
+    prihdr['NAXIS2'] = image_shape[0]
+    exthdr['BUNIT'] = 'MJy/sr'
+    exthdr["HIERARCH DATA_LEVEL"] = 'L3'
```
maxwellmb, line 2248: I think you can start transitioning to the new headers. See the new create_default_L?_headers functions

corgidrp/l3_to_l4.py
```diff
@@ -23,19 +24,83 @@ def distortion_correction(input_dataset, distortion_calibration):
...
+        thetaOffsetGuess=thetaOffsetGuess,
+        satellite_spot_parameters=satellite_spot_parameters,
+        observing_mode=observing_mode,
+    )
+
+    # Add star location to frame headers
+    # TODO: Do we want to save the spot location for diagnostics too?
+    for sci_frame in dataset.frames:
+        sci_frame.ext_hdr['STARLOCX'] = star_xy[0]
+        sci_frame.ext_hdr['STARLOCY'] = star_xy[1]
```
maxwellmb, line 101: If I understand correctly this returns an offset relative to the center of the array. Let's return the actual position of the star center in pixels on the array so that the user doesn't have to guess where the reference point is.

corgidrp/l3_to_l4.py
```diff
@@ -23,19 +24,83 @@ def distortion_correction(input_dataset, distortion_calibration):
...
+        xOffsetGuess=xOffsetGuess,
+        yOffsetGuess=yOffsetGuess,
+        thetaOffsetGuess=thetaOffsetGuess,
+        satellite_spot_parameters=satellite_spot_parameters,
+        observing_mode=observing_mode,
+    )
+
+    # Add star location to frame headers
+    # TODO: Do we want to save the spot location for diagnostics too?
+    for sci_frame in dataset.frames:
```
maxwellmb, line 99: it is preferable to use the `update_after_processing_step` function of `Dataset`, in part so that you're also forced to fill in a `history_entry`.

tests/test_find_star.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # Read initial parameters for testing
+    satellite_spot_parameters = star_center.satellite_spot_parameters
+    separation = satellite_spot_parameters['NFOV']['separation']['spotSepPix']
+
+    # Generate test data
+    sci_dataset, plus_dataset, minus_dataset = mocks.create_satellite_spot_observing_sequence(
+            n_sci_frames=3, n_plus_frames=3, n_minus_frames=3, 
+            image_shape=(201, 201), bg_sigma=1.0, bg_offset=10.0,
+            gaussian_fwhm=5.0, separation=separation, center_offset=(0, 0), angle_offset=0,
```
maxwellmb, line 25: Maybe we also want to test when things aren't exactly centered as well?

maxwellmb: I don't think we need to be extreme, but just something that isn't the exact nominal value

maxwellmb (changes requested):
Hi Matthias, 

Looks like a great start. Just a few comments. Let me know if you have any questions!

I think we'll want to have a broader discussion about how exactly the walker will get triggered for a dataset that contains satellite spots, and how we'll pass in all the data. I think that will more likely come up when we start writing the e2e tests, which we can discuss next week. 

Cheers, 
Max

corgidrp/l3_to_l4.py
```diff
@@ -23,19 +24,83 @@ def distortion_correction(input_dataset, distortion_calibration):
...
+        assert frame.pri_hdr["SATSPOTS"] == 0, "Science frames should not have satellite spots."
+        observing_mode.append(frame.ext_hdr['FSMPRFL'])
+    for frame in plus_satspot_dataset:
+        assert frame.pri_hdr["VISTYPE"] == "SCI", "Input has to be science frames."
+        assert frame.pri_hdr["SATSPOTS"] == 1, "Satellite spot frames should have satellite spots."
+        observing_mode.append(frame.ext_hdr['FSMPRFL'])
+    for frame in minus_satspot_dataset:
+        assert frame.pri_hdr["VISTYPE"] == "SCI", "Input has to be science frames."
+        assert frame.pri_hdr["SATSPOTS"] == 1, "Satellite spot frames should have satellite spots."
+        observing_mode.append(frame.ext_hdr['FSMPRFL'])
```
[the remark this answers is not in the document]
maxwellmb: On second thought (Jason pointed this out) we probably want to output both science and sat spot frames with the PSF centers by default. But maybe optionally (with a kwarg?) drop the sat spot files?

corgidrp/l3_to_l4.py
```diff
@@ -23,19 +24,83 @@ def distortion_correction(input_dataset, distortion_calibration):
...
+
+    # Check if all frames have the same observing mode
+    same_observing_mode = all(mode == observing_mode[0] for mode in observing_mode)
+    assert same_observing_mode, "All frames should have the same observing mode."
+    observing_mode = observing_mode[0]
+
+    # For now we simply take the median of the provided frames to compute the center
+    img_ref = np.median(dataset.all_data, axis=0)
+    img_plus = np.median(plus_satspot_dataset.all_data, axis=0)
+    img_minus = np.median(minus_satspot_dataset.all_data, axis=0)
```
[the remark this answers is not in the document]
m-samland: Agreed and implemented.

corgidrp/l3_to_l4.py
m-samland: I removed the check.

corgidrp/l3_to_l4.py
m-samland: My bad! Indeed this should be changed.

corgidrp/l3_to_l4.py
m-samland: Done.

tests/test_find_star.py
m-samland: I have a question, I'm refactoring the mock function and other functions to take an absolute image coordinate as the input position. Do you agree? 

That would mean that in the find_star routine we also provide not an offset from the image center but an absolute position and default to the image center is None is provided. 

My question is how well will we known the position of the star after acquisition. I assume the coronagraph position is fixed, so we start out with a good idea of where the star will be located on the detector?

corgidrp/star_center.py
m-samland: That sounds good. Does the walker support dictionaries as input?
I'll implement it as a dictionary for now and then we can iterate on it if there are issues.

maxwellmb: My instinct is to have internal defaults that can be overwritten by a dictionary that's passed into an argument that defaults to None. The input dictionary would only overwrite the needed keywords. @semaphorep might have inputs

semaphoreP: The walker supports dictionaries as input!

Agreed with Max that the input dictionary should only have the overwrite the needed keywords, to avoid having to put all the entries dictionary into the recipe each time.

m-samland: OK! This is implemented in the newest version.

corgidrp/l3_to_l4.py
m-samland: Done.

corgidrp/mocks.py
```diff
@@ -637,6 +638,10 @@ def create_noise_maps(FPN_map, FPN_map_err, FPN_map_dq, CIC_map, CIC_map_err, CI
     exthdr['B_O'] = 0.01
     exthdr['B_O_UNIT'] = 'DN'
     exthdr['B_O_ERR'] = 0.001
+
+    err_hdr = fits.Header()
+    err_hdr['BUNIT'] = 'detected electrons'
+    exthdr['DATATYPE'] = 'DetectorNoiseMaps'
```
maxwellmb, line 644: errr, are these because you're not up to date with main?

corgidrp/star_center.py
```diff
@@ -0,0 +1,1037 @@
...
+    img_spots = img_sat_spot - img_ref
+
+    # Grab the relevant tuning parameters
+    tuningParamDict = satellite_spot_parameters
+
+    # estimate star location:
+    # calc_star_location_from_spots should return (xOffsetEst, yOffsetEst)
+    xOffsetEst, yOffsetEst = calc_star_location_from_spots(
+        img_spots, xOffsetGuess, yOffsetGuess, tuningParamDict['offset']
+    )
```
maxwellmb, line 1004: does this break if `satellite_spot_parameters` is None? Should it just be an argument rather than keyword argument?

maxwellmb (changes requested):
Looking good to me! Only a few remaining things: two minor comments and it looks like the flake8 is failing. Once these things are resolved I'll have to double check that the tests are passing, but otherwise we should be good to go.

maxwellmb (approved):

Issue #272, opened 2024-12-18
Possible mismatch in data type in bad pixel map test.


ecady-jpl:
**Describe the bug**
At [this location](https://github.com/roman-corgi/corgidrp/blob/v1.1.2/tests/test_badpixelmap.py#L75) in the badpixelmap test code, `np.unpackbits` is used to separate the bits in the bad pixel map.  However, the DQ element [is a 64-bit integer](https://github.com/roman-corgi/corgidrp/issues/267#issuecomment-2549317604), but `np.unpackbits` [only works on uint8s](https://numpy.org/doc/stable/reference/generated/numpy.unpackbits.html).  Any bits beyond the first 8 will be dropped, and a DeprecationWarning will be raised if any are set.

This is not currently a breaking bug as the corgidrp implementation document only assigns values up to bit 7, and higher ones are TBD.  But when that TBD is defined, this is likely to appear when running the test suite, and may cause test failures.

**To Reproduce**
```
>>> import numpy as np
>>> np.unpackbits(np.array([[0], [1], [255]], dtype=np.uint8), axis=1)
array([[0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 1],
       [1, 1, 1, 1, 1, 1, 1, 1]], dtype=uint8)
>>> np.unpackbits(np.array([[0], [1], [255], [256]], dtype=np.uint8), axis=1)
<stdin>:1: DeprecationWarning: NumPy will stop allowing conversion of out-of-bound Python integers to integer arrays.  The conversion of 256 to uint8 will fail in the future.
For the old behavior, usually:
    np.array(value).astype(dtype)
will give the desired result (the cast overflows).
array([[0, 0, 0, 0, 0, 0, 0, 0],
       [0, 0, 0, 0, 0, 0, 0, 1],
       [1, 1, 1, 1, 1, 1, 1, 1],
       [0, 0, 0, 0, 0, 0, 0, 0]], dtype=uint8)
```

**Expected behavior**
A different tool needs to be used to separate the bits in the DQ bit mask

**corgidrp version:**
 - v1.1.2

semaphoreP:
I think this is now addressed in #317. We use a custom function to unpack bits that should support > 8 bits.

maxwellmb:
@alexisyslau, @semaphoreP does this need to be re-opened?

semaphoreP:
[...]

PR #317, opened 2025-03-05
Bug fix/bad pixel map to address "Possible mismatch in data type in bad pixel map test"

alexisyslau:
## Describe your changes
Situation: numpy.unpackbits only supports uint8 array but ours are uint64, meaning when value exceeds 255, it throws an error. 

I wrote 2 functions for packing and unpacking the bits array, while maintaining the dimension of the DQ array. They can be found data.py module (at the bottom of the file). Unit test was changed (test_badpixelmap.py) to test the packing and unpacking. 

## Type of change

#272 

- [x] I have linted my code

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -1416,3 +1416,30 @@ def autoload(filepath):
     frame = data_class(filepath)
 
     return frame
+
+def unpackbits_64uint(arr, axis):
+    """
+    Unpacking bits from a 64-bit unsigned integer array
+
+    Args:
+        arr (np.ndarray): the array to unpack
+        axis (int): axis to unpack
+
+    Returns:
+        _type_: np.ndarray of bits
+    """
+    n = np.array(arr).view("u1")
+    return np.unpackbits(n, axis=axis, bitorder='little')
+
+def packbits_64uint(arr, axis):
+    """
+    Packing bits into a 64-bit unsigned integer array
+
+    Args:
+        arr (np.ndarray): the array to pack 
+        axis (int): axis to pack
+
+    Returns:
+        _type_: np.ndarray of 64-bit unsigned integers
+    """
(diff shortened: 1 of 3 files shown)


corgidrp/data.py
```diff
@@ -1416,3 +1416,30 @@ def autoload(filepath):
...
+    Packing bits into a 64-bit unsigned integer array
+
+    Args:
+        arr (np.ndarray): the array to pack 
+        axis (int): axis to pack
+
+    Returns:
+        _type_: np.ndarray of 64-bit unsigned integers
+    """
+    return np.packbits(arr, axis=axis, bitorder='little').view(np.uint64)
```
semaphoreP, line 1445: Do you understand why you are using little endien, when the now-deleted comment in line 74 says unpakcing the picks with big endien?

alexisyslau: I chose one arbitrarily between 'big' and 'little'. The difference is in how the bits within each byte are ordered.

        'big': most significant bit → least significant bit within each byte.
        'little': least significant bit → most significant bit within each byte.

I can set both functions to use 'big', and the dq array will still unpack and pack correctly, as long as the same bitorder is used consistently in both directions.

What bitorder does the downstream product expect?"

semaphoreP: I looked through and I think we are never working on unpacked bits (only on the unsigned ints so far), so I think sticking with little endien is fine. Can we just add a comment to the docstring of the function that this is the behavior?

alexisyslau: Ok, I can write some examples and explain what is going to be the impact. As far as I understand, depending on how you would like to proceed with flags, the order potentially will have a significant impact. This is because it might not give your the right flag, if you decide to work on bits instead of decimal.

semaphoreP (approved):
Ok I think we are good!

PR #320, opened 2025-03-06
update northup function incorporating WCS solutions

taichiuyama:
## Describe your changes
update the northup step function by incorporating WCS solutions

## Type of change

- [ ] I have linted my code

### DIFF
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -182,15 +182,15 @@ def do_psf_subtraction(input_dataset, reference_star_dataset=None):
 
     return input_dataset.copy()
 
-def northup(input_dataset,correct_wcs=False):
+def northup(input_dataset,correct_wcs=True):
     """
     Derotate the Image, ERR, and DQ data by the angle offset to make the FoV up to North. 
-    Now tentatively assuming 'ROLL' in the primary header incorporates all the angle offset, and the center of the FoV is the star position.
-    WCS correction is not yet implemented - TBD.
+    Now tentatively assuming the center of the FoV as the star position.
+    WCS correction is incorporated - the angle offset is calculated based on the CD information.
 
     Args:
         input_dataset (corgidrp.data.Dataset): a dataset of Images (L3-level)
-	correct_wcs: if you want to correct WCS solutions after rotation, set True. Now hardcoded with not using astr_hdr.
+	correct_wcs: if you want to correct WCS solutions after rotation, set True. 
 
     Returns:
         corgidrp.data.Dataset: North is up, East is left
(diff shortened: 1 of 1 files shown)


semaphoreP:
Looks like you are using a variable `sci_hd` that is not defined. This seems like you have a wrong variable name?

taichiuyama:
yeah had different definitions, now it's fixed and will retry CI tests

taichiuyama:
now the step function worked, will update the unit test

taichiuyama:
added assert to check if the north position angle after northup function is facing up, and it works. It seems the issue #327 does not directly affect the unit test. The input dataset refers to northangle from only the one from boresight_calibration.

tests/test_northup.py
```diff
@@ -1,73 +1,192 @@
...
+        im_derot = derot_data.data
+        dq_input = input_data.dq
+        dq_derot = derot_data.dq
+
+        # the location for test, where the mock file has 1 in DQ
+        x_value1 = input_dataset[0].ext_hdr['X_1VAL']
+        y_value1 = input_dataset[0].ext_hdr['Y_1VAL']
+
+        # check if rotation works properly
+        assert(im_input[y_value1,x_value1] != im_derot[y_value1,x_value1])
```
semaphoreP, line 126: It would be great to also programmatically check that the bright pixels have been rotated to the correct position (in addition to the bright pixels no longer being in the same place). We're going to be normally running these tests in automated mode, so we don't have the time to make the figure and visually inspect it. You could be able to prediction based on the rotation angle + center where the bright pixels will land up.

You can either do it in this PR, or you can make a new issue for this as an enhancement, and we pull this PR in.

semaphoreP (commented):
Looking good! I have 2 very minor comments, and 1 last suggestion for the unit test to test for correctness.

taichiuyama:
@semaphoreP ok I addressed your comments, please review them

corgidrp/l3_to_l4.py
```diff
@@ -313,66 +314,80 @@ def do_psf_subtraction(input_dataset, reference_star_dataset=None,
...
     return dataset_out
 
-def northup(input_dataset,correct_wcs=False):
+def northup(input_dataset,use_wcs=True):
     """
     Derotate the Image, ERR, and DQ data by the angle offset to make the FoV up to North. 
-    Now tentatively assuming 'ROLL' in the primary header incorporates all the angle offset, and the center of the FoV is the star position.
-    WCS correction is not yet implemented - TBD.
+    Now tentatively assuming the center of the FoV as the star position.
+    WCS correction is incorporated - the angle offset is calculated based on the CD information.
```
semaphoreP, line 321: I think we want to be clear in the docstring (Both here) and in the description of the `use_wcs` variable name that, to determine the angle we need to rotate, we either use the WCS header information (if `use_wcs` is True) or use just the `ROLL` header keyword if `use_wcs` is False (not recommended).

semaphoreP (commented):
Ok, thanks for making those changes! Especially adding that correctness check in the unit tests. I have one very tiny comment left, but I want to make sure the docstring of the northup() function is clear, so people know what it is doing without needing to read the code.

semaphoreP (approved):

PR #321, opened 2025-03-07
Bug fix/going back to uint8 badpixelmap

alexisyslau:
## Describe your changes
1. Change the definition of hot pixel and bad pixel used in badpixelmap back to `uint8`. 
2. Enforce the endianness to 'big' to make sure it aligns with the bit definition. 

## Type of change

- [ ] I have linted my code

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -1836,8 +1836,9 @@ def unpackbits_64uint(arr, axis):
     Returns:
         _type_: np.ndarray of bits
     """
-    n = np.array(arr).view("u1")
-    return np.unpackbits(n, axis=axis, bitorder='little')
+    arr = arr.astype('>u8')
+    n = arr.view('u1')
+    return np.unpackbits(n, axis=axis, bitorder='big')
 
 def packbits_64uint(arr, axis):
     """
@@ -1850,4 +1851,4 @@ def packbits_64uint(arr, axis):
     Returns:
         _type_: np.ndarray of 64-bit unsigned integers
     """
-    return np.packbits(arr, axis=axis, bitorder='little').view(np.uint64)
+    return np.packbits(arr, axis=axis, bitorder='big').view('>u8')
--- a/tests/test_badpixelmap.py
+++ b/tests/test_badpixelmap.py
@@ -72,15 +72,10 @@ def test_badpixelmap():
     ###### make the badpixel map (input the flat_dataset just as a dummy):
     badpixelmap = create_bad_pixel_map(flat_dataset, dark_frame,flat_frame, dthresh=6) # you have integer in here
     
-    current_value = badpixelmap.data[0,0]
-    badpixelmap.data[0,0] = 256 # a value to test uint64
-
     # Use np.unpackbits to unpack the bits - big endien integer to binary
-    badpixelmap_bits = data.unpackbits_64uint(arr=badpixelmap.data[:, :, np.newaxis], axis=2)  # unit64 to binary
-    badpixelmap_repacked = data.packbits_64uint(badpixelmap_bits, axis=2).reshape(badpixelmap.data.shape)
+    badpixelmap_bits = np.unpackbits(badpixelmap.data[:, :, np.newaxis], axis=2, bitorder='big')  # unit64 to binary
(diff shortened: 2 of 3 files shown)


corgidrp/data.py
```diff
@@ -1836,8 +1836,9 @@ def unpackbits_64uint(arr, axis):
     Returns:
         _type_: np.ndarray of bits
     """
-    n = np.array(arr).view("u1")
-    return np.unpackbits(n, axis=axis, bitorder='little')
+    arr = arr.astype('>u8')
+    n = arr.view('u1')
+    return np.unpackbits(n, axis=axis, bitorder='big')
 
 def packbits_64uint(arr, axis):
```
semaphoreP, line 1843: can we write 1 or 2 unit tests to test the functionality of these functions, even if we aren't using them right now?

tests/test_badpixelmap.py
```diff
@@ -8,6 +8,11 @@
 from corgidrp.darks import build_trad_dark
 
 np.random.seed(456)
+
```
alexisyslau, line 11: Getting the flags from function so we no longer need to manually count the index for test

tests/test_badpixelmap.py
```diff
@@ -71,37 +76,61 @@ def test_badpixelmap():
...
     bp_locations = np.argwhere(badpixelmap.data)
+    
     for ii in bp_locations[:,0]:
         assert ii in col_hot_pixels_test or ii in col_dead_pixel_test
     for jj in bp_locations[:,1]:
         assert jj in row_hot_pixels_test or jj in row_dead_pixel_test
 
-    # Checking that hot pixels are at the expected locations - bit #4
-    hot_pixel_locations = np.where(badpixelmap_bits[:,:,-4])
+    # Checking that hot pixels are at the expected locations - bit #3
```
alexisyslau, line 93: It should bit 3 instead, please check

tests/test_badpixelmap.py
```diff
@@ -71,37 +76,61 @@ def test_badpixelmap():
...
+    dead_pixel_locations = np.where(badpixelmap_bits[:,:,dead_pixel_bit_position])
     for ii in dead_pixel_locations[0]:
-        assert ii in col_dead_pixel_test
+        assert ii in col_dead_pixel_test 
     for jj in dead_pixel_locations[1]:
-        assert jj in row_dead_pixel_test 
+        assert jj in row_dead_pixel_test
+
+
+def test_packing_unpacking_uint64():
```
alexisyslau, line 110: Unit tests for packing and unpacking uint64 data: 
TBD should have the decimal value of 256, or bit 8. I opted for an arbitrary array just to test and make sure the functions are not doing something weird: 
- unpacking to the wrong bit
- unpacking correctly such that we only have the '1' at the correct index.

alexisyslau: packing and unpacking were being tested in the test_badpixelmap.py The test in here is for explicitly bit 8 and above

corgidrp/data.py
```diff
@@ -1848,6 +1849,53 @@ def packbits_64uint(arr, axis):
...
 
     Returns:
-        _type_: np.ndarray of 64-bit unsigned integers
+        np.ndarray of 64-bit unsigned integers
     """
-    return np.packbits(arr, axis=axis, bitorder='little').view(np.uint64)
\ No newline at end of file
+    return np.packbits(arr, axis=axis, bitorder='big').view('>u8')
+
+def get_flag_to_bit_map():
```
alexisyslau, line 1856: I added these 2 functions to get the conversion between flag value (decimal) and bit (binary) using dictionary, so we do not need to constantly lookup the table. However, we do need to update the dictionary when there are changes in definition. I can remove them if you guys think this is unnecessary.  
```

Example usage is as follow:

FLAG_TO_BIT_MAP = data.get_flag_to_bit_map()
FLAG_TO_VALUE_MAP = data.get_flag_to_value_map()

bit_position = FLAG_TO_BIT_MAP["TBD"] #gives which index it should be in with the key -> expected index will be equal to [63 - bit_position]
flag_value = FLAG_TO_VALUE_MAP["TBD"] #gives which value it should have with the bit 
```

semaphoreP:
@alexisyslau I think the only thing left is to follow-up on my comment to but the example usage in the docstring the new `get_flag_to_bit_map` and `get_flag_to_value_map` functions.

PR #329, opened 2025-03-10
E2E tests working with updated headers

semaphoreP:
## Describe your changes

Fixed all the e2e tests to run with updated headers. @juliamilton, if this works for you, please merge it into your branch. 

## Type of change

- [ ] I have linted my code

### DIFF
--- a/tests/e2e_tests/astrom_e2e.py
+++ b/tests/e2e_tests/astrom_e2e.py
@@ -14,6 +14,33 @@ import corgidrp.detector as detector
 
 thisfile_dir = os.path.dirname(__file__) # this file's folder
 
+def fix_headers_for_tvac(
+    list_of_fits,
+    ):
+    """ 
+    Fixes TVAC headers to be consistent with flight headers. 
+    Writes headers back to disk
+
+    Args:
+        list_of_fits (list): list of FITS files that need to be updated.
+    """
+    print("Fixing TVAC headers")
+    for file in list_of_fits:
+        fits_file = fits.open(file)
+        prihdr = fits_file[0].header
+        exthdr = fits_file[1].header
+        # Adjust VISTYPE
+        prihdr['OBSNUM'] = prihdr['OBSID']
+        exthdr['EMGAIN_C'] = exthdr['CMDGAIN']
+        exthdr['EMGAIN_A'] = -1
+        exthdr['DATALVL'] = exthdr['DATA_LEVEL']
+    # exthdr['KGAINPAR'] = exthdr['KGAIN']
+        prihdr["OBSNAME"] = prihdr['OBSTYPE']
+        prihdr['PHTCNT'] = False
+        # Update FITS file
+        fits_file.writeto(file, overwrite=True)
+
+
 @pytest.mark.e2e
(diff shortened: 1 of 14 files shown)


tests/e2e_tests/astrom_e2e.py
```diff
@@ -14,6 +14,33 @@
...
+        # Adjust VISTYPE
+        prihdr['OBSNUM'] = prihdr['OBSID']
+        exthdr['EMGAIN_C'] = exthdr['CMDGAIN']
+        exthdr['EMGAIN_A'] = -1
+        exthdr['DATALVL'] = exthdr['DATA_LEVEL']
+    # exthdr['KGAINPAR'] = exthdr['KGAIN']
+        prihdr["OBSNAME"] = prihdr['OBSTYPE']
+        prihdr['PHTCNT'] = False
+        # Update FITS file
+        fits_file.writeto(file, overwrite=True)
```
juliamilton, line 41: Using overwrite=True when saving files causes issues for my implementation (and potentially others?) since I read everything directly from Box, due to the folders being quite large and also wanting to use the original copies. This setup results in the original shared files being overwritten.

Would it be possible to save these modified files separately to avoid altering the originals? Alternatively, we could consider moving the necessary test files to the repo and adding them to .gitignore.

semaphoreP: This would have the consequence of potentially more than doubling the space that these test files would take on a computer. I'm not sure that is particularly better in terms of practicality. 

In the future, I'd like to move all the data from Box, and have a code download directly from a version-controlled server, which would remove this issue.

semaphoreP: I think the Box is already being deprecated. We're currently hosting the data on sharepoint (as still a temporary solution)

semaphoreP: Could you instead make a copy of the original files and run the tests on those instead?

juliamilton: Sounds good. In that case, perhaps worth noting to people to not run these tests locally while reading directly from Box, otherwise they will overwrite the original files and the headers (and these tests will start running into issues). I was trying to get around this by setting up a new folder on Box with 'tvac data updated headers' for people to use in these tests, which I still could do if you think it is risky otherwise.

juliamilton: > Could you instead make a copy of the original files and run the tests on those instead?

Oops, responded before I saw this. Yes, I can. It's a bit annoying because the relevant folders on box at this point are literally too big for me to download locally and so I have to go through them and individually identify the files referenced in these tests and download those, but I can do it.

juliamilton:
Thank you @semaphoreP!! Will merge and resolve the comments from the existing PR on my branch

tests/e2e_tests/astrom_e2e.py
semaphoreP: Ok, I'll update the README about not directly messing with the Box data. 

And we'll look into shrinking the total directory size. I don't think everything is needed to run the e2e tests

Issue #275, opened 2024-12-20
Core throughput: CTC 1090883


hsergi:
1090883 - Given 1) an array of PSF pixel locations and 2) the location of the center of the FPAM coronagraphic mask in EXCAM pixels during core throughput calibrations, and 3) corresponding core throughputs for each PSF, the CTC GSW shall compute a 2D floating-point interpolated core throughput map.

See also: https://github.com/roman-corgi/corgidrp/issues/6

hsergi:
By default, 1090883 will consider a square grid of (x,y) values for the core throughput map. The grid size will be an argument to the function. It will also allow the user to parse an optional, pre-defined set of points.

**Input data**

- Core throughput cal file providing:

> - The center of the FPAM mask (two values) with respect to EXCAM's pixel (0,0) during core throughput observations 
> - Nx2-dimensional array PSF pixel locations (two values per PSF) with respect to pixel (0,0) within the area of the HLC mask
> - Nx1-dimensional array of core throughput values between (0,1]. One value for each PSF


- Target pixels for the core throughput map:

> Default option: square grid defined by x_range = [xmin,xmax], y_range = [ymin,ymax], n_gridx = number_of_x_gridpoints, n_gridy = number_of_y_gridpoints
> - Alternatively, allow to parse a user-defined list of positions

**Method**

- Use bi-linear 2D interpolation method and scipy interpolation method griddata for sparse data

**Output**

- A core throughput map. That is, a 3-dimensional array (x,y,ct) where (x,y) is the position of each target pixel location wrt the FPM center, and ct  is the interpolated core throughput value at each location
- Return the core throughput map only at valid locations. That is, if some location(s) lies outside the input CT dataset, remove it from the output
- If all locations fall outside the area defined by the input CT dataset, raise Exception.
- Add this rule in the function docstring

hsergi:
This function uses well-known python libraries. We use an interpolation method from scipy (interpolated.griddata). If we assume that this interpolation method has been validated, the unit tests could be:

1/ First, use OS11 simulations to estimate a PSF at one of its locations, using the N-1 OS11 PSFs. Compare the CT of the OS11 PSF with its interpolated result. Repeat for a few locations b/w 6-9 l/D. The (expected small) tolerance will be dictated by the results. Again, one will have to make sure that there are no other minor bugs looking at the code's implementation and scipy documentation for this interpolation function.

[...]

Issue #6, opened 2023-12-15
Create core throughput correction calibration files


semaphoreP:
- Simulate data scanning a star across the FOV to map the core throughput
- Process data to create a 2-D map of the off-axis throughput for each pixel (values between 0-1)
- Create a library of off-axis PSFs and corresponding separation/PA of each PSF
- Both of the above should go into a single calibration file
- Write a function that allows one to pass in a manual (x,y) and obtain the interpolated core throughput correction at that location
- Write a function that allows one to pass in a manual (x,y) and obtain the interpolated PSF at that location

maxwellmb:
A good place to start is reviewing @rzellem's paper: https://ui.adsabs.harvard.edu/abs/2022SPIE12180E..1ZZ/abstract

hsergi:
Marie and I have started to work on this and after considering the different tasks, the amount of work to be done, especially by myself, the deliveries, and due dates have come up with the following action plan:

We will first deal with satisfying the CTC requirements 1090881, 1090882, 1090883 and 1090884,  as per the JPL FDD: [https://collaboration.ipac.caltech.edu/display/romancoronagraph/JPL+Functional+Design+Descriptions#JPLFunctionalDesignDescriptions-CalibrateCoreThroughput](https://collaboration.ipac.caltech.edu/display/romancoronagraph/JPL+Functional+Design+Descriptions#JPLFunctionalDesignDescriptions-CalibrateCoreThroughput)
These must be delivered and are necessary to close this issue.

Thus, we will open new issues on GitHub to deal with each of them: Write the functions within a core throughput calibration module, write and document their UTs, submit their PRs, etc until satisfied. Once they are satisfied, we may find that some of the tasks of this issue have been addressed and we can focus on the remaining ones.

In parallel, we will submit a plan to address the tasks in this issue that are not directly addressed by the CTC requirements. The plan will consist of a minimum set of simulated data and steps to complete this issue. As time permits, we may add more complexity.

If you have any questions please iterate on Slack to avoid cluttering GitHub.

hsergi:
These are the inks to each CTC req related to CT:

https://github.com/roman-corgi/corgidrp/issues/273
https://github.com/roman-corgi/corgidrp/issues/274
https://github.com/roman-corgi/corgidrp/issues/275
https://github.com/roman-corgi/corgidrp/issues/276

hsergi:
[...]

PR #331, opened 2025-03-15
CT Interpolation

hsergi:
## Describe your changes
I have added the capability of interpolating the CT values to an arbitrary set of positions on EXCAM during coronagraphic observations. See https://github.com/roman-corgi/corgidrp/issues/275

## Type of change

Added some methods to CoreThroughputCalibration and to corethroughput.py

-https://github.com/roman-corgi/corgidrp/issues/275
-https://github.com/roman-corgi/corgidrp/issues/6

- [x ] I have linted my code
- [ x] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [ x] I have verified that all docstrings are properly formatted and added new documentation, as needed
- [ x] I have filled out the Unit Test Definition Table on confluence, as needed


maxwellmb (dismissed):

maxwellmb (changes requested):
I think this is good to go with the exception of one test: having different PAM H and V positions, such that the shift is non-zero. If that's already in the test and I've misunderstood it would be great to write it explicitly that this is happening. It's admittedly a bit hard to follow all the datasets flowing around.

hsergi:
We had the test of the FPM shifts in the previous PR, so I think we did not consider it a must to have it again. The setup module describes each data set's rationale. I do not mind adding it though and add some more lines to clarify what is being done within the tests themselves.

hsergi:
I tested it with different values of PAM H and V positions though I left one case only given that the test looks to me already with many steps. I can add a loop over PAM values if needed.

corgidrp/data.py
```diff
@@ -1757,6 +1757,7 @@ def InterpolateCT(self,
        
         # Raise ValueError if CT < 0, CT> 1
         if np.any(interpolated_values < 0) or np.any(interpolated_values > 1): 
+            breakpoint()
```
maxwellmb, line 1760: should this be there?

maxwellmb (changes requested):
One last thing

corgidrp/data.py
hsergi: No! I thought I had removed it!

maxwellmb (approved):

Issue #240, opened 2024-10-29
Measure properties of detected companions in PSF-subtracted coronagraphic images


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Want to measure the position, flux ratio, and apparent magnitude of any point sources in a coronagraphic image

**Describe the solution you'd like**
A step function, given a list of detected sources listed in the header, that measures precise positions and fluxes (may require forward modeling, may be able to get away with regular PSF fitting for RDI)

**Additional context**
Could use code from pyklip/spaceklip as a start

maxwellmb:
Linking Issue #328, because there was useful discussion there. 

@juliamilton, I think you can assume that the APP_MAG header keyword is already populated, but to implement it as @semaphoreP has described I think you may need to assume that there's an off-axis stellar image in the dataset and we'll have to be able to identify that.

semaphoreP:
I would say let's get something working, and then we add contingencies? We could probably produce a calspec file for each of our stars given that we won't be looking at very many stars.

semaphoreP:
implementation complete in #364 with the acknowledgement we can improve it (See #377)

PR #337, opened 2025-03-18
Properties of companions in PSF-subtracted coronagraphic images

juliamilton:
## Describe your changes
This is an initial PR for to address issue #240. The procedure implemented in the code is described below:

1. Calculate off-axis image of host star if it was at the same position as the companion

- Reference PSF: assume off-axis PSF with highest CT has negligible effect from the mask (it is near 6 lam/D), or use simulated PSF w/o FPM masks
- Measure the photometry of the reference PSF in some way (aper phot, gauss fit, etc.)
- Measure the photometry of the direct image of the host star in the same way as above
- Compute the ratio of photometry_host_star/photometry_reference_PSF
- Get/interpolate the off-axis PSF at the separation of the companion
- Scale the off-axis PSF by the computed ratio above to get the off-axis PSF of the host star if it was at the same position as the companion

2. Measure flux ratio of companion to host star

- Have:  either coronagraphic dataset with companion (if forward modeling needed) or PSF-subtracted image (if no forward modeling needed, see Note)
- Have: Off-axis image of host star if it was at the same position as the companion
- if needed- forward model the off-axis image of the host star through the PSF subtraction process
- fit PSF (forward modeled or not) to data after PSF subtraction

## Type of change

#240

- [ ] I have linted my code


corgidrp/measure_companions.py
```diff
@@ -0,0 +1,450 @@
...
+    # 2) Prepare the 'psfs' argument for FMPlanetPSF
+    #    off_axis_psf might be 2D or 3D (if multi-wavelength)?
+    psf_array = off_axis_psf.data 
+    dataset_shape = coronagraphic_dataset.input.shape 
+    dataset_shape = coronagraphic_dataset.input.shape[-2:]
+
+    # May (or may not) need multiple wavelength channels:
+    # For now assume single-channel data:
+    # Otherwise, replace wvs=[some_array_of_wavelengths].
+    wvs = [1.0]  # placeholder single wavelength
```
semaphoreP, line 300: I don't think we'll need multiple wavelengths, so can always fake to 1 wavelength (unlikely we'll use the same function for spectroscopy).

tests/test_measure_companions.py
```diff
@@ -0,0 +1,156 @@
...
+        psf_sub_image=psf_sub_frame, coronagraphic_dataset=coron_data,
+        phot_method='aperture',
+        ct_cal=ct_cal, ct_dataset=dataset_ct, FpamFsamCal=FpamFsamCal,
+        fluxcal_factor=fluxcal_factor,
+        forward_model=False, direct_star_image= direct_star_image,
+        reference_psf=reference_psf, output_dir = out_dir, verbose=True
+    )
+
+    print("\nResult Table:\n", result_table)
+    assert len(result_table) == 2, "Expected 2 measured companions"
```
semaphoreP, line 151: I assume we'll eventually want to check that the flux/magnitude we get is as expected.

tests/test_measure_companions.py
```diff
@@ -0,0 +1,156 @@
...
+
+    coron_data = mocks.generate_coron_dataset_with_companions(
+        n_frames=1, shape=(200, 200), companion_xy=[(120, 80), (90, 130)], companion_counts=[host_star_counts / 2, host_star_counts / 3],
+        host_star_counts=host_star_counts, roll_angles=[10.0], add_noise=True, noise_std=1e-8, outdir=out_dir
+    )
+
+    return direct_star_image, host_star_counts, fluxcal_factor, zero_point, dataset_ct, ct_cal, FpamFsamCal, psf_sub_frame, coron_data
+
+
+def test_measure_companions_wcs(out_dir):
```
semaphoreP, line 127: pytest tests shouldn't have unspecified args, so we should default to a test dir

corgidrp/measure_companions.py
```diff
@@ -0,0 +1,450 @@
...
+            psf_guess_counts, _ = measure_counts(scaled_psf_at_companion_loc, phot_method, None, 
+                                                                 **phot_method_kwargs)
+
+            print("checking star loc", x_star, y_star)
+
+            fit = forward_model_psf(coronagraphic_dataset, scaled_psf_at_companion_loc, guesssep, guesspa,
+                                    psf_guess_counts, outputdir=output_dir)
+            print("forward model fit", fit)
+        else:
+            psf_sub_star_at_companion_loc, efficiency = classical_psf_sub(scaled_psf_at_companion_loc)
```
semaphoreP, line 132: are we really doing classical PSF subtraction, or we are just passing in any psf-subtracted frame (`psf_sub_image`) and measuring the flux relative to the `psf_sub_star_at_companion_loc`? 

It's totally ok if we're not. Directly measuring the flux of the companion and ignoring algorithm throughput is what I was basically thinking for a first pass (and we could use Ell's algorithm throughput term that they are writing into the output of the L4 PSF-subtracted image).

semaphoreP (commented):
Nice work! This looks in really good shape already. Happy to help debug pyklip further. There's a bunch of small minor comments. 

For classical PSF subtraction, I think we might have used the wrong term here. (maybe I was the first to accidentally use it) Basically, I think (if I understand your code correctly) we're just taking the L4 data product and measuring the flux of the companion in that L4 data product. I think we just need to add Ell's algorithm throughput that they compute here. I think this is totally good and valid. I think it's not exactly "classical PSF subtraction".

corgidrp/measure_companions.py
```diff
@@ -0,0 +1,516 @@
...
+                - counts_ratio : counts_corr / host_star_counts (if host_star_counts is given)
+                - mag : apparent magnitude (computed via host_star_apmag or zero point or fluxcal_factor)
+    """
+    phot_method_kwargs = get_photometry_kwargs(phot_method)
+
+    # i. Select reference PSF: assume off-axis PSF with highest CT has negligible effect from the mask 
+    # (it is near 6 lam/D see Figs. 9 and 11 in John Krist's paper), or use simulated PSF w/o FPM masks
+   
+    # ii. Measure host star counts from direct image 
+    host_star_counts, _ = measure_counts(direct_star_image, phot_method, None, **phot_method_kwargs)
```
semaphoreP, line 82: if the direct_star_image is taken with the ND (almost certainly), probably need to correct for the transmission of the ND first, before anything else.

corgidrp/measure_companions.py
```diff
@@ -0,0 +1,516 @@
...
+    ct_cal,
+    phot_method='aperture',
+    coronagraphic_dataset=None,
+    fluxcal_factor=None,
+    FpamFsamCal = None,
+    host_star_in_calspec=True,
+    forward_model=False,
+    output_dir = ".",
+    verbose=True
+):
```
semaphoreP, line 34: After discussion with Vanessa, let's assume that host star apparent magnitude is an input to this function as a float. So assume someone else has calculated it for you.

semaphoreP (commented):
A couple more higher-level comments based on discussions with Vanessa

Issue #241, opened 2024-10-29
Measure fluxes of sources in unocculted dataset


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Get the apparent magnitude of sources in an unocculted image

**Describe the solution you'd like**
A step function that does the above. Likely needs to take as input the absolute flux calibration

JuergenSchreiber:
This should to be implemented straight forward after merging of ND Filter calibration product #279

semaphoreP:
Implemented in #347

PR #347, opened 2025-03-20
Measure flux

JuergenSchreiber:
## Describe your changes
added function determine_flux to l4_to_tda. This function calculates the absolute flux of a source using the fluxcalFactor calibration file. It calculates the sum of photoelectrons/s + error + local background for each image target and averages the found values (realistic scenario at this stage of the pipeline?). It also converts the flux to VEGA apparent magnitudes and writes everything in the header -> FLUX, FLUXERR, LOCBACK, APP_MAG 

## Type of change

#241

- [x] I have linted my code


JuergenSchreiber:
I should think about a better handling of the kwargs, input only parameters to change, rest should stay default, I guess somebody did that already

nasavbailey:
I have one more question about units. fluxcal.py says it expects input images to already be normalized by exposure time. If that's the case, I think it should check that BUNIT is electrons/s and throw an error if not?

corgidrp/data.py
```diff
@@ -1375,8 +1375,8 @@ def __init__(self, data_or_filepath, err = None, pri_hdr=None, ext_hdr=None, err
                 orig_input_filename = input_dataset[0].filename.split(".fits")[0]
   
             self.ext_hdr['DATATYPE'] = 'FluxcalFactor' # corgidrp specific keyword for saving to disk
-            self.ext_hdr['BUNIT'] = 'erg/(s * cm^2 * AA)/electron'
-            self.err_hdr['BUNIT'] = 'erg/(s * cm^2 * AA)/electron'
+            self.ext_hdr['BUNIT'] = 'erg/(s * cm^2 * AA)/electron/s'
```
semaphoreP, line 1378: (electron/s)?

corgidrp/l4_to_tda.py
```diff
@@ -139,6 +139,81 @@ def convert_to_flux(input_dataset, fluxcal_factor):
     flux_dataset.update_after_processing_step(history_msg, new_all_data=flux_cube, header_entries = {"BUNIT":"erg/(s*cm^2*AA)", "FLUXFAC":fluxcal_factor.fluxcal_fac})
     return flux_dataset
 
+
+def determine_flux(input_dataset, fluxcal_factor,  photo = "aperture", phot_kwargs = None):
```
semaphoreP, line 143: can we have an optional x/y input to specify where to do the photometry that overrides finding the brightest source? If you don't have time to do this tomorrow, please make an issue for a future feature addition

corgidrp/l4_to_tda.py
```diff
@@ -139,6 +139,81 @@ def convert_to_flux(input_dataset, fluxcal_factor):
...
+    flux = ap_sum * factor
+    flux_err = np.sqrt(ap_sum_err**2 * factor**2 + factor_error**2 *ap_sum**2)
+    #Also determine the apparent Vega magnitude
+    filter_file = fluxcal.get_filter_name(flux_dataset[0])
+    vega_mag = fluxcal.calculate_vega_mag(flux, filter_file)
+    
+    history_msg = "star {0} flux calculated as {1} erg/(s * cm^2 * AA) corresponding to {2} vega magnitude".format(flux_dataset[0].pri_hdr["TARGET"],flux, vega_mag)
+
+    # update the output dataset with this converted data and update the history
+    flux_dataset.update_after_processing_step(history_msg, header_entries = {"FLUXFAC": fluxcal_factor.fluxcal_fac, "LOCBACK": back, "FLUX": flux, "FLUXERR": flux_err, "APP_MAG": vega_mag})
```
semaphoreP, line 213: would it be easy to calculate the error on the app mag? if not, please make an issue for us to do it later.

semaphoreP (commented):
I think it's looking good. a few minor comments that aren't blocking, please either address or make new issues for us to address later.

corgidrp/l4_to_tda.py
JuergenSchreiber: I think @juliamilton has done this already in her PR 337 in the relevant fluxcal functions, so there should appear a merge conflict  which can then be solved by doing so? Does this make sense?

corgidrp/data.py
JuergenSchreiber: Generally, I think at one point we should review all our used BUNITs and adjust them to some standards, like astropy.units. E.g. sometime we say electrons instead of electron or microns instead of micron etc.

semaphoreP:
> > I have one more question about units. fluxcal.py says it expects input images to already be normalized by exposure time. If that's the case, I think it should check that BUNIT is electrons/s and throw an error if not?
> 
> Hi Vanessa, I could do this, but as far as I know in Corgidrp we never check units or if certain pipeline steps are run before a certain step. I think we just assume that when we start from LX level (L4 in this case) everything was done correctly before. Maybe @semaphoreP could you please comment on this?

I think most other steps don't depend on the units of the data, whereas this one does, so it's not a bad thing to check. Let's add it if it's just a few lines of work? If it's too complicated, feel free to make an issue for future release.

corgidrp/data.py
semaphoreP: Can you make a Github issue?

semaphoreP:
@JuergenSchreiber, let me know how you want to proceed with Vanessa's comment and fix merge conflicts (from your other PR). Then I think we should be good.

JuergenSchreiber:
> @JuergenSchreiber, let me know how you want to proceed with Vanessa's comment and fix merge conflicts (from your other PR). Then I think we should be good.

I opened #354

semaphoreP (approved):

Issue #238, opened 2024-10-29
Compute calibrated contrast curve


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Need a step function to produce a calibrated contrast curve for a PSF-subtracted coronagraph dataset

**Describe the solution you'd like**
A step function that uses the PSF subtracted frame and its algorithm throughput vs separation to produce a calibrated contrast curve, also accounting for the throughput of the coronagraph

semaphoreP:
This should be a 1-sigma contrast curve

PR #350, opened 2025-03-20
KLIP Throughput Calibration

ell-bogat:
## Describe your changes:

This PR adds the ability to calculate a 1D KLIP throughput curve and core throughput curve as a function of separation during the PSF subtraction step function. These arrays will be used during L4-to-TDA to calculate the calibrated contrast curve.

The 1D KLIP throughput is calculated by injecting model PSFs (pulled from the CoreThroughputCalibration object associated with the dataset), performing KLIP subtraction using the same settings as the main dataset used, and recovering the flux of each injected PSF after subtraction. This gets saved to the hdu list extension of the output Image object.

The 1D core throughput is calculated by simply selecting the core throughput value in the calibration object nearest to the desired separation. We do this for each of the separations sampled in the KLIP throughput and then save it to the hdu list extension of the output Image object.

Here is an example of the result following ADI KLIP subtraction of data with fake injected PSFs:
image

And the associated KLIP throughput curve: 
image

### The following functions have been added:
- Some helper functions in `astrom.py`:
  - `get_polar_dist()` gets the linear separation in pixels between two points measured in polar coordinates.
  - `seppa2dxdy()` converts from a position in separation/position angle to x- and y- separation from the center in pixels.
  - `seppa2xy()` converts from a position in separation/position angle to x- and y- position in the array in pixels.
- A function to get the 1D core throughput in `corethroughput.py`:
  - `get_1d_ct()`: reads the CT calibration object and grabs the core throughput value measured at the positions nearest to the desired separations. There's infrastructure to swap in alternate methods, e.g. interpolation, in the future.
- Functions in `klip_fm.py` to measure the KLIP throughput:
  - `get_closest_psf()` fetches the PSF model in the CT cal object closest to a desired position.
  - `inject_psf()` inserts a psf model into a corgidrp Image, with a desired amplitude at a desired position. Right now it drops the PSF in centered on the nearest pixel center, and we'll need to add the subpixel shift soon. 
  - `measure_noise()` measures the standard deviation of counts in an image at a variety of separations from the center star.
  - `meas_klip_thrupt()` is the master function that calculates the KLIP throughput. These are the steps:
    1. Pick the separations to sample, if not provided. This is done by starting at the IWA, then stepping out by some multiple of the FWHM until you get to the OWA. (default spacing is 5 FWHMs).
    2. Measure the noise in the psf-subtracted images at each separation using `measure_noise()`
    3. Inject PSFs into the pre-subtracted images at an amplitude of 5 * the noise level at each separation (using `inject_psf()` and `get_closest_psf()`), and hold onto copies of the PSF model to refer to later.
    4. Do PSF subtraction using the same KLIP parameters the same way as for the main dataset.
    5. Measure and subtract a constant background level via sigma clipping (sigma threshold=3). 
    6. For each injected planet, fit a 2d gaussian to the psf model that we saved earlier and to the planet after KLIP subtraction. This is done using `pyklip.fakes.gaussfit2d()`.
    7. Calculate the total counts that the 2d gaussian fits correspond to, since they might have different FWHMs: `counts = np.pi * amplitude * fwhm**2 / 4. / np.log(2.)`. 
    8. Record the ratio of counts after PSF subtraction to counts before PSF subtraction.
    9. Take the mean of the KLIP throughputs at each separation.
    10. Repeat for each number of output KL modes.

### The following functions have been modified:
- `l3_to_l4.do_psf_subtraction()` now has toggles to measure the KLIP throughput and/or measure the core throughput, both of which default to `True`. 
  - if `measure_klip_thrupt`: After the main PSF subtraction is done, the pre-subtracted data, post-subtracted data, and core throughput calibration object are passed to `klip_fm.meas_klip_thrupt()` and the result gets recorded in the output dataset's hdu_list extension. 
  - if `measure_1d_core_thrupt`: Similarly, the core throughput is measured at the very end via `corethroughput.get_1d_ct()` and also appended to the hdu_list. 

### I have some open questions for discussion:
- SNR calculation. Currently using SNR 20 and saying amplitude of peak flux is 20 * noise (stdev) value at that separation.
- Need to include small number statistics correction in noise calculation?
- PSF fitting is sensitive to local background level -> probably need to measure it locally and subtract before fitting, or find a way to fit the background level simultaneously. I don't think the pyklip function I'm using right now can do this.
- How closely can we space the psfs we inject? Is there a way to figure this out other than empirically testing it and seeing when the measured KLIP throughput starts to increase due to contamination?
- PSF injection currently snaps to the nearest pixel center. Do we need to do the sub-pixel shift for this PR? I don't imagine we expect the throughput to change that much over the span of 1 pixel.
- How to preserve knowledge of star/mask location in full science array after cropping since we need this to get the right PSF from the CT cal object (save "pixel 0" location so you can just add it to a pixel location in the smaller array?)
- The error tolerance on the upper limit we accept for KLIP throughput (aka it shouldn't be greater than 1 within the noise) I'm currently calculating via the per-pixel noise standard deviation * the twice PSF FWHM**2. This is somewhat arbitrary, I'm currently seeing max throughput (at the widest separations) for both ADI and RDI around 1.02. This could also be related to the background subtraction issue.
- Do we need to calculate the error on the KLIP or core throughput measurements?
- In different sections I refer to `STARLOCX/Y` (typically after psf subtraction and during planet injection) or `MASKLOCX/Y` (typically when grabbing core throughput things). Is there a rule for when to use which keyword, or are they going to be so close together that it shouldn't matter for this step?
- Repeating the measurement for each output KL mode takes a long time. The noise is slightly different for each number of modes, so you have to do all the KLIP subtraction over again. In the interest of computation time, do we expect the noise to vary all that much? Do we want to default to just calculating the throughput curve for the first of last number of KL modes?
- The discrepancy between the result and the analytical solution for the ADI-only PSF subtraction test (not including KLIP injection-recovery stuff) increased with this PR to be about 3% off from the expected peak. Is this acceptable given that this isn't a super analytical process and we're calibrating the throughput anyway? 

### And these are the things I know I need to do before merging:
- [x] Reduce injection SNR in tests from 20 to 5 and make sure the throughput results are still reasonable.
- [x] Test psf subtraction + automated cropping + injection recovery (a.k.a. all the default settings of the PSF subtraction step functions)
- [x] Give tests more informative error messages.
- [x] Add tests for `astrom.get_polar_dist()`.
- [x] Expand tests for `klip_fm.injec_psfs()`.
- [x] Read in or calculate IWA, OWA, and FWHM from somewhere in the headers instead of hard coding.
- [ ] Update unit testing and data format documentation pages.

## Type of change
Addresses part of Issue #238 

- [x] I have linted my code
- [ ] Update unit testing and data format documentation pages on Confluence


ell-bogat (commented):
Starting the draft PR for CI testing while I fiddle with things!

semaphoreP:
* SNR calculation. Currently using SNR 20 and saying amplitude of peak flux is 20 * noise (stdev) value at that separation.
Need to include small number statistics correction in noise calculation?

Yes, eventually, can be an issue for the future. 

 * PSF fitting is sensitive to local background level -> probably need to measure it locally and subtract before fitting, or find a way to fit the background level simultaneously. I don't think the pyklip function I'm using right now can do this.

Let's deal with this later. Your local background looks close to 0, so you really shouldn't be affected by it.

 * How closely can we space the psfs we inject? Is there a way to figure this out other than empirically testing it and seeing when the measured KLIP throughput starts to increase due to contamination?

The relevant quantity is based on the FWHM (for a given radius, fwhm, and fwhm spacing, we know how many planets we can inject). Can also be a future improvement. You clearly have something working.

 * PSF injection currently snaps to the nearest pixel center. Do we need to do the sub-pixel shift for this PR? I don't imagine we expect the throughput to change that much over the span of 1 pixel.

Throughput shouldn't change much across 1 pixel. Ok for now. Just need to interpolate and inject. 

 * How to preserve knowledge of star/mask location in full science array after cropping since we need this to get the right PSF from the CT cal object (save "pixel 0" location so you can just add it to a pixel location in the smaller array?)

If you crop down to so that [250, 260] becomes [0,0], subtrat (250,260) from star and mask centers.

 * The error tolerance on the upper limit we accept for KLIP throughput (aka it shouldn't be greater than 1 within the noise) I'm currently calculating via the per-pixel noise standard deviation * the twice PSF FWHM**2. This is somewhat arbitrary, I'm currently seeing max throughput (at the widest separations) for both ADI and RDI around 1.02. This could also be related to the background subtraction issue.

Let's say we want the throughput computed to 5% for now (I just made up a number, but we have to start somewhere). We can always inject for fake planets to improve the uncertainty in the throughput. 

 * Do we need to calculate the error on the KLIP or core throughput measurements?

Sounds like a future enhancement. I don't see it as a requirement. 

 * In different sections I refer to STARLOCX/Y (typically after psf subtraction and during planet injection) or MASKLOCX/Y (typically when grabbing core throughput things). Is there a rule for when to use which keyword, or are they going to be so close together that it shouldn't matter for this step?

Star should be used for alignment and derotation. Mask center I think is only relevant for grabbing core throughput. So I think you are ok. You are right that likely it won't matter in practice. 

 * Repeating the measurement for each output KL mode takes a long time. The noise is slightly different for each number of modes, so you have to do all the KLIP subtraction over again. In the interest of computation time, do we expect the noise to vary all that much? Do we want to default to just calculating the throughput curve for the first of last number of KL modes?

There's a chance the noise will vary a decent amount. Maybe by default, we just test the code for a single value for numbasis? (i.e., it could loop over, but that's not how we unit test/set default). 

 * The discrepancy between the result and the analytical solution for the ADI-only PSF subtraction test (not including KLIP injection-recovery stuff) increased with this PR to be about 3% off from the expected peak. Is this acceptable given that this isn't a super analytical process and we're calibrating the throughput anyway?

Would the KLIP injection-recovery account for it?

ell-bogat:
Thanks so much @semaphoreP ! 

I'll work on a test to see how closely we can recover the original injected planet flux based on gaussian psf fitting and multiplying by the measured throughput, but my intuition is that if we are applying the same process to both images then it should be pretty close? I'll let you know what I find out!

corgidrp/klip_fm.py
```diff
@@ -0,0 +1,549 @@
...
+        contains the KLIP throughput measured at each separation for each KL mode truncation choice.
+    """
+    
+    if sci_dataset_in[0].ext_hdr['CFAMNAME'] == '1F':
+        lam = 573.8e-9 #m
+    else:
+        raise NotImplementedError("Only band 1 observations using CFAMNAME 1F are currently configured.")
+
+    d = 2.36 #m  
+    pixscale_mas = 0.0218 * 1000 # mas
```
semaphoreP, line 248: isn't PLTSCALE written into the header by the `create_wcs` step that we can use?

semaphoreP (commented):
Looking good Ell! Only a few small comments/questions.

corgidrp/klip_fm.py
ell-bogat: Fixed!

corgidrp/klip_fm.py
```diff
@@ -0,0 +1,549 @@
...
+        
+        seps = np.arange(iwa_pix,owa_pix,res_elem) # Some linear spacing between the IWA & OWA, around 5x the fwhm
+    if pas is None:
+        pas = np.linspace(0.,360.,n_pas+1)[:-1] # Some linear spacing between the IWA & OWA, around 5x the fwhm
+
+    thrupts = []
+    for k,klmode in enumerate(klip_params['numbasis']):
+        
+        sci_dataset = sci_dataset_in.copy()
+        ref_dataset = ref_dataset_in.copy() if not ref_dataset_in is None else None
```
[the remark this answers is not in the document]
ell-bogat: Also true, removed this line and now we pass ref_dataset_in directly to do_psf_subtraction()

corgidrp/l3_to_l4.py
```diff
@@ -538,9 +576,70 @@ def do_psf_subtraction(input_dataset, reference_star_dataset=None,
     dataset_out = nan_flags(dataset_out,threshold=1)
     
     history_msg = f'PSF subtracted via pyKLIP {mode}.'
-    
     dataset_out.update_after_processing_step(history_msg)
+    
+    if measure_klip_thrupt:
+        
+        # Determine flux of objects to inject (units?)
+        inject_snr = 20.0
```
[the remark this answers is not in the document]
ell-bogat: Yep! Made that change.

corgidrp/l3_to_l4.py
```diff
@@ -538,9 +578,69 @@ def do_psf_subtraction(input_dataset, reference_star_dataset=None,
...
+                            klip_params,
+                            kt_snr,
+                            cand_locs = cand_locs, # list of (sep_pix,pa_deg) of known off axis source locations
+                            seps=kt_seps,
+                            pas=kt_pas
+                            )
+        thrupt_hdr = fits.Header()
+        # Core throughput values on EXCAM wrt pixel (0,0) (not a "CT map", which is
+        # wrt FPM's center 
+        thrupt_hdr['COMMENT'] = ('KLIP Throughput as a function of separation for each KLMode '
```
semaphoreP, line 606: can we update the headers here to reflect we are saving fwhms?

semaphoreP (approved):

PR #361, opened 2025-03-24
Contrast curve

kjl0025:
## Describe your changes
Added a step function which uses the PSF-subtracted frame and its algorithm throughput vs separation to produce a calibrated 1-sigma contrast curve, also accounting for the throughput of the coronagraph.

## Type of change

#238 

- [ ] I have linted my code


corgidrp/l4_to_tda.py
```diff
@@ -142,6 +145,48 @@ def convert_to_flux(input_dataset, fluxcal_factor):
...
+        sep_spacings = separations - np.roll(separations, 1)
+        # ignore the meaningless first entry (b/c of looping around with np.roll)
+        min_spacing = np.min(sep_spacings[1:])
+        if halfwidth is None:
+            halfwidth = min_spacing/2
+        check.real_positive_scalar(halfwidth, 'halfwidth', ValueError)
+        if halfwidth > min_spacing/2:
+            warnings.warn('Halfwidth is wider than half the minimum spacing between separation values.')
+        annular_noise = measure_noise(frame, separations, halfwidth)
+        contrast_curve_vals = klip_tp*core_tp*annular_noise.T
```
semaphoreP, line 181: I think you need to divide by the throughput terms, not multiply by them.

corgidrp/l4_to_tda.py
```diff
@@ -142,6 +145,48 @@ def convert_to_flux(input_dataset, fluxcal_factor):
...
+        contrast_curve = np.vstack([separations, contrast_curve_vals])
+        hdr = fits.Header()
+        hdr['BUNIT'] = "erg/(s*cm^2*AA)"
```
semaphoreP, line 184: we want to measure "flux ratio noise" in the end, so we need to divide by the flux of the star. The units should be "Fp/Fs".

At the end of L4, the data is still in counts of photoelectrons/s, so the noise you measure isn't in ergs/(s*cm^2*AA). 

Converting the L4 data product to Fp/Fs requires being careful with the units. You are measure the noise per pixel in photoelectrons/s. To get Fp/Fs, you need to divide by the peak pixel value of the star in photoelectrons/s. This can be done with the unocculted image of the host star taken behind the ND filter, the transmission of the ND filter, and fitting a 2-D gaussian to it to get the peak flux.

corgidrp/l4_to_tda.py
```diff
@@ -142,6 +145,48 @@ def convert_to_flux(input_dataset, fluxcal_factor):
...
+        frame.add_extension_hdu('CON_CRV', data = contrast_curve, header=hdr)
```
semaphoreP, line 185: contrast curve usually implies 5-sigma equivalent false positive fraction curve. Let's just call it what it is. a noise curve.

corgidrp/l4_to_tda.py
```diff
@@ -142,6 +145,48 @@ def convert_to_flux(input_dataset, fluxcal_factor):
     return flux_dataset
 
 
+def compute_calibrated_contrast_curve(input_dataset, halfwidth=None):
```
semaphoreP, line 148: there should be an option to pass in a requested list of separations. and we do interpolation with the KL_THRU and CT_THRU curves to get those values at the requested separations.

corgidrp/l4_to_tda.py
```diff
@@ -142,6 +145,48 @@ def convert_to_flux(input_dataset, fluxcal_factor):
...
+        # ignore the meaningless first entry (b/c of looping around with np.roll)
+        min_spacing = np.min(sep_spacings[1:])
+        if halfwidth is None:
+            halfwidth = min_spacing/2
+        check.real_positive_scalar(halfwidth, 'halfwidth', ValueError)
+        if halfwidth > min_spacing/2:
+            warnings.warn('Halfwidth is wider than half the minimum spacing between separation values.')
+        annular_noise = measure_noise(frame, separations, halfwidth)
+        contrast_curve_vals = klip_tp*core_tp*annular_noise.T
+        contrast_curve = np.vstack([separations, contrast_curve_vals])
```
semaphoreP, line 182: is separations in pixels? it would also be nice to also save (as another row) the separations in milliarcseconds.

corgidrp/klip_fm.py
```diff
@@ -0,0 +1,462 @@
...
+            the noise level.
+        klmode_index (int, optional): If provided, returns only the noise values for the KL mode with 
+            the given index. I.e. klmode_index=0 would return only the values for the first KL mode 
+            truncation choice.  If None (by default), all indices are returned.
+        hw (float): halfwidth of the annulus to use for noise calculation.
+
+    Returns: np.array 
+    """
+
+    cenx, ceny = (frame.ext_hdr['MASKLOCX'],frame.ext_hdr['MASKLOCY'])
```
semaphoreP, line 119: I think technically we want to compute noise centered around the star, not the mask (so STARLOCX/Y)

tests/test_contrast_curve.py
```diff
@@ -0,0 +1,108 @@
...
+                                measure_1d_core_thrupt=True)
+    
+    # getting what the contrast curve for the first frame should be:
+    frame0 = psfsub_dataset_rdi[0]
+    klip = frame0.hdu_list['KL_THRU'].data
+    separations = klip[0]
+    klip_tp = klip[1:]
+    core_tp = frame0.hdu_list['CT_THRU'].data[1]
+    annular_noise = measure_noise(frame0, separations, hw=3)
+    contrast_curve_expected = klip_tp*core_tp*annular_noise.T
```
semaphoreP, line 87: see previous comment that these should be divisions.

semaphoreP (commented):
Some comments about what we actually want for a contrast curve. See the FDD's description of "Flux Ratio Noise" in case it's more useful.

corgidrp/l4_to_tda.py
```diff
@@ -168,6 +174,147 @@ def convert_to_flux(input_dataset, fluxcal_factor):
...
+        kde = gaussian_kde(positions)
+        density = kde(positions)
+        # Find the index of the highest density
+        highest_density_index = np.argmax(density)
+        # Get the row and column of the highest-density location
+        median_row = closest_indices[0][highest_density_index]
+        median_col = closest_indices[1][highest_density_index]
+        fwhm_guess = 2*np.sqrt((median_row-guess_row)**2 + (median_col-guess_col)**2)
+
+        psf_phot = fit_2dgaussian(data, xypos=pos, fit_shape=fit_shape, fwhm=fwhm_guess, fix_fwhm=False,
```
semaphoreP, line 277: why is fit fwhm false? Are you worried about it being unstable?

tests/test_calibrate_darks_lsq.py
```diff
@@ -133,14 +133,14 @@ def test_expected_results_sub():
 
     # check headers
     assert(noise_maps.data.ndim == 3)
-    assert('DetectorNoiseMaps' in noise_maps.filename)
+    assert(noise_maps.filename.endswith('DNM_CAL.fits'))
```
semaphoreP, line 136: Can we check that the filename also is based on the last file in the input dataset?

tests/test_pumptrap_calibration.py
```diff
@@ -107,7 +107,8 @@ def test_tpump_analysis():
                         cs_fit_thresh = cs_fit_thresh, 
                         input_T=input_T,
                         bins_E=bins_E, bins_cs=bins_cs)
-
+    # filename check
+    assert tpump_calibration.filename.endswith('TPU_CAL.fits')
```
semaphoreP, line 111: Can we check that the filename also is based on the last file in the input dataset?

corgidrp/l4_to_tda.py
```diff
@@ -168,6 +174,147 @@ def convert_to_flux(input_dataset, fluxcal_factor):
...
+    produce a calibrated 1-sigma flux ratio "contrast" curve (or "noise curve" since contrast curve is typically 5-sigma), also accounting for the throughput of the coronagraph.
+    It calculates flux ratio noise curve value for each radial separation from the subtracted star location, interpolating KLIP and core throughput values at these input separations.
+    It uses a dataset of unocculted stars and ND transmission to determine the integrated flux of the Gaussian-fit star (where each frame in the dataset is assumed to correspond to the frames 
+    in the input_dataset), and the an estimate of planet flux per frame of inupt_dataset is made by calculating the integrated flux of a Gaussian with amplitude equal to 
+    the annular noise and FWHM equal to that used for KLIP algorithm througput for each radial separation.
+
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of PSF-subtracted Images
+        NDcalibration (corgidrp.data.NDFilterSweetSpotDataset): ND filter calibration
+        unocculted_star_dataset (corgidrp.data.Dataset): a dataset of unocculted star Images corresponding to the Images in input_dataset
```
semaphoreP, line 189: the docstring should specify that there should be exactly as many unocculted_star images as input_dataset images (one unocculted star image per image dataset image). 

In real world use cases: we'll likely only have 1 input dataset image and one unocculted_star_dataset image.

semaphoreP: Actually, if it's not too much work, can the input_dataset and unocculted_star_dataset both be input images instead? this would better mirror the interface of measure_companions.measuure_companions()? If it's not too much work to modify.

semaphoreP (commented):
Looks good! I just have a small request to test for the filename of the output cals, and a documentation request/API change request for compute_flux_ratio.

corgidrp/l4_to_tda.py
kjl0025: I did fix_fwhm=False so that the fitter would treat FWHM as a fitted parameter and not fix it as my input fwhm_guess.

semaphoreP: ooh misread the keyword.

tests/test_calibrate_darks_lsq.py
kjl0025: Oh yes, sorry I missed that.

corgidrp/l4_to_tda.py
kjl0025: Since it's a step function, I thought we needed a Dataset input.  But if you want an Image input, I can do that, but perhaps not today before the deadline.  I could make an issue to change it after the deadline if that's okay?

semaphoreP: yeah the TDA functions will break the normal step function rules because we aren't nominally going to pipeline them. Can be a future issue then.

semaphoreP (approved):

Issue #356, opened 2025-03-21
DRP v2 (CTC R3.0.2) data products need to conform to filename conventions


nasavbailey:
All data products delivered in DRP v2 (CTC R3.0.2) need to conform to the filename conventions described in the Data Format pages on their respective pages on the CPP wiki. https://collaboration.ipac.caltech.edu/display/romancoronagraph/Data+Format+and+Volume 

For example L3s: `CGI_<VisitID: PPPPPCCAAASSSOOOVVV>_<Last TimeUTC>_L3_.fits`

semaphoreP:
This issue seems vague. @nasavbailey can you comment on what actions you are looking for?

nasavbailey:
Based on discussion in DRP WG telecon today: Update L#, TDA, or CAL FITS file creation so the output filename follows the convention _whenever it's given an input dataset that also obeys the convention._  

For example, for L3->L4: if given a set of L3 filenames that follow CGI_<VisitID: PPPPPCCAAASSSOOOVVV>_<TimeUTC>_L3_.fits, then confirm the L4 filename is CGI_<Last science target VisitID: PPPPPCCAAASSSOOOVVV>_<Last science target TimeUTC>_L4_.fits. There should be a unit test.

Each product owner should do this for their own product. Please check the box when your PR addressing this has been merged. We're collecting all these together in this issue, rather than tracking each in a separate issues.

- [ ] L2a  @semaphoreP  
- [ ] L2b analog  @semaphoreP 
- [ ] L2b Photon Counted @kjl0025 
- [ ] L3 @maxwellmb 
- [ ] L4 PSF-subtracted @maxwellmb 
- [ ] L4 : not PSF-subtracted	same as owner of https://github.com/roman-corgi/corgidrp/issues/362 
- [ ] ABF (absolute flux) @JuergenSchreiber 
- [ ] AST (astrometry) @manduhmia 
- [ ] BPM (bad pixel map) @juliamilton 
- [ ] CTP (core throughput) @hsergi 
- [ ] DNM (detector noise maps) @kjl0025 
- [ ] FLT (flat fields) @semaphoreP 
- [ ] NDF (ND filter sweet spot) @juliamilton 
- [ ] NLN (nonlin calibrations) @hsergi 
- [ ] TPU (chargetrap/trap pumping call) @kjl0025

JuergenSchreiber:
I see also other data products like KGAIN which are not in that filename scheme, should they be added?

[...]

PR #371, opened 2025-03-26
PSF interp

hsergi:
## Describe your changes

I've added a simple PSF interpolation function as agreed in Confluence. This is the last request on [CT Cal files](https://github.com/roman-corgi/corgidrp/issues/6).

I have also added the UT related with the filename convention for the CT cal file within test_cal_file(). I have followed the name conventions in Confluence and the example in [Flat Fields PR](https://github.com/roman-corgi/corgidrp/pull/365/files#diff-3ae4c6786798dbc30deccc44453cd91c6a669ba8ec672c8e90a328fb82371054). See #356 

## Type of change

## Details

From [CT Implementation](https://collaboration.ipac.caltech.edu/display/romancoronagraph/CT+Implementation) and [DRP Test Implementation](https://collaboration.ipac.caltech.edu/display/romancoronagraph/DRP+Test+Definition) on Confluence:

Write a function that allows one to pass in a manual (x,y) wrt FPM's center and obtain the interpolated PSF at that location

There's no requirement from the FDD to satisfy related to PSF interpolation. It is an additional request from the DRP useful for future developments, recorded in https://github.com/roman-corgi/corgidrp/issues/6

First method: nearest PSF in a polar sense: Given an (x,y) position wrt FPM's center, the associated PSF is the one in the CT calibration dataset whose radial distance to the FPM's center is the closest to sqrt(x**2+y**2). If there is more than one CT PSF at the same radial distance, choose the one whose angular distance to the (x,y) location is the smallest. If there are two such PSFs (i.e., the interpolated location lies exactly between two PSF at the same radial distance, return the average of both PSFs.

Write it as part of the CoreThroughput class: GetPSF()

UT:

Build an input dataset with a partial coverage of the HLC FOV, and a set of target locations across all quadrants covering the HLC FOV (some positions can be inside the IWA or outside the OWA) to (try to) retrieve the interpolated PSF. For each target location (x,y):

If a valid location is passed ( * ) :

--Check that the returned PSF is the one whose radial distance to the FPM's center is closest to the input (x,y) radial distance
--If there's more than one such case( ** ), check the returned PSF is the one whose angular location to (x,y) is the closest
--In the previous case, if there are two such PSFs, i.e., the interpolated location's azimuth distance to the two input PSFs with the same radius is the same ( ** ), return the average of both PSFs
--If an invalid location is passed, follow the same procedure as in the CT interpolation: Return only valid locations PSF and (valid) location. If there are no valid positions at all, an exception is raised (it could be a single invalid position, or a set of all invalid positions)

( * ) Invalid positions are defined as outside the min and max radii of the polar region defined by the CT measurements. Azimuthal values are always valid since a modulus operation based on the azimuth range of the CT measurements is applied to the azimuthal coordinate of the target locations. ( ** ) In this test, we will consider that two PSFs are at the same location if their difference is less than 1/10th pf an EXCAM pixel

[CT Cal files](https://github.com/roman-corgi/corgidrp/issues/6)

- [x ] I have linted my code
- [x ] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [x ] I have verified that all docstrings are properly formatted and added new documentation, as needed
- [x ] I have filled out the Unit Test Definition Table on confluence, as needed


corgidrp/data.py
```diff
@@ -1795,6 +1796,156 @@ def InterpolateCT(self,
...
+          method (str): Interpolation method that will be used:
+              'polar-nearest': Given an (x,y) position wrt FPM's center, the
+               associated PSF is the one in the CT calibration dataset whose
+               radial distance to the FPM's center is the closest to
+               sqrt(x**2+y**2). If there is more than one CT PSF at the same
+               radial distance, choose the one whose angular distance to the
+               (x,y) location is the smallest.
+              
+        Returns:
+          Returns interpolated PSF
```
semaphoreP, line 1835: it sems like this function returns 3 variables: psf_interp_list, x_interp_list, y_interp_list. They should all be documented in the Returns part of the doc string, including dimensions. For exmaple, psf_interp_list is a list of 2-D psfs?

corgidrp/data.py
```diff
@@ -1795,6 +1796,156 @@ def InterpolateCT(self,
...
+        if len(x_cor[r_good]) == 0:
+            raise ValueError('All target radius are less than the minimum '
+                'radius in the core throughout data: {:.2f} EXCAM pixels'.format(radii.min()))
+        radius_cor = radius_cor[r_good]
+        # Update x_cor and y_cor
+        x_cor = x_cor[r_good]
+        y_cor = y_cor[r_good]
+        r_good = radius_cor <= radii.max()
+        if len(x_cor[r_good]) == 0:
+            raise ValueError('All target radius are greater than the maximum '
```
semaphoreP, line 1889: the error at this point should be all radii are either less than the minimum radius or greater than the maximum radius of the CT data. Since some may have been discarded in the first discard step.

semaphoreP (commented):
Looks good. I just have a few documentation questions for you. Otherwise the functionality looks good to me.

corgidrp/data.py
```diff
@@ -1795,6 +1796,156 @@ def InterpolateCT(self,
             x_cor[is_valid],
             y_cor[is_valid]])
 
+    def GetPSF(self,
```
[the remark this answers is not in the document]
hsergi: That's Max's choice, as for other function names in the same class.

corgidrp/data.py
hsergi: Interesting flake did not catch it

semaphoreP (approved):

PR #376, opened 2025-03-26
Cal product naming

juliamilton:
## Describe your changes
Updates ND Filter and Bad Pixel Map calibration products to have naming conventions consistent with recent documentation. Contributes to issue #356.

## Type of change

- [x] I have linted my code


juliamilton:
Hi @semaphoreP - this PR updates the ND filter and bad pixel map calibration products to have the correct naming scheme, and also updates the ND filter tests to pass with some changes recently made to the mocks.

semaphoreP:
@juliamilton do you check that the VISITID and FTIMEUTC that is used is from the last file in the input dataset that creates both calibration files?

juliamilton:
> @juliamilton do you check that the VISITID and FTIMEUTC that is used is from the last file in the input dataset that creates both calibration files?

I do that in the functions themselves that create the calibration products, not in the classes. Also, I use the last file in the input dataset by list order, but don't actually check every single file's FTIMEUTC to make sure that the last file is actually the most recent. Should I?

semaphoreP:
It's ok to not check that the last file is the most recent.

I'm mainly asking that since the unit tests don't seem to be checking for the last file's FIMEUTC and VISITID. Can we add that, if it's not too hard?

corgidrp/data.py
```diff
@@ -956,16 +957,10 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
-            # filename could be from an data level or filename oculd be from a flat field
-            base_filename = input_dataset[-1].filename
-            if "_FLT_CAL" in base_filename:
-                self.filename = base_filename.replace("_FLT_CAL", "_BPM_CAL")
-            else:
-                # not created from a flat
-                self.filename = re.sub('_L[0-9].', '_BPM_CAL', input_dataset[-1].filename)
-
+            # Save it with filename that matches documentation conventions
+            formatted_time = format_ftimeutc(self.ext_hdr['FTIMEUTC'])
```
semaphoreP, line 961: can't you just take the filename of the last file in the input dataset rather than needing to reformat this? The naming philosophy is only the last bit of the file name needs to be changed.

corgidrp/mocks.py
```diff
@@ -129,7 +129,7 @@ def create_default_L1_headers(arrtype="SCI"):
     prihdr['BITPIX']    = 8            # Array data type (no array in this HDU)
     prihdr['NAXIS']     = 0            # Number of array dimensions
     prihdr['EXTEND']    = True         # Denotes FIT extensions
-    prihdr['VISITID']   = '1'          # Full visit ID (placeholder positive integer)
+    prihdr['VISITID']   = '0000000000000000000' # Full visit ID (placeholder positive integer)
```
semaphoreP, line 132: if we just modify the input_dataset filenames, we don't need to worry about the filenames/headers having the exact same values

tests/test_badpixelmap.py
```diff
@@ -133,5 +147,15 @@ def test_packing_unpacking_uint64():
     # Check that only one bit is set
     assert np.sum(unpacked_bits[0, 0]) == 1, "Only one bit should be set"
 
+def test_output_filename_convention():
+    print("**Testing output filename naming conventions**")
+    badpixelmap, dark_frame, flat_frame = generate_badpixel_map()
+    badpixelmap.save(filedir=default_cal_dir)
+    pattern = r"^CGI_[A-Z0-9]{19}_\d{8}T\d{7}_BPM_CAL\.fits$"
+    matched_files = [fn for fn in os.listdir(default_cal_dir) if re.match(pattern, fn)]
+    assert matched_files, "No files found matching naming convention."
```
semaphoreP, line 156: please also check that the filename is based on the last filename in the input dataset (visit and utc time).

semaphoreP (commented):
My main comment is to implement the filename modification using the last file in the filename, rather than using the header keywords. The spirit of the filename convention was to modify the file string, rather than needing to check against header keywords (which technically should work, but let's not introduce another possible way for thigs 

We should also have a unit test check that we are using the last filename. So not just that the format is correct, but that the numbers in the visitid and utc time are the same as the last filename's.

corgidrp/data.py
juliamilton: Sure, but they should be equivalent

corgidrp/mocks.py
juliamilton: I think it is good to confirm that the headers match the filenames, no?

corgidrp/data.py
semaphoreP: Yes, but I'm more thinking I prefer we follow the same procedure, rather than be a combination of replacing filenames and reconstructing from the header. That just complicates the implementation.

tests/test_badpixelmap.py
juliamilton: Done!

corgidrp/mocks.py
semaphoreP: I just don't want to add unnecessary constraints. It's fine here though, we don't need to change this.

semaphoreP:
sorry, just added a merge conflict by merging in your other PR

corgidrp/data.py
juliamilton: Ok, I use the "format_ftimeutc" function anyway to confirm that the FTIMEUTC in the header matches the rounded version in the filename. With that function, I think either route of creating a new filename should be like 2 lines of code max. I think the benefit of reconstructing and not modifying an existing file name for now is that many of the functions are using mock input images to make the cals and are not assigning properly-formatted filenames to their mock input images, so that would cause some additional work to make everything correctly named.

tests/test_badpixelmap.py
semaphoreP: I would prefer we test the specification of this naming convention, which is that we are using the last visitid and UTC from the input data that made this Dark. We don't test that's true (e.g. the visitid and ftimeutc actually comes from the last input data).

semaphoreP: I think we should have an input dataset for this test that we start with.

corgidrp/data.py
semaphoreP: Yeah it's not about the lines of code, it's about having them use the same procedure. 

I did this with flat field and didn't run into much issue dealing with mock input images and making sure those are fine.

semaphoreP (approved):

PR #382, opened 2025-03-28
Nonlin filename

hsergi:
## Describe your changes

I added the current filename convention for non-linearity: _NLN_CAL to both the unit test and the e2e test.

https://github.com/roman-corgi/corgidrp/issues/356

- [x] I have linted my code

### DIFF
--- a/tests/test_nonlin_cal.py
+++ b/tests/test_nonlin_cal.py
@@ -203,9 +203,12 @@ def test_expected_results_nom_sub():
     assert np.equal(nonlin_out.data[norm_ind+1,0], norm_val)
 
     # Test filename follows convention (as of R3.0.2)
-    nonlin_out.save(filedir=corgidrp.default_cal_dir)
+    datadir = os.path.join(os.path.dirname(__file__), "simdata")
+    if not os.path.exists(datadir):
+        os.mkdir(datadir)
+    nonlin_out.save(filedir=datadir)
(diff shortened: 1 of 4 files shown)


tests/test_nonlin_cal.py
```diff
@@ -202,6 +202,27 @@ def test_expected_results_nom_sub():
     # check that norm_val is correct
     assert np.equal(nonlin_out.data[norm_ind+1,0], norm_val)
 
+    # Test filename follows convention (as of R3.0.2)
+    nonlin_out.save(filedir=corgidrp.default_cal_dir)
```
semaphoreP, line 206: this isn't the appropriate directory to save test files, because if somehow it doesn't get deleted (e.g., by this test crashing), then it can accidentally be used for processing real data. I would save them in the test folder itself: e.g., https://github.com/roman-corgi/corgidrp/blob/f842ee1f821760be2b8d299a19859856bc6585a0/tests/test_non_linearity_correction.py#L151.

semaphoreP (commented):
looks good. just one small comment on where to save your test file, to avoid cross-contamination with real data.

tests/test_nonlin_cal.py
hsergi: Sounds good. In CT, I start by deleting it. This is better.

semaphoreP (approved):

PR #387, opened 2025-03-29
correction to traditional master dark units

kjl0025:
## Describe your changes
The "traditional" master dark is the result of mean-combining a bunch of darks at the same exposure time and EM gain. 
 For this PR, I changed units for traditional master dark (which is not EM gain-divided) to "Detected EM Electrons" from "Detected Electrons", which are the default units used for the other master darks which are intended for the current pipeline (PC and synthesized from detector noise maps) since those two are EM gain-divided.  I also removed an unnecessary resetting of "OBSNUM" in the synthesized master dark creation.  

The Confluence header page for Dark will be fully consistent with the headers and units as they are in the code after this PR is accepted.

## Type of change

- [x] I have linted my code

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -65,6 +65,7 @@ tests/testcalib/*
 tests/e2e_tests/*/*.json
 tests/ops_output/*.json
 tests/test_data/L4_to_TDA_Inputs/*
+tests/e2e_tests/pc_sim_test_data/l2a/*.json
 
 # editor files
 .vscode/*
--- a/README.md
+++ b/README.md
@@ -146,7 +146,7 @@ End-to-end testing refers to processing data as one would when we get the real d
 To run the existing end-to-end tests, you need to have downloaded all the TVAC data to your computer. In a terminal, go to the base directory of the corgidrp repo and run the following command, substituting paths for paths on your computer as desired:
 
 ```
-pytest --which e2e --tvacdata_path /path/to/CGI_TVAC_Data --e2eoutput_path tests/e2e_tests/ tests/e2e_tests/
+pytest --which e2e --e2edata_path /path/to/CGI_TVAC_Data --e2eoutput_path tests/e2e_tests/ tests/e2e_tests/
 ```
 
 ### Linting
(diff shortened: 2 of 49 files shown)


kjl0025:
@nasavbailey :  I seem to remember making the traditional master dark without dividing by EM gain at the end for a reason.  If it is divided by EM gain, then this type of master dark would fit into the pipeline to L2b just fine, but without dividing by EM gain, it does not.  Maybe we decided on doing things this way because CTC required it to be this way?

nasavbailey:
@kjl0025 Hmm, I don't remember that discussion. I would have expected analog darks should have the same units whether they are synthetic or traditional. Perhaps it was a missed bug in the first release? CTC doesn't use DRP-produced traditional darks, except indirectly when they use a DRP-produced L2b product.  (You're just asking about analog darks, right? Darks for photon-counted & corrected data need to be photon-counted & corrected themselves, so EM gain should have been accounted for already there?)

kjl0025:
Okay, in that case, if there are no restrictions from CTC, I will just edit the DRP to include EM gain division in the creation of the traditional master dark and change header Wiki accordingly.  I'll push this PR again once I've made this change in the code.  And yes, of the 3 cases for master darks (PC, analog traditional, and analog synthesized), only the traditional had this lack of EM gain division.

mygouf:
> @kjl0025 Hmm, I don't remember that discussion. I would have expected analog darks should have the same units whether they are synthetic or traditional. Perhaps it was a missed bug in the first release? 

We didn't have time to check traditional darks in the previous release, only synthetic darks were tested so that could explain why we haven't caught this. I might have missed that discussion as well so I'm not sure either why it was decided to treat traditional darks differently. Seems like a good thing to update this, although it would be good to hear from people who were part of that discussion to understand if that change will impact other code features.

kjl0025:
The only people I thought were involved were you, Vanessa, and I.  I've searched my Slack and email and didn't find anything pertaining to leaving off EM gain division for traditional master darks.  I apologize if it was just my oversight.  I'm not sure.  In any case, I just pushed a change to include the EM gain division.

semaphoreP (approved):

Issue #393, opened 2025-04-02
Flux calibration: write the used SED filenames in the history header


JuergenSchreiber:
**Is your feature request related to a problem? Please describe.**
People want to know which SED (e.g. CALSPEC was used to calibrate the flux)

**Describe the solution you'd like**
The name of the used SED file(s) should be written in the HISTORY header

**Additional context**

JuergenSchreiber:
I guess the keywords FILEX could be used for this, too?

PR #396, opened 2025-04-04
add SED file names in History header

JuergenSchreiber:
## Describe your changes
write the applied SED filenames (mostly CALSPEC files) in the History header, when determining the FluxcalFactor, the apparent magnitude and the color correction.

## Type of change

#393

- [x] I have linted my code

### DIFF
--- a/corgidrp/fluxcal.py
+++ b/corgidrp/fluxcal.py
@@ -45,7 +45,7 @@ def get_calspec_file(star_name):
         star_name (str): 
     
     Returns:
-        str: file path
+        str: file path, fits file name
     """
     if star_name not in calspec_names:
         raise ValueError('{0} is not in list of anticipated standard stars {1}, please check naming'.format(star_name, calspec_names.keys()) )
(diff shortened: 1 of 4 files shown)


JuergenSchreiber:
merge with develop?

semaphoreP:
I would err towards merging with develop for now.

corgidrp/fluxcal.py
```diff
@@ -45,7 +45,7 @@ def get_calspec_file(star_name):
         star_name (str): 
     
     Returns:
-        str: file path
+        str: file path, fits file name
```
semaphoreP, line 48: each return should be on a separate line, e.g.:

```
str: file path
str: fits file name
```

Also, is the fits name just the filename (not including the path)?

semaphoreP (approved):
One minor docstring comment, otherwise looks good.

corgidrp/fluxcal.py
JuergenSchreiber: yes, exactly, the name from the calspec dictionary

PR #399, opened 2025-04-08
Merge PSF Centroid Computation in Spectroscopy

Ezarsh02:
## Describe your changes
Created computing PSF centroid function:
- Computes centroids for a grid of PSFs
- Takes PSF array and initial centroid guesses as input
- Returns calibrated centroid positions
- Handles FITS header management and file output
Testing framework:
- Created test_psf_centroid function to validate the function
- Tests PSF centroid computation accuracy
- Validates input/output file handling

## Type of change

- [x] I have linted my code


tests/test_spec.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Load FITS data and table (contains true values for testing)
+    with fits.open(file_path) as hdul:
+        psf_array = hdul[0].data
+        psf_table = Table(hdul[1].data)
+        print(f"FITS file loaded. PSF Array shape: {psf_array.shape}")
+        print(f"Metadata rows: {len(psf_table)}")
+
+    # Create Dataset object for traceability
+    dataset = Dataset([file_path])
```
semaphoreP, line 25: we should put each 2-D slice of the PSF array as an individual Image class into the dataset. See: https://github.com/roman-corgi/corgidrp/blob/spectroscopy/tests/test_dataset.py#L52

tests/test_spec.py
```diff
@@ -0,0 +1,91 @@
...
+   
+    calibration_path = os.path.join(output_dir, calibration.filename)
+    assert os.path.exists(calibration_path), "Calibration file not created"
+
+    # Test loading the calibration file
+    calibration_dataset = Dataset([calibration_path])
+    print(f"\nSuccessfully loaded the calibration FITS as a Dataset.")
+    print(f"Number of frames in dataset: {len(calibration_dataset)}")
+
+    with fits.open(calibration_path) as hdul:
```
semaphoreP, line 80: what is the purpose here? should we have some asserts?

semaphoreP (commented):
Nice work. Here are some comments to work on.

corgidrp/spec.py
```diff
@@ -0,0 +1,410 @@
...
+                                 }
+        output_path (str, optional): Path to write the resulting FITS file. Directory must exist.
+        verbose (bool): If True, prints fitted centroid values for each frame.
+        halfwidth (int): Half-width of the PSF fitting box. Default is 10.
+        halfheight (int): Half-height of the PSF fitting box. Default is 10.
+
+    Returns:
+        SpectroscopyCentroidPSF: Calibration object with fitted (x, y) centroids.
+    """
+    from corgidrp.data import Dataset
```
semaphoreP, line 356: very minor, but imports should be done at the start of the module, not inside of functions (unless we have a very good reason for it).

corgidrp/spec.py
```diff
@@ -0,0 +1,410 @@
...
+            print(f"Slice {idx}: x = {xfit:.3f}, y = {yfit:.3f}")
+
+    calibration = SpectroscopyCentroidPSF(
+        centroids,
+        pri_hdr=pri_hdr,
+        ext_hdr=ext_hdr,
+        input_dataset=dataset
+    )
+
+    if output_path:
```
semaphoreP, line 402: we generally don't have save functions inside of step functions. There's a separate save function we use for all saving.

corgidrp/spec.py
```diff
@@ -0,0 +1,410 @@
...
+        calibration.save(output_path)
+
+        # Set EXTNAME keyword before saving
+        calibration.ext_hdr["EXTNAME"] = "CENTROIDS"
```
semaphoreP, line 406: We should set EXTNAME on creation of SpectroscopyCentroidPSF (i.e., in its `__init__` function), rather than on save

semaphoreP (commented):
A few minor comments. we're almost there !

PR #395, opened 2025-04-03
CT e2e

hsergi:
## Describe your changes
I have implemented an e2e for the CT calibration file as agreed in the past weeks. Note: There's a separate e2e test for the CT map (CT map uses a CT cal file to get its results)

## Type of change
- New feature (non-breaking change which adds functionality)
## Reference any relevant issues (don't forget the #)
#372 

## Checklist before requesting a review
- [x] I have linted my code
- [x] I have verified that all unit tests pass in a clean environment and added new unit tests, as needed
- [x] I have verified that all docstrings are properly formatted and added new documentation, as needed
- [ ] I have filled out the Unit Test Definition Table on confluence, as needed

Issue #373, opened 2025-03-26
Create e2e test for CT map


hsergi:
**Is your feature request related to a problem? Please describe.**
We need an e2e test for the CT map, starting with mock data and using the walker

**Additional context**

See #346

PR #400, opened 2025-04-09
CT map e2e

hsergi:
## Describe your changes
I have implemented an e2e for the CT map alone. The CT cal file e2e test is dealt with in #395.
 
## Type of change

CT map e2e issue: #373 

- [x] I have linted my code


corgidrp/recipe_templates/l3_to_corethroughput_map.json
```diff
@@ -0,0 +1,24 @@
+{
+    "name" : "l3_to_corethroughput_map",
```
semaphoreP, line 2: this should be l2b_to_corethroughput_map. The steps are fine, we are just starting from L2b (divide by exptime is a l2b_to_l3 step).

corgidrp/recipe_templates/l3_to_corethroughput_map.json
```diff
@@ -0,0 +1,24 @@
...
+            "name": "divide_by_exptime"
+        },
+        {
+            "name" : "CreateCTMap",
+            "calibs": {
+                "FpamFsamCal": "AUTOMATIC",
+                "CoreThroughputCalibration" : "AUTOMATIC"
+            },
+              "keywords" : {
+                "save" : true
```
semaphoreP, line 20: In general, all of the save steps are being done by the walker, not by the step function, because only the walker currently knows the outputdir (we don't want to save to the default_cal_dir, except for a few default calibrations that don't depend on real data).

This is a bigger change, but can we make a CTMap class in data.py, have CreateCTMap output that, and use the "save" step in the pipeline to save the CTMap rather than it being a bool in the step function.

semaphoreP (commented):
Hi Sergi, I just have one main comment that to align with corgidrp data saving philosophy, we should be doing the save function as an independent step (just like all other recipes). As such, the CTMap needs to be a data class so we can save it.

corgidrp/recipe_templates/l3_to_corethroughput_map.json
hsergi: It was agreed at UCSB with @maxwellmb that CTmap would be an exception due to the scope and use of the CT map itself. That is, the CT map is not a calibration product -as can be seen from Confluence.

maxwellmb: Yes, this is 100% my fault in the way that I described this. I was thinking originally that this would be more like a TDA-like product, but that doesn't really work with a recipe like this. So I think Jason is correct here about the infrastructure of the DRP and that in the end it would make sense to make a CTMap class. It doesn't have to be anything fancy, and should be relatively quick to model off the existing classes. My apologies for leading us in the wrong direction.

hsergi: Should I include in this PR the changes in the ct_map UT? Or in a separate PR?

hsergi: @semaphoreP It can be reviewed again. UT changed. I changed some other suggestions that were not high priority (syntax for the name of the function, folder to use with UT different to default's corgidrp folder)

corgidrp/recipe_templates/l3_to_corethroughput_map.json
semaphoreP: @hsergi can you change the name of this recipe to l2a_*? Last comment from me.

hsergi: I changed it to "l2a_to_corethroughput_map.json"

semaphoreP (approved):

PR #406, opened 2025-04-17
Update mock headers to match what is in the documentation 

juliamilton:
## Describe your changes
Some changes were made asynchronously to the header documentation pages of L1-L4 data products. This PR updates the mock header generation to match what is in the documentation pages for this release. I am requesting that we merge these changes into this current release so that the code matches the documentation, and also so that we can generate example mock data products for each of the documentation pages that match the documented header listing.

To prevent asynchronous changing of headers in the future, we will introduce a header change request process for review soon. This PR is to correct out-of-sync headers that currently exist and is not the procedure we'd like to use going forward.

- [x] I have linted my code


semaphoreP:
@juliamilton have you ran the e2e tests? I'm getting a lot of test failures from running the e2e tests. I can send you the error log if it's useful.

juliamilton:
> @juliamilton have you ran the e2e tests? I'm getting a lot of test failures from running the e2e tests. I can send you the error log if it's useful.

Most likely I just need to update the function calls, but would you mind sending me the error log? Some of the e2e tests just don't work locally for me (ongoing issue, unrelated to this PR).

semaphoreP:
See attached log here: [pytest_log.txt](https://github.com/user-attachments/files/19893942/pytest_log.txt)

juliamilton:
Thanks, I updated all the function calls in those files had failed in the log. My e2e tests fail or assert locally on l1_to_l2a_e2e.py and l1_to_l2b_e2e.py even on the unmodified main branch though. Also, the trap_pump_cal_e2e.py is now failing because the array sizes of the mock calibration fits files do not match what is expected in the metadata reference files, although I want to confirm that the metadata reference yamls are correct for that one.

semaphoreP:
The trap pump test works for me FYI.

maxwellmb:
UT or e2e? I suspect there might be a threshold that is too small, such that due to random variations it sometimes passes sometimes fails.

juliamilton:
> UT or e2e? I suspect there might be a threshold that is too small, such that due to random variations it sometimes passes sometimes fails.

e2e, I have never gotten it to work locally. I get "TPumpAnException('metadata file inconsistent with frame')". The other 2 e2e tests (l1_to_l2a_e2e.py and l1_to_l2b_e2e.py) do assert due to failing their difference threshold checks.

semaphoreP:
Do you need to clean up your ~/.corgidrp folder and your test directory and re-run? Those 3 tests all work for me on your branch.

semaphoreP:
@juliamilton I updated the header doc pages with how the headers of the e2e products changed with your new changes. Can you check these look ok with you, and if there's anyone else you should check with on this?

juliamilton:
Also, it would be good to record somewhere either in a header, in the recipe, and/ or in the confluence page which level data product (L1, L2b, L3) is used to make a given calibration product.

semaphoreP:
@juliamilton these sound like tasks for a different PR from yours. My original intent of the question is whether these header changes you've made impact any customers. The doc pages I uploaded just indicates what you've changed in your PR>

PR #407, opened 2025-04-18
clean out output files in trad dark e2e

kjl0025:
## Describe your changes
Made the traditional dark e2e tests remove any files from the output folder before and after.  This prevents the code from finding any previous *_DRK_CAL.fits file in the output folder.

## Type of change

- [ ] I have linted my code

### DIFF
--- a/tests/e2e_tests/trad_dark_e2e.py
+++ b/tests/e2e_tests/trad_dark_e2e.py
@@ -72,6 +72,10 @@ def test_trad_dark(e2edata_path, e2eoutput_path):
     if not os.path.exists(build_trad_dark_outputdir):
         os.mkdir(build_trad_dark_outputdir)
 
+    # remove any files in the output directory that may have been there previously
+    for f in os.listdir(build_trad_dark_outputdir):
+        os.remove(os.path.join(build_trad_dark_outputdir, f))
+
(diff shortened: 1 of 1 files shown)


tests/e2e_tests/trad_dark_e2e.py
```diff
@@ -237,6 +241,9 @@ def test_trad_dark(e2edata_path, e2eoutput_path):
     # remove from caldb
     trad_dark = data.Dark(generated_trad_dark_file.replace("_L1_", "_L2a_", 1))
     this_caldb.remove_entry(trad_dark)
+    # clean out output files that were made
+    for f in os.listdir(build_trad_dark_outputdir):
```
semaphoreP, line 245: Do we also need to be cleaning out the directory afterwards? Seems like it may be useful to keep the files for manual inspection, and the cleaning of files before the test solves the bug?

semaphoreP (commented):
Do we need to be removing the files after the test as well? Seems like overkill.

tests/e2e_tests/trad_dark_e2e.py
kjl0025: True, sounds good.  Just updated accordingly.

semaphoreP (approved):

PR #417, opened 2025-04-29
PC e2e VAP

### LINKED ISSUE #344: Update trap pumping header keywords to fit within 8 characters
Current trap pumping-only headers implemented in the DRP, TPSCHEME1-TPSCHEME4, are > 8 characters. Update the code (and the wiki L1 header page) to make them TPSCHEM1-TPSCHEM4. This is already implemented on the SSC side and does not require a change request to them.

Tagging @kjl0025 for awareness.

kjl0025:
## Describe your changes
I added extra tests into the PC e2e test for VAP purposes.  Currently, I have the recipes run for PC to exercise the walker for the e2e, but I also run separately from L2a each step needed to get to L2b b/c the output from each of those steps is needed for a VAP test for PC.  We could add a `--VAP` flag to the `pytest` to avoid duplicating runtime for steps if it comes to that.

I also edited the traditional dark e2e and noise map e2e to ensure all expected headers are in the output calibration file (to get a bona fide sample file on the Confluence pages for these calibration products).

I also fixed a few headers that had typos in them to be consistent with Confluence:  B_0_* --> B_O_* and TPSCHEME* --> TPSCHEM*.

Closes #344 .

## Type of change

- [x] I have linted my code


kjl0025:
Also, I have this set to merge into main instead of dev.  @mygouf , the VAP testers need this on main, right?

semaphoreP:
It seems like to me you've: 1) appended to the current e2e test another reduction from scratch, and 2) added a bunch of tests that check the output of the second set of reductions. 

Is there any reason you can't structure it as: a new e2e test function that does all the checks of the output in a single test? I don't really like all these global variables floating around.

mygouf:
> Also, I have this set to merge into main instead of dev. @mygouf , the VAP testers need this on main, right?

Yes, correct. Thanks Kevin!

kjl0025:
The VAP testers want to see what all failed in the output.  If multiple `assert` statements are in a single test function, the function stops once the first `assert` fails, and any other failures from `assert` statements after that don't show up in the output.  

Good point about the global vars, though.  I should `del` them at the end.  I'll push that change.

ajeldorado:
We also want to see some things that passed in the output, but assert statements don't print anything for successful tests. Tim found that some of the functionality print statements are put into a really long header string that is parseable, though, so for some things we could go through the effort of parsing that.

semaphoreP:
Have you tried running `pytest -sv`? The `-sv` flag should print out all print statements to the console.

kjl0025:
I just reopened it to pull in the non-VAP changes, to the dev branch.  I can remove all the VAP test things from here and then tag you on here when I finish if that sounds good?

semaphoreP:
Sounds good. Please also switch the target branch to `develop`.

kjl0025:
@semaphoreP , this should be good to go now.

maxwellmb:
@kjl0025 I think you're due to merge in main (and resolve a conflict)

semaphoreP:
Looks like someone changed it to develop.

The changing of header keywords may hold up the PR. We're trying to establish a new system to handle header changes.

kjl0025:
Yeah, I had switched it to merge to develop, but actually, I agree that these changes here should go into main b/c they fix some headers currently in main that deviate from what is in the Confluence pages (B_0_* --> B_O_* and TPSCHEME* --> TPSCHEM*).

semaphoreP:
We're been asked to freeze changes to main for VAP testing, and, again, headers process is being re-evaluated, so let's wait on this for a bit.

hsergi:
This is also needed for VAP testing, so main is the place. I just had an exchange with Eric:

https://github.jpl.nasa.gov/cgi-fsw/cgi-sit/blob/3bdabcb7aaf8ef3782720ad8e96c8564815129da/13-tdd/vi_tdd_06_ttr5_on_sky_nfov/run_vi_tdd_06_case_01.py#L36

There may be other changes etc

ecady-jpl:
We'll try to work around it for VAP testing for now.

ecady-jpl:
For the record, we found a workaround by manually building the bad frames for the VAP test with other corgidrp hooks, so no further action is needed on this branch to enable VAP testing at this time.

semaphoreP:
@kjl0025 I had to modify the astrometry e2e test (which uses the dark TVAC data) because of the header changes, but it should be ok now. 

@mygouf the headers for the DRP Dark file has changed slightly (it seems like to better reflect TVAC headers). Can you check if these header changes are fine with you? https://github.com/roman-corgi/corgidrp/pull/417/files#diff-4536e15e7e600ad56ec07b89dcf4ed8a7b930be12ee66a78dae2697e489a0ac9

docs/source/data_formats/dark.rst
```diff
@@ -51,8 +51,66 @@ Primary Header (HDU 0)
 +------------+------------+--------------------------------+----------------------------------------------------+
```
juliamilton, line 51: OBSID, BUILD, OBSTYPE keywords should not be included in primary header (no longer being passed in by L1s)

semaphoreP:
@kjl0025 can you edit the e2e input data degeneration to remove these?

kjl0025:
Should be good to go now.  I corrected the reference .rst file and updated the e2e to get the headers correct in the saved output file (from which I assume you made the reference .rst file?).

semaphoreP:
@kjl0025 Could we do this header modification on the input L1 files, rather than the output Dark? We want to get the right answer with the right input data, rather than fudging the output.

kjl0025:
Sure, I can change when in the e2e I delete obsolete headers.  Since the .rst reference files are made from the calibration products made by the e2es, other reference .rst files have the same obsolete headers present.  I can modify their e2es and .rst files as well.

semaphoreP:
@juliamilton, are your comments addressed with these changes?

Issue #293, opened 2025-01-30
There should be the possibility to input a single float data value and error value in the Image class


JuergenSchreiber:
**Is your feature request related to a problem? Please describe.**
We have calibration classes like KGain and FluxcalFactor derived from the Image class, where only 1 float value for the factor and for its error is stored. Up to now this is input using artificial generated 2/3D arrays first.

**Describe the solution you'd like**
Look into the possibility to input single float values to initialize the Image and/or the derived calibration classes.

**Additional context**

PR #423, opened 2025-05-05
Single dim

JuergenSchreiber:
## Describe your changes
make it possible to input single float values to initialize the Image and/or the derived calibration classes like FluxcalFactor or KGain for data and error.
Input float value has to be converted to a 1d array within the initialization.
## Type of change

#293

- [x] I have linted my code

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -344,22 +344,27 @@ class Image():
                 raise ValueError("Missing primary and/or extension headers, because you passed in raw data")
             self.pri_hdr = pri_hdr
             self.ext_hdr = ext_hdr
-            self.data = data_or_filepath
             self.filedir = "."
             self.filename = ""
-
-            # self.hdu_names = [hdu.name for hdu in self.hdu_list]
-
-            if err is not None:
-                if np.shape(self.data) != np.shape(err)[-self.data.ndim:]:
-                    raise ValueError("The shape of err is {0} while we are expecting shape {1}".format(err.shape[-self.data.ndim:], self.data.shape))
-                #we want to have a 3 dim error array
-                if err.ndim == self.data.ndim + 1:
-                    self.err = err
+            if isinstance(data_or_filepath, float):
+                self.data = np.array([data_or_filepath])
+                if err is not None and isinstance(err, float):
+                    self.err = np.array([err])
                 else:
-                    self.err = err.reshape((1,)+err.shape)
+                    self.err = np.array([0])
             else:
-                self.err = np.zeros((1,)+self.data.shape)
+                self.data = data_or_filepath
+
+                if err is not None:
+                    if np.shape(self.data) != np.shape(err)[-self.data.ndim:]:
+                        raise ValueError("The shape of err is {0} while we are expecting shape {1}".format(err.shape[-self.data.ndim:], self.data.shape))
+                    #we want to have a 3 dim error array
+                    if err.ndim == self.data.ndim + 1:
+                        self.err = err
+                    else:
+                        self.err = err.reshape((1,)+err.shape)
+                else:
+                    self.err = np.zeros((1,)+self.data.shape)
 
             if dq is not None:
                 if np.shape(self.data) != np.shape(dq):
@@ -847,11 +852,11 @@ class KGain(Image):
         if 'RN_ERR' not in self.ext_hdr:
             self.ext_hdr['RN_ERR'] = ''
         # File format checks
-        if self.data.shape != (1,1):
+        if self.data.shape != (1,):
             raise ValueError('The KGain calibration data should be just one float value')
 
-        self._kgain = self.data[0,0] 
-        self._kgain_error = self.err[0,0]
+        self._kgain = self.data[0] 
+        self._kgain_error = self.err[0]
         
         if isinstance(data_or_filepath, str):
             # a filepath is passed in
@@ -1364,7 +1369,7 @@ class FluxcalFactor(Image):
         super().__init__(data_or_filepath, err=err, pri_hdr=pri_hdr, ext_hdr=ext_hdr, err_hdr=err_hdr)
         # if filepath passed in, just load in from disk as usual
         # File format checks
-        if self.data.shape != (1,1):
+        if self.data.shape != (1,):
             raise ValueError('The FluxcalFactor calibration data should be just one float value')
         
         #TBC
@@ -1398,8 +1403,8 @@ class FluxcalFactor(Image):
             self.ext_hdr['DRPCTIME'] =  time.Time.now().isot
             
         # make some attributes to be easier to use
-        self.fluxcal_fac = self.data[0,0]
-        self.fluxcal_err =  self.err[0,0,0]
+        self.fluxcal_fac = self.data[0]
+        self.fluxcal_err =  self.err[0]
 
         # if this is a new FluxcalFactors file, we need to bookkeep it in the header
         # b/c of logic in the super.__init__, we just need to check this to see if it is a new FluxcalFactors file
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -290,10 +290,12 @@ class Image():
 
                 # we assume that if the err and dq array is given as parameter they supersede eventual err and dq extensions
                 if err is not None:
-                    if np.shape(self.data) != np.shape(err)[-self.data.ndim:]:
+                    if isinstance(err, float):
+                        self.err = np.array([err])
+                    elif np.shape(self.data) != np.shape(err)[-self.data.ndim:]:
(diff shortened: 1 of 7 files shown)


corgidrp/data.py
```diff
@@ -344,22 +346,27 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq
...
-
-            # self.hdu_names = [hdu.name for hdu in self.hdu_list]
-
-            if err is not None:
-                if np.shape(self.data) != np.shape(err)[-self.data.ndim:]:
-                    raise ValueError("The shape of err is {0} while we are expecting shape {1}".format(err.shape[-self.data.ndim:], self.data.shape))
-                #we want to have a 3 dim error array
-                if err.ndim == self.data.ndim + 1:
-                    self.err = err
+            if isinstance(data_or_filepath, float):
```
semaphoreP, line 351: I generally try to avoid multiple levels of if statements when possible, for maintaining readability.

Can we do the check for float first, and then continue as is. (i.e., starting around line 345 do something like:

```
if isinstance(data_or_filepath, float):
    data_or_filepath = np.array([data_or_filepath])
     if err is not None and isinstance(err, float):
         err = np.array([err])
     else:
          err = np.array([0])
```

After this, we can keep the same logic as before (which already has two levels of if statements).

corgidrp/data.py
```diff
@@ -847,11 +854,11 @@ def __init__(self, data_or_filepath, err = None, ptc = None, pri_hdr=None, ext_h
         if 'RN_ERR' not in self.ext_hdr:
             self.ext_hdr['RN_ERR'] = ''
         # File format checks
-        if self.data.shape != (1,1):
```
semaphoreP, line 850: We may need to check if this will affect anyone other than the DRP (e.g., does GITL use this?), as we don't want to break compatibility.

semaphoreP (changes requested):
A few comments. 
 
 In addition, the e2e tests need to be updated for Kgain and FluxcalFactor, since it appears this breaks backwards compatability? (Maybe we could not break backwards compatability?)

corgidrp/data.py
```diff
@@ -290,10 +290,12 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq
 
                 # we assume that if the err and dq array is given as parameter they supersede eventual err and dq extensions
                 if err is not None:
-                    if np.shape(self.data) != np.shape(err)[-self.data.ndim:]:
+                    if isinstance(err, float):
+                        self.err = np.array([err])
```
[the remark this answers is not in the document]
JuergenSchreiber: right

JuergenSchreiber:
> A few comments.
> 
> In addition, the e2e tests need to be updated for Kgain and FluxcalFactor, since it appears this breaks backwards compatability? (Maybe we could not break backwards compatability?)

I adjusted calibrate_kgain and calibrate_fluxcal, so the e2e tests work for me.

semaphoreP:
Tagging @mygouf @juliamilton @ajeldorado @ecady-jpl: I'd like to get your input on this PR's proposed change to change the Kgain and FluxCalFactor data format from a shape of (1,1) to a shape of (1,). (This is for the data stored in the Image HDU of the output FITS file). Particularly, for Kgain, does this effect FSW?

ecady-jpl:
There's no FSW, impact, no, if it's just that scalar value.  That value has to get moved into a different command outside of Python to go to the spacecraft, so whether it's a 1D or 2D ndarray makes no difference there.

semaphoreP:
> > A few comments.
> > In addition, the e2e tests need to be updated for Kgain and FluxcalFactor, since it appears this breaks backwards compatability? (Maybe we could not break backwards compatability?)
> 
> I adjusted calibrate_kgain and calibrate_fluxcal, so the e2e tests work for me.

I just tried again and kgain_e2e.py and nonlin_e2e.py still fail for me, even though they work on develop branch for me? 

Can you merge `develop` into this branch so I can run the e2e tests on github actions to double check?

corgidrp/data.py
JuergenSchreiber: > I'm not sure I follow. Aren't we in the case data is passed in directly (not loading an existing FITS file), so err and dq do need to be passed in?

I was talking about the other case, so if loading an FITS file, we do not need to check about the err and dq array input.

semaphoreP: Let's not worry about that for this PR. I do agree that it's not really necessary.

JuergenSchreiber:
> Ok, right PR this time. @JuergenSchreiber, please check and fix the e2e tests: https://github.com/roman-corgi/corgidrp/actions/runs/15166946113

Ah, now I see what you mean....

semaphoreP:
@JuergenSchreiber the e2e tests are still failing for me..

JuergenSchreiber:
Most e2e tests work now but I get some errors I think that come from another problem?
test_trap_pump_cal - cal.tpumpanalysis.tpump_final.TPumpAnException: metadata file inconsistent ..
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_l2banalog_dataformat_e2e - AssertionError: assert 9 == 0
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_l2bpc_dataformat_e2e - AssertionError: assert 18 == 0
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_astrom_dataformat_e2e - AssertionError: assert 10 == 0
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_bpmap_dataformat_e2e - IndexError: list index out of range
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_flat_dataformat_e2e - IndexError: list index out of range
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_noisemaps_dataformat_e2e - AssertionError: assert 10 == 0
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_dark_dataformat_e2e - AssertionError: assert 10 == 0
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_tpump_dataformat_e2e - IndexError: list index out of range

semaphoreP:
Can you try deleting the `trap_pump_cal_output` dir and trying to rerun the trap_pump_cal test again? 

For the non-trap-pump related ones, I think these tests are failing because the headers changed due to the datatype of kgain/kgain_er changing (some switched between int and float). If you copy the .rst files from `tests/e2e_tests/data_format_docs/` t `docs/source/data_formats/` and commit the new data_format_docs, that will make the tests pass and show us the changes that have been made to the data_format_docs as part of the git commit history.

Let me know if that makes sense.

JuergenSchreiber:
ok, thanks, now its much better, but still 2 failures
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_bpmap_dataformat_e2e - IndexError: list index out of range
FAILED tests/e2e_tests/zz_dataformat_e2e.py::test_flat_dataformat_e2e - IndexError: list index out of range

semaphoreP:
I checked on Github pipelines, and all the e2e tests pass. Horray! 

@juliamilton Some data format changes that affect confluence. Let us know if these sound reasonable to you:
 * Kgain and absflux data are now shape (1,) instead of (1,1) previously. We already checked with Eric above that this shouldn't affect customers, but will change the confluence documentation for v3
 * KGAINPAR and KGAIN_ER could be int or float sometimes. Is that ok? do we have to be strict between float and int?

corgidrp/data.py
```diff
@@ -340,27 +344,31 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq
...
+            else:
+                self.data = data_or_filepath
+                if err is not None:
+                    if np.shape(self.data) != np.shape(err)[-self.data.ndim:]:
+                        raise ValueError("The shape of err is {0} while we are expecting shape {1}".format(err.shape[-self.data.ndim:], self.data.shape))
+                    #we want to have a 3 dim error array
+                    if err.ndim == self.data.ndim + 1:
+                        self.err = err
+                    else:
+                        self.err = err.reshape((1,)+err.shape)
```
semaphoreP, line 362: I think these lines of code (362, 364) produce an error of default datatype int, instead of float. Please change this is the KGAIN_ER should output only floats.

semaphoreP:
@JuergenSchreiber we should fix the KGAIN_ER being int in some cases.  I think I figured out why above. Otherwise, looks good.

JuergenSchreiber:
> @JuergenSchreiber we should fix the KGAIN_ER being int in some cases. I think I figured out why above. Otherwise, looks good.

Ok, in Image I now tried to allow only inputs that are float or have a len attribute. If float data input I only allow a float error input. What do you think?

PR #439, opened 2025-05-13
apply the DQ values added from detect_cosmic_rays to nonlin and kgain calibrations

JuergenSchreiber:
## Describe your changes
For kgain and nonlinearity calibration functions masked pixels from cosmic ray detection are set to NaN, most mean operations are changed to nanmean.

## Type of change

#202 

- [x] I have linted my code

### DIFF
--- a/corgidrp/calibrate_kgain.py
+++ b/corgidrp/calibrate_kgain.py
@@ -146,7 +146,7 @@ def diff2std(diff_frame, detector_regions=None):
         
     selected_area = slice_section(diff_frame, 'SCI', 'prescan_reliable',
         detector_regions)
-    std_value = np.std(selected_area.reshape(-1), ddof=1)
+    std_value = np.nanstd(selected_area.reshape(-1), ddof=1)
     # dividing by sqrt(2) since we want std of one frame
     return std_value / np.sqrt(2)
     
@@ -261,8 +261,8 @@ def sigma_clip(data, sigma=2.5, max_iters=6):
     clipped_data = data.copy()
       
     for i in range(max_iters):
-        mean = np.mean(clipped_data)
-        std = np.std(clipped_data)
+        mean = np.nanmean(clipped_data)
+        std = np.nanstd(clipped_data)
         mask = np.abs(clipped_data - mean) > sigma * std
         if not np.any(mask):
             break
@@ -524,11 +524,11 @@ def calibrate_kgain(dataset_kgain,
             in range(len(frames_diff))]
         
         # split each frame up into bins, take std and mean of each region
-        mean_frames_mean_curr0 = np.mean(frames, axis=0)
-        mean_frames_mean_curr = np.mean(mean_frames_mean_curr0[rowroi,colroi])
+        mean_frames_mean_curr0 = np.nanmean(frames, axis=0)
+        mean_frames_mean_curr = np.nanmean(mean_frames_mean_curr0[rowroi,colroi])
         
         # Calculate the means
-        mean_good_mean_frame_roi = np.mean(good_mean_frame[rowroi,colroi])
+        mean_good_mean_frame_roi = np.nanmean(good_mean_frame[rowroi,colroi])
     
         # Compute the scaling factor
         scaling_factor = mean_frames_mean_curr / mean_good_mean_frame_roi
@@ -588,11 +588,11 @@ def calibrate_kgain(dataset_kgain,
     
         # Compute statistics if there are data points within the bin
         if current_binned_averages.size > 0:
-            binned_averages_compiled.append(np.mean(current_binned_averages))
-            binned_averages_error.append(np.std(current_binned_averages, ddof=1) / np.sqrt(current_binned_averages.size))
-            binned_shot_deviations_compiled.append(np.mean(current_binned_deviations))
-            binned_deviations_error.append(np.std(current_binned_deviations, ddof=1) / np.sqrt(current_binned_averages.size))
-            binned_total_deviations.append(np.mean(current_binned_total_deviations))
+            binned_averages_compiled.append(np.nanmean(current_binned_averages))
+            binned_averages_error.append(np.nanstd(current_binned_averages, ddof=1) / np.sqrt(current_binned_averages.size))
+            binned_shot_deviations_compiled.append(np.nanmean(current_binned_deviations))
+            binned_deviations_error.append(np.nanstd(current_binned_deviations, ddof=1) / np.sqrt(current_binned_averages.size))
+            binned_total_deviations.append(np.nanmean(current_binned_total_deviations))
         else:
             # Append NaN or some other placeholder if no data points in the bin
             binned_averages_compiled.append(np.nan)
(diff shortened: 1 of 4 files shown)


kjl0025:
The changes here don't break anything, but there shouldn't be any NaN pixels in the data itself, so I don't think this enhances anything either.  Per Issue #202 , the code for calibrating k gain and nonlin ideally should be changed so that the frames are combined in a way that takes into account masked pixels as marked by the DQ, and the functions in combine.py do this.  However, combine.py inherently assumes that statistical error over the set of frames in a dataset is included in each frame's ERR (along with any inherent error present per frame), and the statistical error is included in each frame's ERR if `add_photon_noise()` has been already applied to the dataset.  `add_photon_noise()` (or your improvement to it in PR #289 ) adds in error that is better than standard error from averaging frames only in the case of a dataset with a sufficiently small number of frames, so I plan to add in an option to that function to take the standard error over the dataset, divide it into N pieces, and add that to each frame's ERR.  Adding this would also make the ERR that goes into the flat field calibration more accurate as it currently uses the `add_photon_noise()` function in its recipe.  I don't mind working on implementing all this on a branch of my own.  I'd like to wait until your PR #289 is pulled in, though.  Does that all sound good to you, @JuergenSchreiber ?

JuergenSchreiber:
> The changes here don't break anything, but there shouldn't be any NaN pixels in the data itself, so I don't think this enhances anything either. Per Issue #202 , the code for calibrating k gain and nonlin ideally should be changed so that the frames are combined in a way that takes into account masked pixels as marked by the DQ, and the functions in combine.py do this. However, combine.py inherently assumes that statistical error over the set of frames in a dataset is included in each frame's ERR (along with any inherent error present per frame), and the statistical error is included in each frame's ERR if `add_photon_noise()` has been already applied to the dataset. `add_photon_noise()` (or your improvement to it in PR #289 ) adds in error that is better than standard error from averaging frames only in the case of a dataset with a sufficiently small number of frames, so I plan to add in an option to that function to take the standard error over the dataset, divide it into N pieces, and add that to each frame's ERR. Adding this would also make the ERR that goes into the flat field calibration more accurate as it currently uses the `add_photon_noise()` function in its recipe. I don't mind working on implementing all this on a branch of my own. I'd like to wait until your PR #289 is pulled in, though. Does that all sound good to you, @JuergenSchreiber ?

@kjl0025  Hi Kevin, I only set NaNs where the mask is set, the NaNs should have been  gone after the nanmean/nanmedian/nanstd operation, that is the same method as in combine.py. If you do not like this we also could work with masked arrays applying image.get_masked_data(). I am not 100% sure whether I understand what you mean, but I am fine with assigning this task to you especially for better error propagation.

semaphoreP:
This PR modifies the kgain/nonlin code to set pixels as NaNs when they have non-zero DQ values, correct? (e.g., line 841 in calibrate_kgain).

kjl0025:
Ah, I apologize!  I saw those lines and didn't realize that actually propagated to the main calibration code, but they do.  And the way that ERR is calculated for the k gain calculation is compatible with this (and the nonlin ERR has yet to be implemented).  So I am fine with all this.  

My thing about add_photon_noise() is indeed a separate issue that I will dig into later (after Juergen's PR #289  is pulled in).

kjl0025 (approved):

semaphoreP (changes requested):
The e2e tests are failing for me. I noticed at least the following need to be fixed:

Line 570 of calibrate_kgain (min -> nanmin, max -> nanmax).

Are you able to run the e2e tests?

JuergenSchreiber:
> The e2e tests are failing for me. I noticed at least the following need to be fixed:
> 
> Line 570 of calibrate_kgain (min -> nanmin, max -> nanmax).
> 
> Are you able to run the e2e tests?

yes, for me the e2e test runs, but I change the min to nanmin and max to nanmax to be on the save side

semaphoreP:
@JuergenSchreiber I just ran the e2e tests on Github actions and got several test failures. Can you look into these? https://github.com/roman-corgi/corgidrp/actions/runs/15166946113/job/42647105657

semaphoreP:
Ah my bad. Can you merge develop into this branch, so I can run the e2e tests on Github actions?

semaphoreP:
E2e tests are still failing here: https://github.com/roman-corgi/corgidrp/actions/runs/15194003502/job/42733553693 (ignore the l1_to_l2b failing.. I think that's just collateral damage). Neither kgain_e2e.py nor nonlin_e2e.py are passing though (which is what I see locally on my machine).

JuergenSchreiber:
> E2e tests are still failing here: https://github.com/roman-corgi/corgidrp/actions/runs/15194003502/job/42733553693 (ignore the l1_to_l2b failing.. I think that's just collateral damage). Neither kgain_e2e.py nor nonlin_e2e.py are passing though (which is what I see locally on my machine).

Yes, there are assertion errors now. I think for the kgain_e2e I can explain it: The Tvac  kgain is about 8.8 while the corgi pipeline kgain is now about 8.2, and we expect 0 difference. We consider now the cosmics but only in the corgi walker pipeline, but for tvac we only run calibrate_kgain. One solution would be to relax the strong constrains and allow the kgain_err of about 1 as difference. 
I guess for nonlin_e2e we have a similar problem.

semaphoreP:
Is it a lot of work to run detect_cosmic_rays and mask the data in TVAC data processing for each e2e test?

semaphoreP:
Do we just want the e2e test to have `apply_dq` to be False, or are we ok with the recipe file by default setting this keyword to false? Note that changing it in the default recipe, effectively overrides the default behavior of the step function.

JuergenSchreiber:
In my opinion, it should only be set to false in the e2etest recipe. The default in the operation pipeline should be that the cosmic rays are considered, what do you think @kjl0025 ?

kjl0025:
Yes, I agree:  apply_dq should be True by default (as not correcting for cosmic rays was a deficiency in the TVAC code which was allowed for time's sake since it wasn't expected to make a huge impact), and the default recipe should have apply_dq as True to have the best determination of k gain and nonlinearity.  We could have separate recipes for the e2es called something like l1_to_kgain_e2e.json and l1_to_l2a_nonlin_e2e.json which set apply_dq to False.

semaphoreP:
You can do what's done in the flat field e2e tests to modify the recipes for the e2e tests specifically, for changes specific to the e2e tests that are different from the default recipe. 

```
    recipe = walker.autogen_recipe(l1_flatfield_filelist, flat_outputdir)
     ### Modify they keywords of some of the steps
    for step in recipe['steps']:
        # if step['name'] in ["desmear", "cti_correction"]:
        #     step['skip'] = True
        if step['name'] == "create_onsky_flatfield":
            step['keywords']['n_pix'] = 165 # full shaped pupil FOV
    walker.run_recipe(recipe, save_recipe_file=True)
```

JuergenSchreiber:
ok, thanks, I hope now the e2etests will run...

JuergenSchreiber:
> @JuergenSchreiber the new modifications to the e2e tests don't work for me. do they work for you? I get the following error
> 
> ```
>         recipe = walker.autogen_recipe(nonlin_l1_list, e2eoutput_path)
>          ### Modify they keywords of some of the steps
> >       for step in recipe['steps']:
> E       TypeError: list indices must be integers or slices, not str
> 
> tests/e2e_tests/nonlin_e2e.py:117: TypeError
> ```

Now I am lost, aren't those lines exactly what you proposed from the flat e2etest?

semaphoreP:
I though so? But this may be a case where `walker.autogen_recipe()` returns a list of recipes rather than a single recipe. You may alternatively need to specify exactly what template you want to start with.

JuergenSchreiber:
Ok, the guess template is a list of kgain and nonlin recipe template, (do we really want to have this?), to get a step["keywords"], I had to add a default keyword to nonlin_e2e recipe template, all a bit shaky...

semaphoreP (approved):
Looks good now!

PR #442, opened 2025-05-20
Automatic Documentation of Data Types

semaphoreP:
## Describe your changes

New E2E test that runs at the very end that:

  1. Grabs the output of each type of data produced by the pipeline
  2. Produces a temporary documentation page in the e2e tests folder of the HDU data and headers (this is an output of the test and not tracked by git)
  3. Compares to the current version of the documentation page in the docs folder

This means that any change to the data format requires an explicit modification to the documentation pages in the docs folder for the e2e tests to pass. These changes will be tracked by git. 

## Type of change

- [x] I have linted my code


semaphoreP:
E2E tests also pass: https://github.com/roman-corgi/corgidrp/actions/runs/15148958827

maxwellmb:
A couple thoughts: 
- Do we want to add to the pages the file name extension format? I can see that at the moment the e2e tests don't conform to the expectation in all cases, but it's also something that's currently on the data format pages
- This infrastructure currently forces us to compare output values not just the presence/absence of the header keywords. Is this overly restrictive? It might be something to accept for now, and we make an issue to relax that. 
- What's the process to generate new pages when we agree that it is needed? Is it using the functions in `zz_dataformat_e2e.py`? If so, should they exist somewhere else? Maybe not, but thought I'd ask. 
-

semaphoreP:
Let's talk about these things at the DRP telecon tomorrow.

maxwellmb:
> A couple thoughts:
> 
>     * Do we want to add to the pages the file name extension format? I can see that at the moment the e2e tests don't conform to the expectation in all cases, but it's also something that's currently on the data format pages

I think we could do this, but it sounds like more work than we want for this PR. The other thing that just caught my attention is the example files currently on the confluence. If we want to get rid of those pages, then we should think of how to incorporate links into these pages for that too. That could be another PR later. 

> 
>     * This infrastructure currently forces us to compare output values not just the presence/absence of the header keywords. Is this overly restrictive? It might be something to accept for now, and we make an issue to relax that.

I think we decided that this is OK 

> 
>     * What's the process to generate new pages when we agree that it is needed? Is it using the functions in `zz_dataformat_e2e.py`? If so, should they exist somewhere else? Maybe not, but thought I'd ask.

We also covered this in the telecon workflow and I think its fine to leave this here as is. 

I think I'm ready to let it through. Any last things?

semaphoreP:
I'm working on some updates to ignore the values of certain header keywords, because the e2e tests do crash due to things like creation time. 

Incorporating example data is interesting... we currently do not the infrastructure to save the example data produced by the e2e tests and link them to readthedocs. We'd need to think about this a bit.

maxwellmb (approved):

Issue #418, opened 2025-04-30
Hardcoded calspec file list does not match list of absflux calibrators and always requires external download


ecady-jpl:
**Describe the bug**
Two concerns:

1. [Hardcoding the list of absolute flux calibrators](https://github.com/roman-corgi/corgidrp/blob/v2.2/corgidrp/fluxcal.py#L18-L36) means that they cannot be updated in ground software outside of a software release cycle.  Already the list differs from [the list being maintained by the Observation Planning group in the targets wiki](https://cpptargs.sioslab.com/index.php/Calibration_Targets#Absolute_Flux).
2. The calibrators are downloaded from a [hardcoded STScI URL](https://github.com/roman-corgi/corgidrp/blob/v2.2/corgidrp/fluxcal.py#L38) every call to `fluxcal.get_calspec_file()`.  This is vulnerable to both link rot and to lack of connectivity when using corgidrp.  Might be less of a big deal if `get_calspec_file()` wasn't called in other places (e.g. [every time `determine_app_mag()` in `l4_to_tda.py` is called](https://github.com/roman-corgi/corgidrp/blob/v2.2/corgidrp/l4_to_tda.py#L67).)

**Expected behavior**
It would be preferable if the list of calibrators was able to be updated outside of the software release cycle, and if no corgidrp commands required an external web connection to function.  If there is a particular finite set of data files that are getting repeatedly downloaded (e.g. Vega), it might be better to Git-LFS them with the repo rather than keep pulling them.

**corgidrp version:**
 - v2.2

JuergenSchreiber:
In l4_to_tda.determine_app_mag() and l4_to_tda.determine_color_cor() both ways are possible: you either input the name of the calspec star and retrieve the corresponding fits file from the url or you already have a calspec fits file and directly input this. 
In case of the flux and also the nd filter calibration it is indeed relying on the stars name (target header) and retrieve the calspec fits file from the internet. 
I think also here, we can easily adjust this and add an additional parameter, so that the user also can input a spectral model fits file that he wants.

ecady-jpl:
My concern with the external internet connection is not it getting overloaded, but it existing at all.  corgidrp isn't standalone if it requires a specific URL at STScI to continue to exist.

Call this a lesson learned: I have personally been on an observing run and had a piece of software we were about to use fail to start in the middle of the night because it couldn't connect to a license server.  Requiring unnecessary external connections can make the software implementation fragile.

juliamilton:
[...]

PR #471, opened 2025-06-24
Calspec

JuergenSchreiber:
## Describe your changes
To reduce the number of internet downloads of calspec fits files for absolute flux calibration and nd filter calibration
an direct interface to input calspec fits files is added. To further reduce the priority of internet download, the .corgidrp/calspec_data directory is scanned for already available local calspec fits files. Additionally  a json file with the association of star name to fits file name is generated in the same directory which can be extended.

## Type of change

#418

- [x] I have linted my code

### DIFF
--- a/corgidrp/nd_filter_calibration.py
+++ b/corgidrp/nd_filter_calibration.py
@@ -128,33 +128,42 @@ def compute_expected_band_irradiance(star_name, filter_name):
     return fluxcal.calculate_band_irradiance(transmission, calspec_flux, wave)
 
 
-def compute_avg_calibration_factor(dim_stars_dataset, phot_method, calspec_file = None, flux_or_irr="irr", phot_kwargs=None):
+def compute_avg_calibration_factor(dim_stars_dataset, phot_method, calspec_files = None, flux_or_irr="irr", phot_kwargs=None):
     """
     Compute the average flux calibration factor using dim stars (no ND filter).
 
     Parameters:
         dim_stars_dataset (iterable): Dataset containing dim star entries.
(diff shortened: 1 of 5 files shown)


corgidrp/nd_filter_calibration.py
```diff
@@ -125,32 +128,42 @@ def compute_expected_band_irradiance(star_name, filter_name):
...
 
-def compute_avg_calibration_factor(dim_stars_dataset, phot_method, flux_or_irr="irr", phot_kwargs=None):
+def compute_avg_calibration_factor(dim_stars_dataset, phot_method, calspec_files = None, flux_or_irr="irr", phot_kwargs=None):
     """
     Compute the average flux calibration factor using dim stars (no ND filter).
 
     Parameters:
         dim_stars_dataset (iterable): Dataset containing dim star entries.
         phot_method (str): Photometry method to use ("Aperture" or "Gaussian").
+        calspec_files (list, optional): list of calspec filepaths
```
semaphoreP, line 138: would be nice to have the option to supply a single filepath (filename?) in the case all of the dim star dataset is from a single star. So we don't need to copy the same str a bunch of times.

semaphoreP (commented):
Looks good overall! I left a bunch of mostly redundant comments. My main clarification is whether we are supplying filenames or filepaths? It seems like filenames and the files need to be in the calspec_data dir.

corgidrp/nd_filter_calibration.py
JuergenSchreiber: ok, I allow also a single str input

juliamilton:
Just reviewed the ND filter changes - looks good on my end

Issue #427, opened 2025-05-06
Spectroscopy: Estimate dispersion angle and fit dispersion profile


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Estimate dispersion angle and fit dispersion profile. Use as input the centroids computed by `spec.compute_psf_centroid()` (see PR #399). 

**Describe the solution you'd like**
Fit the dispersion parameters to input centroids using estimate_dispersion_clocking_angle() and fit_dispersion_polynomials(). Output a “Dispersion Calibration” class (does not necessarily need to be saved). New code should go in `spec.py`

See test_spectroscopy.py `test_dispersion_fit()` for reference code. Need to port over from spectroscopy branch.

JuergenSchreiber:
To better understand: Who else is also working on the spectroscopy branch? The stuff in spectroscopy.py is from Neil and should be ported to spec.py in corgi format? I  see some functions like Gaussian1/2d , centroiding, which I think we should import from astropy and/or photutils?

semaphoreP:
@Ezarsh02 worked on the first step (measuring centroids in prep for measuring dispersion clocking angle and dispersion profile), but is finishing her semester-long internship, so we have no one else currently.

JuergenSchreiber:
@semaphoreP thanks Jason for the info, should I then work on on the spectroscopy-test branch?

semaphoreP:
That was just used for the dev work that has already been PR'd back into spectroscopy. So, you should feel free to branch off of spectroscopy however you want. Thanks! Can I assign you to this issue?

JuergenSchreiber:
OK, I see, I was too fast, I am already assigned to it.

semaphoreP:
Great! Implementation note: I think you can take as input a `SpectroscopyCentroidPSF` instance, instead of the dictionaries in the original spectroscopy.py code.

JuergenSchreiber:
@semaphoreP Hi Jason, I saw that the spectroscopy branch is more than 1000 commits behind main, which might cause some problems. Should I pull main in or (develop?) first before going on?

semaphoreP:
Pull in develop. Thanks!

JuergenSchreiber:
[...]

PR #472, opened 2025-06-26
Spec dispersion

JuergenSchreiber:
## Describe your changes
convert code from Neil to corgidrp.
Improve centroid fitting of spectral trace.
Generate DispersionModel containing fit parameters of dispersion relation and clocking angle including errors.

## Type of change

#427

- [x] I have linted my code


semaphoreP:
Note to self: check usage of test data (e.g., headers).

tests/test_data/simastrom/guesses.csv
semaphoreP, line 1: is this needed for something or can it be deleted?

tests/test_data/spectroscopy/test_products/narrowband_LSF_fit.png
semaphoreP, line 1: for everything in the "test_products" folder, are they used by as input by anything, or can they be removed?

corgidrp/spectroscopy.py
semaphoreP, line 1: I'm ok keeping this file around for now, but we should eventually aim to remove it.

tests/test_spectroscopy.py
semaphoreP, line 1: the same comment for this function as spectroscopy.py

corgidrp/spec.py
```diff
@@ -0,0 +1,648 @@
...
+            raise ValueError("Mismatch between dataset length and centroid guess arrays.")
+    
+    if template_file == None:
+        if 'FILTERS' in dataset[0].pri_hdr:
+            default_file = default_psf_files.get('filtersweep')
+        elif dataset[0].pri_hdr['SLIT'] == 'N':
+            default_file = default_psf_files.get('3d_noslit')
+        else:
+            default_file = default_psf_files.get('3d_slit')
+        test_data_path = os.path.join(os.path.dirname(corgidrp.__path__[0]), "tests", "test_data", "spectroscopy")
```
semaphoreP, line 401: all data that will be used by step functions (not just in unit tests) should live in corgidrp/data/ rather than tests/test_data. The main reason is that when we package the code for PyPI/pip, the files in the test folder don't get packaged for release with it.

corgidrp/spec.py
```diff
@@ -0,0 +1,648 @@
...
+        template_file = os.path.join(test_data_path, 
+            default_file)
+        if verbose:
+            print("no template file given, take simulated file: " + default_file)
+    temp_psf_array = fits.getdata(template_file, ext=0)
+    if len(dataset) != np.shape(temp_psf_array)[0]:
```
semaphoreP, line 407: I don't think we can enforce that the data and template must be the same size. 

@neilzim in the real data, I presume the science dataset will not have the same number of frames as one of these template files. how do we select a template frame? just grab the first one? or some other criteria?

corgidrp/spec.py
```diff
@@ -0,0 +1,648 @@
...
+        raise ValueError("Mismatch between dataset length and template arrays.")
+    try:
+        temp_psf_table = fits.getdata(template_file , ext=1)
+        (xcent_template, ycent_template) = (temp_psf_table['xcent'], temp_psf_table['ycent'])
+        if len(dataset) != len(xcent_template) or len(dataset) != len(ycent_template):
+            raise ValueError("Mismatch between dataset length and template centroid arrays.")
```
semaphoreP, line 413: I don't think this valueerror does anything because it immediately gets silenced by the all-catching except statement in the next line.

corgidrp/spec.py
```diff
@@ -0,0 +1,648 @@
...
+        err_hdr = fits.Header(),
+        err = centroids_err,
+        input_dataset=dataset
+    )
+        
+    return calibration
+
+def read_cent_wave(filter_file, band):
+    """
+    read the csv filter file containing the band names and the central wavelength
```
semaphoreP, line 471: add a sentence in the docstring about the format of the csv file it assumes when reading (i.e., do you need column names, do wvs and transmission need to be at specific column indexes?)

corgidrp/spec.py
```diff
@@ -0,0 +1,648 @@
...
+        #bandpass = [610, 710]
+        #halfheight = 10
+    else:
+        subband_list = ['3A', '3B', '3C', '3D', '3E', '3G']
+        ref_wavlen = 730.0
+        ref_cfam = '3'
+        #bandpass = [675, 785]
+        #halfheight = 30
+        
+    filters = centroid_psf.pri_hdr['FILTERS'].upper().split(",")
```
semaphoreP, line 612: I couldn't figure out who writes "FILTERS" into the primary header. this isn't a keyword by default.  Perhaps we need to construct a list of filters from the CFAMNAME keyword in from the L1 headers. @neilzim may have added this keyword for convenience into the test files, but we may not get it for the real data

corgidrp/spec.py
```diff
@@ -0,0 +1,648 @@
...
+    return pfit_pos_vs_wavlen, cov_pos_vs_wavlen, pfit_wavlen_vs_pos, cov_wavlen_vs_pos
+
+
+def calibrate_dispersion_model(centroid_psf, band_center_file, prism = 'PRISM3', pixel_pitch_um = 13.0):
+    """ 
+    Generate a DispersionModel of the spectral dispersion profile of the CGI ZOD prism.
+
+    Args:
+       centroid_psf (SpectroscopyCentroidPsf): instance of SpectroscopyCentroidPsf calibration class
+       band_center_file (str): file name of the band centers
```
semaphoreP, line 586: should there be a default band_center_file that lives in corgidrp/data alongside the filter curves? so a user does not need to always pass in a file path?

tests/test_spec.py
```diff
@@ -0,0 +1,229 @@
...
+    
+    (xtest_min, xtest_max) = (tvac_pos_func_wavlen((bandpass[0] - ref_wavlen)/ref_wavlen),
+                              tvac_pos_func_wavlen((bandpass[1] - ref_wavlen)/ref_wavlen))
+    xtest = np.linspace(xtest_min, xtest_max, 1000)
+    tvac_model_wavlens = tvac_wavlen_func_pos(xtest)
+    corgi_model_wavlens = wavlen_func_pos(xtest)
+    wavlen_model_error = corgi_model_wavlens - tvac_model_wavlens
+    worst_case_wavlen_error = np.abs(wavlen_model_error).max()
+    print(f"Worst case wavelength disagreement from the test input model: {worst_case_wavlen_error:.2f} nm")
+    assert worst_case_wavlen_error == pytest.approx(0, abs=0.5)
```
semaphoreP, line 222: did you pick 0.5 nm specifically because it is a requirement or just arbitrary?

semaphoreP (commented):
Looks good! I have some clean-up to-dos, some refactoring comments, and a couple of comments on how the code will work on real data. There may be 1-2 things we should consult with Neil.

corgidrp/spec.py
neilzim: I set this "FILTERS" list in the test filter sweep image cube for convenience. It would be preferable to have the test operate on a list of individual image files in the expected format.

corgidrp/spec.py
JuergenSchreiber: Ok, I move everything we need to data/spectroscopy

corgidrp/spec.py
JuergenSchreiber: Done

tests/test_data/spectroscopy/test_products/narrowband_LSF_fit.png
JuergenSchreiber: can be removed

corgidrp/spec.py
JuergenSchreiber: I have converted now the tvac template files to single drp style images and stored tehm in data/spectroscopy/templates and added x/ycent and x/yoffset to the ext_hdrs, since I am not sure whether we have already corresponding L1 header keywords.
For now I use the template frame one after the other, if we have more frames in the input dataset I just use the last template frame. For filtersweep I search the available template frames for the same cfamname.

semaphoreP: based on the slack conversation with Neil, I think we just need one template per filter? Does that work for you Juergen?

corgidrp/spec.py
```diff
@@ -0,0 +1,629 @@
...
+    else:
+        xcent = np.asarray(initial_cent.get("xcent"))
+        ycent = np.asarray(initial_cent.get("ycent"))
+        if xcent is None or ycent is None:
+            raise ValueError("initial_cent dictionary must contain 'xcent' and 'ycent' arrays.")
+        if len(dataset) != len(xcent) or len(dataset) != len(ycent):
+            raise ValueError("Mismatch between dataset length and centroid guess arrays.")
+    
+    if template_dataset is None:
+        template_dataset, filtersweep = get_template_dataset(dataset)
```
semaphoreP, line 355: It seems like this line overrides the filtersweep argument passed in. Perhaps the filtersweep docstring comment could emphasize that it's only used if template_dataset is not None?

corgidrp/spec.py
```diff
@@ -0,0 +1,629 @@
...
+        )
+
+        centroids[idx] = [xfit, yfit]
+        centroids_err[idx] = [x_precis, y_precis]
+
+        if verbose:
+            print(f"Slice {idx}: x = {xfit:.3f}, y = {yfit:.3f}")
+    
+    if filtersweep:
+        ext_hdr_centroid['FILTERS'] = ",".join(filters)
```
semaphoreP, line 430: It seems to me that the FILTERS keyword is absolutely necessary for the next step. Is there a case where we aren't doing a filtersweep? Shold we just assume we are, and that by default, we write something into this FILTERS header?

semaphoreP (commented):
Just a couple small comments here, and there's one unresolved comment from the last review regarding what are the necessary template files that I'd like to get to the bottom of.

corgidrp/spec.py
JuergenSchreiber: I was not sure whether we use the centroid fitting only for the dispersion determination. At least we could use it also for other things, e.g. we have the psfs with offsets as templates in data.

Issue #428, opened 2025-05-06
Spectroscopy: Create Wavelength Solution


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Create Wavelength Solution (wavelength-to-pixel map) and save as calibration file

**Describe the solution you'd like**
Using Dispersion Calibration and an input zero-point coordinate (manual for now potentially(?) -- look into automating it), make a Wavelength Solution with create_wave_cal_map().

**Additional context**
Needs issue #427 to produce dispersion calibration
See test_spectroscopy.py `test_wave_cal` function for guidance.

hsergi:
Notice that the source position needs to be estimated by a combination of target knowledge, star location on EXCAM and EXCAM orientation and pixel scale (see also Neil's comments on Spectroscopy FDD on confluence: [link](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/154410149/Spectroscopy+FDD+development+area)).

If this information is not captured yet (as a Header keyword, for instance), use a place holder. For instance, input keyword parameters in the step function. Consult!

semaphoreP:
> Notice that the source position needs to be estimated by a combination of target knowledge, star location on EXCAM and EXCAM orientation and pixel scale (see also Neil's comments on Spectroscopy FDD on confluence: [link](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/154410149/Spectroscopy+FDD+development+area)).
> 
> If this information is not captured yet (as a Header keyword, for instance), use a place holder. For instance, input keyword parameters in the step function. Consult!

Note that this issue just creates the wavelength calibration file, not solve for the wavelength of any arbitrary science data.

hsergi:
This is part of the FDD AAC requirements, which are reviewed and accepted. [AAC](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/106725541/JPL+Functional+Design+Descriptions#JPLFunctionalDesignDescriptions-SpectroscopicCalibrations). Work can start w/o any further delay

hsergi:
NOTE: this is an L3->L4 step in the Data Processing pipeline

PR #478, opened 2025-07-29
Spectroscopy wavelength calibration map

JuergenSchreiber:
## Describe your changes
Added a WavelengthZeropoint and WaveCal calibration classes to data, and converted create_wave_cal to spec.py.
It can be chosen whether WaveCal adds a position lookup table extension or not.

## Type of change

#428

- [x] I have linted my code


tests/test_spec.py
```diff
@@ -306,7 +310,7 @@ def test_calibrate_dispersion_model():
     psf_centroid.xfit_err = psf_centroid.xfit_err[:-1]
     psf_centroid.yfit_err = psf_centroid.yfit_err[:-1]
     
-    disp_model = steps.calibrate_dispersion_model(psf_centroid, prism = 'PRISM3')
+    disp_model = steps.calibrate_dispersion_model(psf_centroid)
```
neilzim, line 313: The relative image shifts between the sub-band filters are here accounted for in test_calibration_dispersion_model(), but not inside the calibration function. Instead, the image shifts will need to be considered an input to calibration_dispersion_model().

neilzim (changes requested):

corgidrp/l3_to_l4.py
```diff
@@ -809,6 +810,49 @@ def northup(input_dataset,use_wcs=True,rot_center='im_center'):
     return processed_dataset 
 
 
+def wave_cal(input_dataset, disp_model, pixel_pitch_um = 13.0):
```
semaphoreP, line 813: could we make this step function name a bit more descriptive, just to make it a bit more distinctive from `spec.create_wave_cal`. Maybe something like `add_wavelength_solution` or `apply wavelength_calibration`?

corgidrp/spec.py
```diff
@@ -613,8 +613,115 @@ def calibrate_dispersion_model(centroid_psf, band_center_file = None, prism = 'P
...
+    wavlen_map = wavlen_vs_pos_poly(dd_mm)
+
+    delta_wav = 0.5
+    n_wav = int(ref_wavlen * bandpass_frac / delta_wav)
+    n_wav_odd = n_wav + (n_wav % 2 == 0) # force odd array length
+    wavlen_beg = ref_wavlen - n_wav_odd // 2 * delta_wav
+    wavlen_end = ref_wavlen + n_wav_odd // 2 * delta_wav
+    
+    wavlens = np.linspace(wavlen_beg, wavlen_end, n_wav_odd)
+    np.testing.assert_almost_equal(wavlens[n_wav_odd // 2], ref_wavlen)
```
semaphoreP, line 673: I generally avoid using asserts in production code (only asserts in tests). Is there a reason we are worried about these not being equal?

corgidrp/spec.py
```diff
@@ -613,8 +613,115 @@ def calibrate_dispersion_model(centroid_psf, band_center_file = None, prism = 'P
...
+        
+    # Use a Monte Carlo error propagation to estimate the uncertainties of the
+    # values in the wavelength calibration map and the position lookup table. 
+    ntrials = 1000
```
semaphoreP, line 677: this is probably good to have as an optional keyword in create_wave_cal and wave_cal function, in case the user wants a different number of ntrials.

corgidrp/data/spectroscopy/templates/spec_unocc_noslit_offset_prism3_3d_12.fits
semaphoreP, line 1: is this file only used for testing? if so it should go in tests/test_data. If it's also used in the corgidrp modules, then it can remain here.

corgidrp/data/spectroscopy/templates/spec_unocc_r1c2slit_offset_prism3_3d_12.fits
semaphoreP, line 1: is this file only used for testing? if so it should go in tests/test_data. If it's also used in the corgidrp modules, then it can remain here.

tests/test_spec.py
```diff
@@ -334,10 +338,55 @@ def test_calibrate_dispersion_model():
     print(f"Worst case wavelength disagreement from the test input model: {worst_case_wavlen_error:.2f} nm")
     assert worst_case_wavlen_error == pytest.approx(0, abs=0.5)
     print("Dispersion profile fit test passed.")
+
+def test_wave_cal():
+    """
+    test l3_to_l4.wave_cal(), the generation of the wavelength map extensions
```
semaphoreP, line 344: should comment in the docstring that this test requires running `test_calibrate_dispersion_model()` beforehand.

tests/test_spec.py
```diff
@@ -334,10 +338,55 @@ def test_calibrate_dispersion_model():
...
+    image.ext_hdr['shapex0'] = wave_0.get('shapex')
+    image.ext_hdr['shapey0'] = wave_0.get('shapey')
+    dataset = Dataset([image])
+    
+    output_dataset = l3_to_l4.wave_cal(dataset, disp_model)
+    
+    out_im = output_dataset.frames[0]
+    
+    wave = out_im.hdu_list["wave"].data
+    assert wave.shape == (wave_0.get('shapex'), wave_0.get('shapey'))
```
semaphoreP, line 372: is there any way to test the values of the wavelength solution in the unit test? or is it too uncertain?

semaphoreP (commented):
Just some small comments. Overall, this looks good to me!

corgidrp/l3_to_l4.py
JuergenSchreiber: I called it now add_wavelength_map()

tests/test_spec.py
JuergenSchreiber: ok, I try to set the zeropoint at the max of the narrow band template, and the ref_wavelen should be approx in the center of the array

tests/test_spec.py
```diff
@@ -315,7 +319,7 @@ def test_calibrate_dispersion_model():
     wavlen_func_pos = np.poly1d(disp_model.wavlen_vs_pos_polycoeff)
```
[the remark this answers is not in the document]
JuergenSchreiber: ok, done

corgidrp/spec.py
```diff
@@ -613,8 +613,115 @@ def calibrate_dispersion_model(centroid_psf, band_center_file = None, prism = 'P
...
+    #bandpass_frac = fwhm/cen_wave
+    bandpass_frac = disp_model.ext_hdr["BANDFRAC"]
+    pos_vs_wavlen_poly = np.poly1d(disp_model.pos_vs_wavlen_polycoeff)
+    wavlen_vs_pos_poly = np.poly1d(disp_model.wavlen_vs_pos_polycoeff)
+    wavlen_c = ref_wavlen
+    d_zp_mm = pos_vs_wavlen_poly((wave_zeropoint.get('wavlen') - wavlen_c) / wavlen_c)
+
+    pixel_pitch_mm = pixel_pitch_um * 1E-3
+    theta = np.deg2rad(disp_model.clocking_angle)
+    x_c, y_c = (wave_zeropoint.get('x') - d_zp_mm * np.cos(theta) / pixel_pitch_mm,
```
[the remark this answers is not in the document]
JuergenSchreiber: ok, done

corgidrp/data.py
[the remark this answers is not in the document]
neilzim: The offsets should remain fixed. They would be stored in EXCAM pixel units.

Issue #450, opened 2025-05-23
Update L1 file name and header & propagate changes


agreenbaum:
**Is your feature request related to a problem? Please describe.**
File names need to be changed to lowercase, by request of SOC/MAST and the L1 header keywords are receiving a small update (additional values)

**Describe the solution you'd like**
The new file name format needs to be updated to lowercase letters throughout all data products. New header keywords should be added to L1 model. It probably only makes sense to propagate "DETECTOR" if any of the new keywords.

**Additional context**
Three new header keywords: 
DETECTOR (always set to EXCAM)
FILENAME (name of the file)
FSWDVERS (version of CGI FSW dictionary used to process the telemetry)
See: https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/161617086/L1+Current+DRP+Development+Version+-+SSC+Release+3.3.0

ecady-jpl:
Does this include the capital T in the middle of the datestamp?

agreenbaum:
Not for SSC current release, no. I do need to check with SOC if that is ok or if we'll have to make another change.

maxwellmb:
@agreenbaum, any updates on the T?

agreenbaum:
Thanks for the reminder. We'll need to make the "T" lowercase to conform with SOC.

PR #480, opened 2025-08-04
Update mocks.py to make mock L1 headers use rst documentation

juliamilton:
## Describe your changes
Update to mocks.py that uses rst documentation of L1 file headers to product correct mock L1 headers.

## Type of change

- [ ] I have linted my code


juliamilton:
This PR is to address updating L1 headers/ propagation of headers per #450. I've made it so that the L1 mocks read from an `l1.rst` file that we can keep updated more easily from changes on confluence. Those L1 headers then propagate to higher level mock data products.

I also have updates that make all of the filenames lowercase, which is the other part of that issue, but it affects many files and may be difficult to review all the changes. Do you want me to do that PR separately?

corgidrp/mocks.py
```diff
@@ -102,382 +103,315 @@
...
     exthdr = fits.Header()
+    
+    # Set up dynamic values
+    dt_str = datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S")
+    # Override NAXIS values to match test expectations (1024x1024)
+    # L1.rst documents actual detector dimensions (2200x1200 for SCI, 2200x2200 for ENG)
+    NAXIS1 = 1024
+    NAXIS2 = 1024
+    
+    def parse_rst_table(rst_file_path, section_name, stop_at=None):
```
semaphoreP, line 129: due to how python packaging works, we can't have our code depend on the docs (docs can only depend on code). For any use that installs the code not in developer mode (i.e., the CTC I believe does this), the docs folder is not "installed". Thus, we cannot use the .rst table. Unfortunately, we'll need to use the previous code we had.

corgidrp/mocks.py
```diff
@@ -102,382 +103,315 @@
...
```
semaphoreP, line 273: same issue here as above

corgidrp/mocks.py
```diff
@@ -2515,7 +2449,7 @@ def add_defect(sch_imgs, prob, ori, temp):
             add_1_dipole(temps[temp][1], 33, 77, 'below', 'sp', 0, 100, temp)
             # add in 'CENel1' trap for all phase times
         #    add_1_dipole(temps[temp][3], 26, 28, 'below', 'mf2', 0, 100, temp)
-        #    add_1_dipole(temps[temp][4], 26, 28, 'above', 'mf2', 0, 100, temp)
+    #    add_1_dipole(temps[temp][4], 26, 28, 'above', 'mf2', 0, 100, temp)
```
semaphoreP, line 2452: is this intend removal and the one below intentional?

corgidrp/walker.py
```diff
@@ -221,7 +221,11 @@ def _fill_in_calib_files(step, this_caldb, ref_frame):
                     step["skip"] = True # skip this step but continue
                     step["calibs"][calib] = None
                     warnings.warn("Skipping {0} because no {1} in caldb and skip_missing_cal_steps is True".format(step['name'], calib))
-                    continue # continue on the for loop
+                    # Don't continue - set all remaining calibrations to None and return
```
semaphoreP, line 224: I don't understand what you are trying to do here? The previous behavior was intentional.

docs/source/data_formats/l1.rst
```diff
@@ -0,0 +1,350 @@
...
++------------+------------+--------------------------------+----------------------------------------------------+
+| CAMPAIGN   | str        | 001                            | The Pass/Campaign in visit hierarchy (digits 8-10) |
++------------+------------+--------------------------------+----------------------------------------------------+
+| SEGMENT    | str        | 001                            | The Segment Number in visit hierarchy (digits 11-13) |
++------------+------------+--------------------------------+----------------------------------------------------+
+| OBSNUM     | str        | 001                            | The Observation Number in visit hierarchy (digits 14-16) |
++------------+------------+--------------------------------+----------------------------------------------------+
+| VISNUM     | str        | 001                            | The Visit number in visit hierarchy (digits 17-19) |
++------------+------------+--------------------------------+----------------------------------------------------+
+| CPGSFILE   | str        | campaign_0200001001001001001.xml | Campaign-level XML containing the current visit    |
```
semaphoreP, line 71: have you tried building this doc page? I think lines like this where the delimteres aren't exactly lined up will break the docs building.

docs/source/data_formats/l1.rst
```diff
@@ -0,0 +1,350 @@
...
+| FRAMET     | float      | 30.0                           | Planned exposure time per frame or if set to "AUTO". Taken from the observation plan |
++------------+------------+--------------------------------+----------------------------------------------------+
+| SATSPOTS   | int        | 0                              | Whether satellite spots are present                |
++------------+------------+--------------------------------+----------------------------------------------------+
+| ISHOWFSC   | int        | 0                              | Images taken as a part of HOWFSC                  |
++------------+------------+--------------------------------+----------------------------------------------------+
+| HOWFSLNK   | int        | 0                              | Does the campaign include a HOWFSC activity        |
++------------+------------+--------------------------------+----------------------------------------------------+
+| FILENAME   | str        | cgi_0200001001001001001_20240403t105401_l1_.fits | The name of the file                               |
```
semaphoreP, line 111: same as above

semaphoreP (changes requested):
Unfortunately, I don't think we can have mocks.py use the documentation source to generate headers. A few other minor comments too.

corgidrp/mocks.py
juliamilton: not intentional, i reverted both

corgidrp/walker.py
juliamilton: this was for debugging and got left in by accident - i reverted it

docs/source/data_formats/l1.rst
juliamilton: Sounds good, I agree it might be a lot to maintain. I deleted the l1.rst file from the docs. I think that is the final change from me on this PR unless you have other items to address

Issue #449, opened 2025-05-23
Spectroscopy: WCS


hsergi:
**Is your feature request related to a problem? Please describe.**
The change in optics for spectroscopy mode will introduce some translation on EXCAM but no magnification or rotation. We can use the same plate scale and orientation from imaging, but we will need to account for an x,y translation on EXCAM when we switch to spectroscopy optics.

**Describe the solution you'd like**

Define the WCS for spectroscopy in L2b→L3, as it is the case for the Imaging mode. Distortion correction is **not** needed.

Note that, although we want to carry over a record of the L3 plate orientation, we should never rotate the prism-dispersed images or the bad pixel maps. We should only translate to align and combine.

semaphoreP:
I think the step function only needs to be updated so that we can pass in an arbitrary [x,y] shift where the center of the WCS coordinate system is defined.

PR #494, opened 2025-08-18
add xy wcs offset

wbalmer:
## Describe your changes
Add a x,y arbitrary offset to the `create_wcs` step.

Addresses #449 

- [x] I have linted my code

### DIFF
--- a/corgidrp/l2b_to_l3.py
+++ b/corgidrp/l2b_to_l3.py
@@ -13,6 +13,7 @@ def create_wcs(input_dataset, astrom_calibration, offset=None):
         input_dataset (corgidrp.data.Dataset): a dataset of Images (L2b-level)
         astrom_calibration (corgidrp.data.AstrometricCalibration): an astrometric calibration file for the input dataset
         offset (optional, tuple(float, float)): x and y offset in units of pixel between the dataset and WCS center (for spectroscopy or other optics offset from imaging mode)
+    
     Returns:
         corgidrp.data.Dataset: a version of the input dataset with the WCS headers added
     """
--- a/corgidrp/l2b_to_l3.py
+++ b/corgidrp/l2b_to_l3.py
@@ -12,7 +12,7 @@ def create_wcs(input_dataset, astrom_calibration, offset=None):
     Args:
         input_dataset (corgidrp.data.Dataset): a dataset of Images (L2b-level)
         astrom_calibration (corgidrp.data.AstrometricCalibration): an astrometric calibration file for the input dataset
-        offest (optional, tuple(float, float)): x and y offset in units of pixel between the dataset and WCS center (for spectroscopy or other optics offset from imaging mode)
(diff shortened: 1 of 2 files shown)


semaphoreP:
For the tests to stop failing, I think you need a new line in the docstring between the definition of "offset" and the "Returns:" line.

corgidrp/l2b_to_l3.py
```diff
@@ -29,6 +30,9 @@ def create_wcs(input_dataset, astrom_calibration):
         image_y, image_x = im_data.shape
         center_pixel = [(image_x-1) // 2, (image_y-1) // 2]
         target_ra, target_dec = image.pri_hdr['RA'], image.pri_hdr['DEC']
+        if offset is not None:
```
semaphoreP, line 33: I think we want to offset the central pixel by of the offset (i.e., CRPIX values in the header) instead of the CRVALS. The use case is that we're pointed at a star for spectroscopy, we know the RA/Dec of the star, but it's shifted from the nominal central of the image. We use this to shift the WCS onto the star, rather than finding the new RA/Dec of the central pixel.

tests/test_northup.py
```diff
@@ -6,7 +6,7 @@
 from matplotlib import pyplot as plt
 import numpy as np
 
-def test_northup(save_mock_dataset=False,save_derot_dataset=False,save_comp_figure=False):
+def test_northup(save_mock_dataset=False,save_derot_dataset=False,save_comp_figure=False,test_offset=False):
```
semaphoreP, line 9: keywords actually don't work in pytest automated testing (at least in how we set it up). You should just make a new test to test your offset.

semaphoreP: it'll be clearing then when the assert fails, why it failed. Right now, it's unclear if the assert fails, whether that's due to your changes or due to issues in the northup step function.

semaphoreP (commented):
Thanks for adding this. A couple comments on the implementation that I think we should change.

corgidrp/l2b_to_l3.py
wbalmer: makes sense! i just changed to update the center_pixel

tests/test_northup.py
```diff
@@ -156,5 +156,50 @@ def test_northup(save_mock_dataset=False,save_derot_dataset=False,save_comp_figu
...
+   astrom_cal = astrom.boresight_calibration(mock_dataset_ori, fieldpath, find_threshold=10)
+
+   mock_dataset =  mock_dataset_ori.copy()
+
+   # add an angle offset
+   mock_dataset[0].pri_hdr['ROLL']=(ang,'roll angle (deg)')
+
+   # create the wcs
+   updated_dataset = create_wcs(mock_dataset,astrom_cal,offset=(3.3, 1.0))
+   # inject fake sources for test
```
semaphoreP, line 193: there should be some assert statements to check that the function did what you expect (e.g., CRPIX is at a different location).

semaphoreP (commented):
just one more minor comment.

semaphoreP (approved):

PR #523, opened 2025-09-08
disable spectroscopy PSF calibration file from being generated by default

kjl0025:
## Describe your changes
I disabled the spectroscopy PSF calibration file from being generated by default and removed the data/spectroscopy file that is not needed (other than for testing).

## Type of change

- [x] I have linted my code


.gitattributes
```diff
@@ -5,6 +5,5 @@ tests/test_data/medcombined-uranus_band_4.fits filter=lfs diff=lfs merge=lfs -te
 tests/test_data/medcombined-uranus_band_1.fits filter=lfs diff=lfs merge=lfs -text
 tests/test_data/mock_northup.fits filter=lfs diff=lfs merge=lfs -text
 tests/test_data/spectroscopy/g0v_vmag6_spc-spec_band3_unocc_CFAM3_R1C2SLIT_PRISM3_offset_array.fits filter=lfs diff=lfs merge=lfs -text
-corgidrp/data/spectroscopy/g0v_vmag6_spc-spec_band3_unocc_CFAM3d_NOSLIT_PRISM3_offset_array.fits filter=lfs diff=lfs merge=lfs -text
```
semaphoreP, line 8: should we be tracking the new location of the file to Git LFS?

semaphoreP (commented):
One quick question for you

.gitattributes
kjl0025: Oh yes, I was thinking it was less than 1MB and was errantly in LFS, but it indeed needs to be there. I'll fix

semaphoreP (approved):

Issue #502, opened 2025-08-20
caldb should pick calibrations based on filter/mask/mode when appropriate


semaphoreP:
**Is your feature request related to a problem? Please describe.**
caldb should be considering the wavelength filter, the coronagraph mask, and mode (imaging, spec, pol) when considering some calibrations (e.g., filter for abs flux and ND, mask for core throughput calibrations, mode for flat field and some others)

**Describe the solution you'd like**
Add new columns to caldb that tracks these properties and add logic that filters by them.

semaphoreP:
Implemented in #536

PR #536, opened 2025-09-11
Add additional columns to caldb and extra criteria for calib selection

everlastingEric:
## Describe your changes
- add new columns to caldb: CFAMNAME, DPAMNAME, and FPAMNAME
- caldb is now backwards compatible, older files with missing columns will have those said columns populated with default values
- for abs flux and ND, caldb now tries to find files with matching color filters
- for ct cal, caldb now tries to find files with matching focal plane masks
- for flats, caldb now tries to find files with matching DPAM prisms/lenses

## Type of change

#502 

- [x] I have linted my code


semaphoreP (approved):
Looks good

Issue #535, opened 2025-09-11
Move cropping into L2b -> L3 stage


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Move crop to L2b to L3 stage to save space. 

**Describe the solution you'd like**
 * Move crop step
 * Change it to use EACQ_ROW and EACQ_COL header keywords rather than star center position from sat spots to figure out crop center. 
 * Maybe give a few pixel buffer on the crop in case the keywords are a few pixels off

semaphoreP:
Completed in PR #539

PR #539, opened 2025-09-11
Crop Upgrades

ell-bogat:
## Describe your changes
Upgrades to the crop step function!
- crop() function has been moved from l3_to_l4.py to l2a_to_l3.py
- Removed crop call within psf subtraction step function (and associated tests
- Updated recipe templates and walker
- Updated default center keywords from `STARLOCX/Y` to `EACQ_ROW/COL`
- Made default array sizes odd, but made sure that we can handle either odd or even shapes to crop to.
- Now we can pad with nans if you try to crop outside the existing array bounds
- Updated crop tests 
  - [x] Add cropping off array edges (all 4 directions)
  - [x] Add even, odd, and mixed shape crops
  - [x] Make sure dq and error arrays are also cropped properly for 2D and 3D case
  - [x] Make sure STARLOCX/Y is updated properly if it exists in the headers
- Updated psf subtraction tests to remove cropping

## Type of change

Should close Issue #535

- [x] I have linted my code

### DIFF
--- a/tests/test_crop.py
+++ b/tests/test_crop.py
@@ -288,7 +288,7 @@ def test_unsupported_input():
 
     test_dataset = make_test_dataset(input_arr_even,centxy=[50.5,50.5])
     for frame in test_dataset:
-        frame.ext_hdr['LSAMNAME'] = 'OPEN'
+        frame.ext_hdr['LSAMNAME'] = 'UNKNOWN'
 
     try:
         _ = crop(test_dataset,sizexy=20,centerxy=None)
@@ -299,6 +299,17 @@ def test_unsupported_input():
     with pytest.raises(UserWarning):
         _ = crop(test_dataset,sizexy=None,centerxy=None)
 
+def test_noncoron():
+    """ Crop function should not automatically crop non-coronagraphic data.
+    """
+
+    test_dataset = make_test_dataset(input_arr_even,centxy=[50.5,50.5])
+    for frame in test_dataset:
+        frame.ext_hdr['LSAMNAME'] = 'OPEN'
+
+    cropped_dataset = crop(test_dataset,sizexy=None,centerxy=None)
+
+    assert cropped_dataset.all_data == pytest.approx(test_dataset.all_data)
 
 def test_detpix0_nonzero():
     """ Tests that the detector pixel header keyword is updated correctly if it 
@@ -404,6 +415,7 @@ if __name__ == "__main__":
     test_non_nfov_input()
     test_detpix0_nonzero()
     test_unsupported_input()
+    test_noncoron()
     test_default_crop()
     test_mixed_oddeven_crop()
     test_missing_cen_kws()


tests/test_crop.py
```diff
@@ -77,147 +138,199 @@ def test_manual_center_crop():
...
+    test_dataset = make_test_dataset(test_arr,centxy=[94.5,94.5])
     cropped_test_dataset = crop(test_dataset,sizexy=10,centerxy=None)
 
-    if not cropped_test_dataset[0].data == pytest.approx(goal_arr):
+    if not cropped_test_dataset[0].data == pytest.approx(goal_arr_even):
         raise Exception("Unexpected result for edge of FOV crop test.")
 
+
 def test_outside_detector_edge():
     """ Tests that trying to crop a region outside the detector fails.
```
semaphoreP, line 235: this test doesn't seem to test cropping out of detector edges, but tests cropping at exactly the detector edge.

tests/test_klipthrupt.py
```diff
@@ -1104,191 +1099,6 @@ def test_psfsub_withklipandctmeas_rdi():
     assert pl_counts == pytest.approx(recovered_pl_counts_ktcorrected,rel = 0.05) 
 
 
-def test_psfsub_withKTandCTandCrop_adi():
```
semaphoreP, line 1107: would it still be useful to check that KLIP throughput and core throughput can be run as part of the PSF subtraction process? If we need to crop down the data to make it faster to run, we can run the crop step function before running do_psf_subtraction.

corgidrp/recipe_templates/l2b_to_l3.json
semaphoreP, line 1: I think the cropping breaks the e2e tests for l2b_to_l4_e2e (I think we're not cropping to the right area by default) and l2b_to_l4_noncoron_e2e (we're cropping the image even when we shouldn't be). Could you fix them? I think it's just fixing header keywords of the input simulated l2b data.

corgidrp/l2b_to_l3.py
```diff
@@ -244,6 +247,192 @@ def split_image_by_polarization_state(input_dataset,
...
+    crop the image to the dark hole region, centered on the pixel intersection nearest 
+    to the star location. Assumes 3D Image data is a stack of 2D data arrays, so only 
+    crops the last two indices.
+
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (any level)
+        sizexy (int or array of int): desired frame size, if only one number is provided the 
+            desired shape is assumed to be square, otherwise xy order. If not provided, 
+            defaults to 61 for NFOV (narrow field-of-view) observations. Defaults to None.
+        centerxy (float or array of float): desired center (xy order), should be a pixel intersection (a.k.a
```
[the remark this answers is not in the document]
ell-bogat: I just clarified the text! I had forgotten to update it.

tests/test_crop.py
ell-bogat: It crops to a region that is centered on the detector edge, but has a width that means the output array is partially off the detector edge. Happy to clarify this in the test name/definition.

corgidrp/recipe_templates/l2b_to_l3.json
ell-bogat: I need to find the e2e test data in order to fix this locally- will try to get it done tomorrow!

tests/test_crop.py
semaphoreP: maybe just edit the docstring to stay test cropping to a region that partially falls outside of the detector edge.

semaphoreP (approved):

PR #540, opened 2025-09-12
updated pol abs fluxcal e2e test

everlastingEric:
## Describe your changes
Updated pol abs fluxcal e2e test with additional checks on input and output data to conform with VAP testing requirements

## Type of change

- [x] I have linted my code

### DIFF
--- a/tests/e2e_tests/fluxcal_pol_e2e.py
+++ b/tests/e2e_tests/fluxcal_pol_e2e.py
@@ -135,13 +135,13 @@ def test_expected_results_e2e(e2edata_path, e2eoutput_path):
         logger.info(f"HDU1 dtype: {fluxcal_image_WP1.data.dtype.type}. Expected: float. FAIL.")
     # check err and dq have the right dimension
     if fluxcal_image_WP1.err.shape == (1,):
-        logger.info(f"Error data shape: (1,1). Expected: (1,1). PASS.")
+        logger.info(f"Err data shape: (1,). Expected: (1,). PASS.")
     else:
-        logger.info(f"Error data shape: {fluxcal_image_WP1.err.shape}. Expected: (1,1). FAIL.")
+        logger.info(f"Err data shape: {fluxcal_image_WP1.err.shape}. Expected: (1,). FAIL.")
     if fluxcal_image_WP1.dq.shape == (1,):
-        logger.info(f"DQ data shape: (1,1). Expected: (1,1). PASS.")
+        logger.info(f"DQ data shape: (1,). Expected: (1,). PASS.")
     else:
-        logger.info(f"DQ data shape: {fluxcal_image_WP1.dq.shape}. Expected: (1,1). FAIL.")
+        logger.info(f"DQ data shape: {fluxcal_image_WP1.dq.shape}. Expected: (1,). FAIL.")
     # check filename convention
     check_filename_convention(getattr(fluxcal_image_WP1, 'filename', None), 'abf_cal.fits', logger=logger)
     # check header keyword values match with what is expected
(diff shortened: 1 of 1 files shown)


tests/e2e_tests/fluxcal_pol_e2e.py
```diff
@@ -60,46 +137,108 @@ def test_expected_results_e2e(e2edata_path, e2eoutput_path):
...
+    else:
+        logger.info(f"HDU0: Contains data. Expected: header only. FAIL.")
+    # check HDU1 data is a single float
+    if fluxcal_image_WP1.data.dtype.type == corgidrp.image_dtype:
+        logger.info(f"HDU1 dtype: float. Expected: float. Pass.")
+    else:
+        logger.info(f"HDU1 dtype: {fluxcal_image_WP1.data.dtype.type}. Expected: float. Fail.")
+    # check err and dq haave the right dimension
+    if fluxcal_image_WP1.err.shape == (1,):
+        logger.info(f"Error data shape: (1,1). Expected: (1,1). Pass.")
```
semaphoreP, line 165: Why is this Error and Pass? 

Also, when I run it, I do get to this logic, but I believe this is the correct behavior. The expected shape is indeed (1,)

tests/e2e_tests/fluxcal_pol_e2e.py
```diff
@@ -60,46 +137,108 @@ def test_expected_results_e2e(e2edata_path, e2eoutput_path):
...
-    assert fluxcal_image_WP1.dq.shape == (1,)
+    logger.info('='*80)
+    logger.info('Test Case 2: Output Calibration Product Data Format and Content for WP1')
+    logger.info('='*80)
+    ## check that the calibration file is configured correctly
+    # check HDU0 have no data
+    if fluxcal_image_WP1.pri_hdr == 0:
+        logger.info("HDU0: Header only. Expected: header only. PASS.")
+    else:
+        logger.info(f"HDU0: Contains data. Expected: header only. FAIL.")
```
semaphoreP, line 157: I got this logic when I ran it. Do you? If so, we should fix this. I think the problem is that you need to check the size of data in HDU0, not whether the header of HDU0 == 0.

tests/e2e_tests/fluxcal_pol_e2e.py
```diff
@@ -60,46 +137,108 @@ def test_expected_results_e2e(e2edata_path, e2eoutput_path):
...
+    verify_header_keywords(fluxcal_image_WP2.ext_hdr, {'DATALVL': 'CAL'}, logger=logger)
+    verify_header_keywords(fluxcal_image_WP2.ext_hdr, {'DATATYPE': 'FluxcalFactor'}, logger=logger)
+    verify_header_keywords(fluxcal_image_WP2.ext_hdr, {'DPAMNAME': flux_image_WP2.ext_hdr['DPAMNAME']}, logger=logger)
+    verify_header_keywords(fluxcal_image_WP2.ext_hdr, {'CFAMNAME': flux_image_WP2.ext_hdr['CFAMNAME']}, logger=logger)
+    logger.info("")
 
-    #output values
+    # baseline performance check WP2
+    logger.info('='*80)
+    logger.info('Test Case 5: Baseline Performance Checks for WP2')
```
semaphoreP, line 234: The VAP testing docs asks us to check "Print FSMX and FSMY values of input image(s), FSMX and FSMY values of OD, and deltas". Can you do that?

semaphoreP (commented):
A few comments to make sure we are in line with VAP testing.

tests/e2e_tests/fluxcal_pol_e2e.py
everlastingEric: sorry I think the wording is a little confusing, this one is checking the error array and if the shape is (1,) it passes

tests/e2e_tests/fluxcal_pol_e2e.py
everlastingEric: I asked Max about it since I was a bit confused on what those are and he said they weren't important. If I should print them out anyways, do you know which headers contain the FSMX, FSMY values of OD, and deltas?

everlastingEric: @semaphoreP for the FSMX and FSMY values of the inputs, they are already being outputted earlier in the code when I do the other input checks, is that fine as is or should I move it down with the other baseline checks?

Issue #453, opened 2025-05-28
Spectroscopy: Extract a spectrum from the PSF-subtracted target frame


hsergi:
**Is your feature request related to a problem? Please describe.**
Extract a spectrum from the PSF-subtracted target frame at the estimated source position. It needs LSF and wavelength solution calibration products.

Wavelength solution: #428 
LSF: #429

hsergi:
The spectrum of the source extracted at L3->L4 is a 1-D array in units of e/s/bin, with a user choice of either a simple sum or a weighted photometric extraction along the EXCAM pixel columns (cross-dispersion axis). The array wavelength bins are determined by the source position-dependent wavelength solution, which map each EXCAM array pixel to wavelength.

hsergi:
Updat: The LSF is not needed to extract the L4 spectrum. The LSF is needed to interpret, but not produce the TDA spectrum. For this reason, we shifted the LSF to the TDA stage of the data processing table.

hsergi:
One more formal note: This is part of the FDD [AAC calibrations]([AAC](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/106725541/JPL+Functional+Design+Descriptions#JPLFunctionalDesignDescriptions-SpectroscopicCalibrations)). Work can start at any time.

PR #551, opened 2025-09-16
extract spectrum from L3 image

JuergenSchreiber:
## Describe your changes
Simple boxcar spectral extraction in l3_to_l4.py around wavelength zeropoint, optionally with weighted sums.
Results in 1-D data and corresponding wavelength array including uncertainties.

## Type of change

#453

- [x] I have linted my code

### DIFF
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -333,175 +333,10 @@ def find_star(input_dataset,
     return dataset
 
 
-def crop(input_dataset, sizexy=None, centerxy=None):
-    """
-    
-    Crop the Images in a Dataset to a desired field of view. Default behavior is to 
-    crop the image to the dark hole region, centered on the pixel intersection nearest 
-    to the star location. Assumes 3D Image data is a stack of 2D data arrays, so only 
-    crops the last two indices. Currently only configured for HLC mode.
-
-    TODO: 
-        - Pad with nans if you try to crop outside the array (handle err & DQ too)
-        - Option to crop to an odd data array and center on a pixel?
-
-    Args:
-        input_dataset (corgidrp.data.Dataset): a dataset of Images (any level)
-        sizexy (int or array of int): desired frame size, if only one number is provided the 
-            desired shape is assumed to be square, otherwise xy order. If not provided, 
-            defaults to 60 for NFOV (narrow field-of-view) observations. Defaults to None.
-        centerxy (float or array of float): desired center (xy order), should be a pixel intersection (a.k.a 
-            half-integer) otherwise the function rounds to the nearest intersection. Defaults to the 
-            "STARLOCX/Y" header values.
-
-    Returns:
-        corgidrp.data.Dataset: a version of the input dataset cropped to the desired FOV.
-    """
-
-    # Copy input dataset
-    dataset = input_dataset.copy()
-
-    # Require even data shape
-    if not sizexy is None and not np.all(np.array(sizexy)%2==0):
-        raise UserWarning('Even sizexy is required.')
-       
-    # Need to loop over frames and reinit dataset because array sizes change
-    frames_out = []
-
-    for frame in dataset:
-        prihdr = frame.pri_hdr
-        exthdr = frame.ext_hdr
-        dqhdr = frame.dq_hdr
-        errhdr = frame.err_hdr
-
-        # Pick default crop size based on the size of the effective field of view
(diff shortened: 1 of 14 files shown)


corgidrp/l3_to_l4.py
```diff
@@ -975,6 +975,58 @@ def add_wavelength_map(input_dataset, disp_model, pixel_pitch_um = 13.0, ntrials
...
+                                                          xcent_round - halfwidth:xcent_round + halfwidth + 1]
+        wave_err_cutout = image.hdu_list["WAVE_ERR"].data[ycent_round - halfheight:ycent_round + halfheight + 1,
+                                                          xcent_round - halfwidth:xcent_round + halfwidth + 1]
+        err_cutout = image.err[:,ycent_round - halfheight:ycent_round + halfheight + 1,
+                                  xcent_round - halfwidth:xcent_round + halfwidth + 1]
+        bad_ind = np.where(dq_cutout > 0)
+        image_cutout[bad_ind] = np.nan
+        err_cutout[bad_ind] = np.nan
+        wave = np.mean(wave_cal_map_cutout, axis=1)
+        wave_err = np.sqrt(np.nansum(np.square(wave_err_cutout), axis=1))
```
semaphoreP, line 1010: these errors on the wavelength solution are not uncorrelated, right? If so, the errors don't add in quadrature. We may just want to take the mean of the errors.

corgidrp/l3_to_l4.py
```diff
@@ -975,6 +975,58 @@ def add_wavelength_map(input_dataset, disp_model, pixel_pitch_um = 13.0, ntrials
...
+        err = np.sqrt(np.nansum(np.square(err_cutout), axis=2))
+        dq = np.sum(dq_cutout, axis = 1)
```
semaphoreP, line 1012: I think propogating dq's is a little more complicated than this (see lines 51-54 of combine.py where it's done in combine_images()). We'll want to bitwise or to keep track of badness reasons, and also set DQ = 0 when there are some good pixels in the sum.

corgidrp/l3_to_l4.py
```diff
@@ -975,6 +975,58 @@ def add_wavelength_map(input_dataset, disp_model, pixel_pitch_um = 13.0, ntrials
...
+        else:
+            spec = np.nansum(image_cutout, axis=1)
+        image.data = spec
+        image.err = err
+        image.dq = dq
+        image.hdu_list["WAVE"].data = wave
+        image.hdu_list["WAVE_ERR"].data = wave_err
+        del(image.hdu_list["POSLOOKUP"])
+    
+    history_msg = "spectral extraction"
```
semaphoreP, line 1027: maybe a bit more information is useful. Done with half-width, half-height, and yes/no on weights.

tests/test_spec.py
```diff
@@ -584,11 +582,46 @@ def test_linespread_function():
...
 
+def test_extract_spec():
+    """
+    test the l3_to_l4.extract_spec() function, that extracts the 1D spectrum 
+    with corresponding wavelengths, error, and dq.
+    """
+    spec_dataset = l3_to_l4.extract_spec(output_dataset)
+    image = spec_dataset[0]
+    #halfwidth = 9, size: 2 * 9 + 1
+    assert np.shape(image.data) == (19,)
```
semaphoreP, line 598: based on what the input data looks like, could we do some assert to check the values of the data? that the mean of the values is typically X or something (we could set a really high tolerance so even if we're off by a factor of 2-3 it's fine). It's just a useful test to check the values in case they go haywire due to a future change.

semaphoreP (commented):
A few comments and questions

corgidrp/l3_to_l4.py
JuergenSchreiber: right, I forgot this.

corgidrp/l3_to_l4.py
semaphoreP: I think the dominant error is probably global uncertainty in the wavelength solution, so if the wavelengths are all pretty close, I think that averaging the wavelengths should produce an error in the wavelength solution comparable to the initial wavelength solution.

tests/test_spec.py
JuergenSchreiber: ok, I check the values at the maximum.

semaphoreP: Did you already add this? If so, I can't find where it is

Issue #454, opened 2025-05-28
Spectroscopy L3->L4:  PSF subtraction algorithm throughput at the source position


hsergi:
**Is your feature request related to a problem? Please describe.**
L3 -> L4: PSF subtraction algorithm throughput for a point source _at the estimated source position only_, not as a function of position (the latter is the case for the imaging mode)

**Describe the solution you'd like**
In the imaging case, in L3->L4, the PSF subtraction algorithm throughput for point sources at the detection limit as a function of separation is provided.

Might be similar as: https://github.com/roman-corgi/corgidrp/issues/238, but at the estimated source position only

semaphoreP:
Yes, this is separate from core throughput. I'm implementing it right now into PR #666

semaphoreP:
Added in #666

PR #566, opened 2025-09-24
PSF subtraction to include measured throughput at source location(s)

kjl0025:
## Describe your changes
I added a new HDU called 'KL_THRUS' which stores for each source found for each KL mode truncation choice the source distance, angle from north/up, KL throughput at source location, and FWHM.  There was no explicit test for the KL_THRU header (which stores the KL throughput at various distances), so I didn't add any tests of this new HDU to any test_*.py file, but I added an agreement check for it in l2b_to_l4_e2e.py.  

## Type of change

Issue #454 

- [ ] I have linted my code

### DIFF
--- a/corgidrp/calibrate_nonlin.py
+++ b/corgidrp/calibrate_nonlin.py
@@ -9,7 +9,10 @@ from pathlib import Path
 import matplotlib.pyplot as plt
 from scipy.interpolate import interp1d
 from statsmodels.nonparametric.smoothers_lowess import lowess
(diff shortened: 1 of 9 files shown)


semaphoreP:
@wbalmer is developing the spec mode PSF subtraction code. It will likely use a fundamentally different algorithm. It is perhaps best to work on a different task, as I imagine William will integration psf subtraction at source location with his algorithm.

semaphoreP:
@kjl0025 can you make a very small PR just for that change?

Issue #568, opened 2025-09-25
Line spread e2e shouldn't need braodband images to be passed in


semaphoreP:
**Describe the bug**
Line spread e2e shouldn't need broadband images to be passed in, but currently they do in order for the code to not crash

We probably need to modify some of the step functions to be able to handle only narrowband images as input. 

**To Reproduce**
Run the line spread function e2e and remove the input broadband data. 

Lower priority as passing in a few broadband images does not break our workflow, but would be nice to have this fixed.

PR #574, opened 2025-10-06
run e2e test with only 3D narrowband frames

JuergenSchreiber:
## Describe your changes
determine wavelength zero point can now handle datasets with only narrow band frames, raising only a warning, returning the frames as if science frames.
Also applied in e2e test spec_linespread_e2e.

## Type of change

#568

- [x] I have linted my code


semaphoreP (commented):
Just one comment, and then we are ok.

Issue #582, opened 2025-10-14
L2a-L3 e2e processing tests for spec and pol modes


juliamilton:
The following requirements in the spectroscopy and polarimetry FDDs address data level processing for spec and pol mode images:
CGI-REQT-5479 - Spectroscopy L2a→ L2b analog data processing
CGI-REQT-5480 - Spectroscopy L2a→ L2b photon counted data processing
CGI-REQT-5481 - Spectroscopy L2b → L3 data processing
CGI-REQT-5486 - Polarimetry L2b → L3 analog data processing

To verify these requirements are met in the latest DRP release, we will run simulated spectroscopy and polarimetry images through the existing e2e processing test framework and validate the resulting products.

Proposed approach:

- [ ] Update L1-> L2b test: Add VAP test logging to the existing l1_to_l2b_e2e.py test, modify its arguments to accept custom input files, and run simulated L1 spectroscopy images through it.
- [ ] Add spectroscopy L2b->L3 test: Create a new e2e test for l2b to l3 spectroscopy processing with VAP test logging and run simulated L2b spectroscopy images through it
- [ ] Add polarimetry L2b->L3 test: Create a new e2e test for l2b to l3 polarimetry processing with VAP test logging and run simulated L2b polarimetry images through it. This may be a separate implementation from the spectroscopy L2b->L3 e2e test above, due to the need to split by polarization at the L2b level (unless this can be handled by the recipe).

PR #583, opened 2025-10-15
L2a->L3 processing and e2e tests for spectroscopy data

juliamilton:
## Describe your changes
Implements recipes and modifies/ adds e2e tests to process spectroscopy images from L2a->L3. 
Includes VAP test logging in the e2e tests. 
Updates e2e tests to accept optional arguments for input and calibration data, so that files to be processed can be specified at runtime rather than hard-coded in the test. To specify simulated spec images as input data, use a command like the following:

 pytest --which e2e \
  --e2edata_path /Users/jmilton/Documents/CGI/E2E_Test_Data2 \
  --e2eoutput_path tests/e2e_tests/ \
  --input_datadir /Users/jmilton/Documents/CGI/E2E_Test_Data2/SPEC_refstar_slitless_prism/L1/ \
  tests/e2e_tests/l1_to_l2b_e2e.py

## Type of change

Addresses the spectroscopy data level processing requirements referenced in #582 

- [ ] I have linted my code

### DIFF
--- a/corgidrp/walker.py
+++ b/corgidrp/walker.py
@@ -350,6 +350,8 @@ def guess_template(dataset):
             recipe_filename = 'l2b_to_corethroughput.json'
         elif image.pri_hdr['VISTYPE'] == "CGIVST_CAL_POL_SETUP":
             recipe_filename = "l2b_to_polcal.json"
+        elif image.ext_hdr['DPAMNAME'] == 'POL0' or image.ext_hdr['DPAMNAME'] == 'POL45':
+            recipe_filename = "l2b_to_l3_pol.json"
         else:
             recipe_filename = "l2b_to_l3.json"
     # L3 -> L4 data processing
--- a/tests/e2e_tests/l1_to_l3_pol_e2e.py
+++ b/tests/e2e_tests/l1_to_l3_pol_e2e.py
@@ -277,7 +277,7 @@ def run_l1_to_l3_e2e_test(l1_datadir, l3_outputdir, processed_cal_path, logger):
     
     # Step 2: L2b -> L3 
     logger.info('Step 2: Running L2b to L3 polarimetry recipe...')
-    walker.walk_corgidrp(l2b_filelist, "", l3_outputdir, template="l2b_to_l3_pol.json")
+    walker.walk_corgidrp(l2b_filelist, "", l3_outputdir)
     logger.info('L2b to L3 complete.')
     logger.info('')
     
--- a/tests/e2e_tests/l1_to_l3_spec_e2e.py
+++ b/tests/e2e_tests/l1_to_l3_spec_e2e.py
@@ -305,7 +305,7 @@ def run_l1_to_l3_e2e_test(l1_datadir, l3_outputdir, processed_cal_path, logger):
     
     # Step 1: L1 -> L2a (generic processing)
     logger.info('Step 1: Running L1 to L2a recipe...')
-    walker.walk_corgidrp(input_data_filelist, "", l3_outputdir, template="l1_to_l2a_basic.json")
+    walker.walk_corgidrp(input_data_filelist, "", l3_outputdir)
     
     # Find the L2a output files
     l2a_files = [f for f in os.listdir(l3_outputdir) if f.endswith('_l2a.fits')]


juliamilton:
Hi @semaphoreP -
(1) No, according to their FDD, spec l2a->l2b does not apply flat-field division whereas regular imaging l2a->l2b does.
(2) The issue I see with that is since there isn’t a third argument for cals_dir, the e2e tests currently rely on hardcoded paths to locate the input and calibration data subfolders under e2edata_path. Having an optional cals_dir argument would make it much easier to point to input folders and calibration folders for different runs (eg. regular imaging vs spec vs TVAC) without editing the test each time.

If adding another argument is really not desirable, an alternative would be to standardize the directory structure so that each dataset now and in the future contains L1/ and Cals/ under {e2edata_path}, and build the relative paths from there.

tests/e2e_tests/l2b_to_l3_e2e.py
```diff
@@ -0,0 +1,357 @@
...
+        e2eoutput_path (str): Output directory path for results and logs
+        input_datadir (str or None): Custom input data directory
+        cals_dir (str or None): Custom calibration directory
+    """
+    # Set up output directory and logging
+    global logger
+    
+    # Use custom paths if provided, otherwise fall back to defaults from e2edata_path
+    if input_datadir is None:
+        l2b_datadir = os.path.join(e2edata_path, "SPEC_targetstar_slit_prism", "L2b")
```
semaphoreP: how are these L2b data produced? Did you produce them using the l1_to_l2b e2e test? If so, I think we should chain this test to use the on-the-fly output of the l1_to_l2b_e2e test to be used as the input here.

juliamilton:
> A few comments. Mainly that the use of this manual input_datadir keyword doesn't allow us to automatically run spec/pol e2e tests when we run all the e2e tests at the same time.

That was intentional. If the preferred approach is to keep all the e2e tests fully static with hardcoded paths, I can change the implementation.

semaphoreP:
Yes, I prefer static with hardcoded paths. It's easier to deploy e2e tests on an automated testing system with everything static.

JuergenSchreiber:
@juliamilton @semaphoreP I have the same problem when starting spec VAP e2e psfsub testing from L3.  Should I chain the e2e tests or should I use the results of Julia e2e tests that will be stored on sharepoint? I think it is quite an effort to always run the files starting from L1.

semaphoreP:
@JuergenSchreiber For this release, do what's easiest (so if that's starting from Julia's L2b/L3 outputs, that's fine). I think in the future, I'd like to move everything to start from L1 (to avoid tests depending on one another, and forcing us to track that).

JuergenSchreiber:
@juliamilton can you please ping me, when your e2e tests are ready, I would like to use them to generate L3 spec simulation files to run l3_to_l4 e2e tests. Or can I already use them?

semaphoreP:
Julia just updated the code/sharepoint data late yesterday to fix a similar bug I ran into yesterday. Are you on the newest version of the code/sharepoint?

corgidrp/walker.py
semaphoreP: can we add `elif image.ext_hdr['DPAMNAME'] == 'POL0' or image.ext_hdr['DPAMNAME'] == 'POL45':` use the l2b_to_l3_pol.json recipe? So that we can automatically identify the data using the walker.

tests/e2e_tests/l1_to_l3_pol_e2e.py
```diff
@@ -0,0 +1,472 @@
...
+    walker.walk_corgidrp(input_data_filelist, "", l3_outputdir, template="l1_to_l2b.json")
+    
+    l2b_files = [f for f in os.listdir(l3_outputdir) if f.endswith('_l2b.fits')]
+    l2b_filelist = [os.path.join(l3_outputdir, f) for f in l2b_files]
+    logger.info(f'L1 to L2b complete. Generated {len(l2b_filelist)} L2b files.')
+    logger.info('')
+    
+    # Step 2: L2b -> L3 
+    logger.info('Step 2: Running L2b to L3 polarimetry recipe...')
+    walker.walk_corgidrp(l2b_filelist, "", l3_outputdir, template="l2b_to_l3_pol.json")
```
semaphoreP: if we add auto identification of pol recipe into the walker, we don't need to specify the template explicitly here.

tests/e2e_tests/l1_to_l3_spec_e2e.py
```diff
@@ -0,0 +1,574 @@
...
+    # ================================================================================
+    # (3) Run Processing Pipeline
+    # ================================================================================
+    logger.info('='*80)
+    logger.info('Running L1 -> L2a -> L2b -> L3 spectroscopy data processing pipeline')
+    logger.info('='*80)
+    
+    # Step 1: L1 -> L2a (generic processing)
+    logger.info('Step 1: Running L1 to L2a recipe...')
+    walker.walk_corgidrp(input_data_filelist, "", l3_outputdir, template="l1_to_l2a_basic.json")
```
semaphoreP: I think this gets auto-detected, so we shouldn't need to specify the template.

corgidrp/walker.py
juliamilton: updated this and the below comments!

juliamilton:
> Julia just updated the code/sharepoint data late yesterday to fix a similar bug I ran into yesterday. Are you on the newest version of the code/sharepoint?

@JuergenSchreiber Yes, this issue is with some of the original spectroscopy files having incorrect flags for photon counting. If you use the most updated files in SPEC_targetstar_slit_prism it should be fine, but if you need to run other simulated spectroscopy images through the pipeline let me know because they'll need to be modified first.

semaphoreP (approved):

JuergenSchreiber:
> > Julia just updated the code/sharepoint data late yesterday to fix a similar bug I ran into yesterday. Are you on the newest version of the code/sharepoint?
> 
> @JuergenSchreiber Yes, this issue is with some of the original spectroscopy files having incorrect flags for photon counting. If you use the most updated files in SPEC_targetstar_slit_prism it should be fine, but if you need to run other simulated spectroscopy images through the pipeline let me know because they'll need to be modified first.

Thank you, Yes, I would also need the SPEC_refstar_slit_prism and SPEC_refstar_satspot L1 data.

Issue #452, opened 2025-05-28
Spectroscopy L3->L4: PSF subtraction


hsergi:
**Is your feature request related to a problem? Please describe.**
In L3 -> L4, it is necessary to implement PSF subtraction for the case of spectroscopy. For instance, RDI. 

**Describe the solution you'd like**
For spectroscopic RDI the estimation of the target speckle image requires allowing for a column-dependent rescaling of the reference speckle image to correct for the SED ratio between the target and reference stars.

Align PSF-subtracted images and corresponding per-frame bad pixel maps; **do no rotate to North Up / East Left.**

We should record the estimated speckle images that were applied to the L3 target frames for spectroscopic RDI.

This may be an extension of https://github.com/roman-corgi/corgidrp/issues/8

hsergi:
Similarly for this other task: "align images and **per-frame bad pixel maps**". The per-frame bad pixel maps will not be part of the products delivered in release 3.1, though we want this functionality to exist in the DRP.

From Jason: "This is already implemented, using the existing code in combine.py. No need to do implement anything if we use existing code for combination." Double check with Jason if there's any UT or similar to write, or if everything nedded exists.

PR #584, opened 2025-10-16
Spec psf subtraction 

kjl0025:
## Describe your changes
Added step function for PSF subtraction (RDI) for spec mode.

## Type of change

#452

- [ ] I have linted my code

### DIFF
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -21,6 +21,8 @@ from corgidrp.spec import compute_psf_centroid, create_wave_cal, read_cent_wave,
 from corgidrp import pol
 from corgidrp import fluxcal
 from corgidrp.combine import combine_subexposures
+from astropy.io.fits.verify import VerifyWarning
+from astropy.wcs import FITSFixedWarning
 
 
 def replace_bad_pixels(input_dataset,kernelsize=3,dq_thresh=1):
@@ -597,7 +599,7 @@ def northup(input_dataset,use_wcs=True,rot_center='im_center'):
     With use_wcs=True it uses WCS infomation to calculate the north position angle, or use just 'ROLL' header keyword if use_wcs is False (not recommended).
   
     Args:
-        input_dataset (corgidrp.data.Dataset): a dataset of Images (L3-level)
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (L3-level) - now handles pol datasets shapes
         use_wcs: if you want to use WCS to correct the north position angle, set True (default). 
 	    rot_center: 'im_center', 'starloc', or manual coordinate (x,y). 'im_center' uses the center of the image. 'starloc' refers to 'STARLOCX' and 'STARLOCY' in the header. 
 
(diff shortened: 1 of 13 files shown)


corgidrp/l3_to_l4.py
```diff
@@ -1078,7 +1079,317 @@ def subtract_stellar_polarization(input_dataset, system_mueller_matrix_cal, nd_m
...
+        ref_col_mean = np.mean(shifted_ref,axis=0)
+        ref_col_mean[ref_col_mean==0] = 1 # prevent div by 0
+        scale = np.mean(frame.data,axis=0)/ref_col_mean
+        shifted_scaled_ref = shifted_ref*scale
+
+        shifted_refdq = np.roll(mean_ref.dq, (shift[0], shift[1]), axis=(0,1))
+        # at this point in the pipeline, the err is mainly shot noise, so multiplying the err is appropriate
+        # shifting may throw off err at the edges of the frame, but those pixels aren't used anyways
+        shifted_scaled_referr = np.roll(mean_ref.err[0], (shift[0], shift[1]), axis=(0,1))*scale
+        # to prevent affecting future steps with error propagated forward from pixels here we don't use:
```
semaphoreP: I don't understand the desire to set the rest of the error map to 0's. Does it do something bad if we don't do this step? in particular, the exact choice of crop region seem arbitrary and I'd rather things be uniform.

corgidrp/l3_to_l4.py
```diff
@@ -1078,7 +1079,317 @@ def subtract_stellar_polarization(input_dataset, system_mueller_matrix_cal, nd_m
...
+        frame.data -= shifted_scaled_ref
+        # update the dq and err arrays
+        frame.dq = np.bitwise_or.reduce([frame.dq, shifted_refdq], axis=0)
+        frame.add_error_term(shifted_scaled_referr, 'spec ref image err after alignment and matching spec image waveband scale')
+        all_data.append(frame.data)
+        all_dq.append(frame.dq)
+        all_err.append(frame.err)
+        image_list.append(frame)
+
+    out_dataset = data.Dataset(image_list+[mean_ref])
```
semaphoreP: i don't think we should be saving the mean reference, unless this is a requirement that I missed? It may otherwise confuse users.

corgidrp/l3_to_l4.py
```diff
@@ -1078,7 +1079,317 @@ def subtract_stellar_polarization(input_dataset, system_mueller_matrix_cal, nd_m
...
+    history_msg = f'RDI PSF subtraction applied using averaged reference image.  Last Image is shifted, scaled mean ref frame.'
```
semaphoreP: It would be useful in the history to state with filenames were used to build the reference image.

corgidrp/l3_to_l4.py
kjl0025: Okay, I'll take that out.  Wasn't sure if it could be needed as a reference later.

tests/test_spec.py
```diff
@@ -895,6 +895,60 @@ def test_linespread_function():
...
+    # now the PSF subtraction
+    ref_filepath = os.path.join("tests", "test_data", "spectroscopy", "sim_rdi_L1", "spec_sim_rdi_reference_L1.fits")
+    sci_filepath = os.path.join("tests", "test_data", "spectroscopy", "sim_rdi_L1", "spec_sim_rdi_target_L1.fits")
+    input_dset = Dataset([sci_filepath, ref_filepath]) 
+    input_dset[0].pri_hdr['PSFREF'] = 0
+    input_dset[1].pri_hdr['PSFREF'] = 1
+    for img in input_dset:
+        img.data = img.data.astype(float)
+        img.dq = np.zeros_like(img.data, dtype=int)
+        # roughly the center of the imaged area
```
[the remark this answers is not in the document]
kjl0025: Yes, I did it this way before I dug a little more into what's actually going on.  I was planning to change this to match up according to wavelength zeropoint.

semaphoreP (commented):
the changes you made for my comments look good, but now the e2e changes have caused some e2e tests to fail on my end. can you investigate?

semaphoreP (approved):

Issue #527, opened 2025-09-09
Pol Step function: Align all the frames and sub-frames relative to each other


ramya-anche:
Align all individual sub arrays relative to each other in x-y direction. (to match the pixel location)
Note: we are not aligning the roll/rotation here, only x-y pixel shift of the sub arrays via cross-correlation.
Using the satellite spots estimate the star location for each slice, update the STARLOCX and STARLOCY keywords. They should have the same value for both slices now that we've aligned them.

PR #586, opened 2025-10-16
Pol Step function: Align all the frames and sub-frames relative to each other 

helloWorlds0:
## Type of change

#527

- [x] I have linted my code

### DIFF
--- a/tests/test_find_star.py
+++ b/tests/test_find_star.py
@@ -57,7 +57,7 @@ def test_find_star_offset():
 
         measured_x, measured_y = (dataset_with_center.frames[0].ext_hdr['STARLOCX'],
                                 dataset_with_center.frames[0].ext_hdr['STARLOCY'])
-        print(measured_x, measured_y)
+
         assert np.isclose(injected_position[0], measured_x, atol=0.1), \
             f"{mode}. Expected {injected_position[0]}, got {measured_x}"
         assert np.isclose(injected_position[1], measured_y, atol=0.1), \
@@ -123,7 +123,7 @@ def test_find_star_polarimetry():
     # Set the star center position for injection of satellite spots
 
     # Add small offset and rotation in the injected data
-    injected_position = [(2, - 1), (2, - 1)]
+    injected_position = [(2, 1), (-2, - 1)]
 
     satellite_spot_angle_offset = 3
     guess_angle_offset = 0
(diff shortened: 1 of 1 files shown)


corgidrp/mocks.py
```diff
@@ -4435,6 +4435,145 @@ def create_mock_l2b_polarimetric_image(image_center=(512, 512), dpamname='POL0',
...
+        corgidrp.data.Image: The simulated L2b polarimetric image containing satellite spots.
+    """
+
+    # Create polarimetric image
+    # Adapted from create_mock_l2b_polarimetric_image
+    assert dpamname in ['POL0', 'POL45'], \
+    "Invalid prism selected, must be 'POL0' or 'POL45'"
+    # create initial frame
+    image_data = np.random.normal(loc=0, scale=bg_sigma, size=image_shape) + bg_offset
+    image_separation_arcsec = 7.5
```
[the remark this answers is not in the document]
maxwellmb: would you actually mind changing it to a kwarg in create_mock_l2b_polarimetric_image() as well?

tests/test_find_star.py
```diff
@@ -113,6 +114,88 @@ def test_overwrite_parameters():
...
+            thetaOffsetGuess=thetaOffsetGuess)
+
+        measured_x, measured_y = (dataset_with_center.frames[0].ext_hdr['STARLOCX'],
+                                dataset_with_center.frames[0].ext_hdr['STARLOCY'])
+
+        assert np.isclose(dataset_with_center.frames[0].data[0].shape[0]//2 + injected_position[0][0], measured_x, atol=0.1), \
+            f"{mode}. Expected {injected_position[0][1]}, got {measured_x}"
+        assert np.isclose(dataset_with_center.frames[0].data[0].shape[1]//2 + injected_position[0][1], measured_y, atol=0.1), \
+            f"{mode}. Expected {injected_position[0][0]}, got {measured_y}"
+
```
maxwellmb: I think we need some kind of test that the alignment between the two polarization states works as well. For example with injected_position not being equal.

corgidrp/l3_to_l4.py
```diff
@@ -874,6 +917,46 @@ def extract_spec(input_dataset, halfwidth = 2, halfheight = 9, apply_weights = F
...
+
+        shift_value = (new_star_center_y - pol45_dataset.frames[0].ext_hdr['STARLOCY'] , new_star_center_x - pol45_dataset.frames[0].ext_hdr['STARLOCX'])
+        for frame in processed_dataset:
+            if frame.ext_hdr['DPAMNAME'] == 'POL45' :
+                frame.data[0] = shift( frame.data[0], shift_value)
+                frame.data[1] = shift( frame.data[1], shift_value)
+                frame.ext_hdr['STARLOCX'] = new_star_center_x
+                frame.ext_hdr['STARLOCY'] = new_star_center_y
+
+    history_msgs = "Image centered on star location."
```
maxwellmb: Could we and the location to this message (I know it's redundant with STARLOCX, but it helps track history to add details here too.

corgidrp/l3_to_l4.py
```diff
@@ -874,6 +917,46 @@ def extract_spec(input_dataset, halfwidth = 2, halfheight = 9, apply_weights = F
...
+
+    Returns:
+        corgidrp.data.Dataset: L3 dataset where all the images are registered to the same pixel
+
+
+    """
+    processed_dataset = input_dataset.copy()
+
+    split_datasets, unique_vals = processed_dataset.split_dataset(exthdr_keywords=['DPAMNAME'])
+    # If two values, align POL45 on POL0
```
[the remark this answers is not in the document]
helloWorlds0: That can work, but then we take the first image as the one we align every other on, even if it's not POL0 ? (for instance, if we have POL0 and POL45 but the first image is POL45) ? 

> I could imagine possibly running different sets of pol data that have different satellite spots

In this situation would it be several datasets (that are all, for instance, pol0 but with different satellite spots) and have been put through find_star (so each of them has only one starloc) and ONLY then concatenated into one dataset?  (Because if they're put together before find_star then the star if found on the median image of all images with satellite spots)

tests/test_find_star.py
maxwellmb: No, I meant more that injected position is a list of two not equal values right now, but I think we want to test when they're not equal too, right? My understanding is that right now when they're equal then there's. no shift between the two polarization states (e.g. 0 and 90), but we'll want to make sure that works too.

helloWorlds0:
> One last comment

Yes, I see now that for some reason I wasn't pushing the latest changes to test_find_star! Fixed now

maxwellmb (approved):

Issue #512, opened 2025-08-28
ndf_cal and bpm_cal products saved in corgidrp.default_cal_dir in their unit tests


kjl0025:
**Describe the bug**
ndf_cal and bpm_cal products are saved in corgidrp.default_cal_dir in their unit tests, but they should be saved to some folder in the tests folder, or the created files should be deleted from corgidrp.default_cal_dir at the end of the unit tests.

**To Reproduce**
Run test_badpixelmap.py and test_nd_filter_calibration.py.

**Expected behavior**
No calibration products which are not added to corgidrp.default_cal_dir by default in caldb.py should be present.

**corgidrp version:**
2.2

**Additional context**

maxwellmb:
fixed with PR #590

PR #590, opened 2025-10-17
#512 ndf_cal and bpm_cal saving fix

juliamilton:
## Describe your changes
ndf_cal and bpm_cal products produced by unit tests are no longer saved in .corgidrp/default_calibs. They are now saved in a dedicated tests/testcalib folder following the logic of other calibration products produced by unit tests.

This should resolve some file permissions errors that we see when trying to run this code in docker containers on yzma/ other configurations. 

## Type of change

#512 

- [ ] I have linted my code

### DIFF
--- a/tests/e2e_tests/spec_prism_disp_e2e.py
+++ b/tests/e2e_tests/spec_prism_disp_e2e.py
@@ -232,13 +232,12 @@ def test_run_end_to_end(e2edata_path, e2eoutput_path):
         if os.path.isfile(file_path):
             os.remove(file_path)
     
-    # Initialize a connection to the calibration database
+
     tmp_caldb_csv = os.path.join(corgidrp.config_folder, 'tmp_e2e_test_caldb.csv')
     corgidrp.caldb_filepath = tmp_caldb_csv
-    # remove any existing caldb file so that CalDB() creates a new one
+
     if os.path.exists(corgidrp.caldb_filepath):
         os.remove(tmp_caldb_csv)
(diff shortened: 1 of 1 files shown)


tests/e2e_tests/spec_prism_disp_e2e.py
```diff
@@ -230,6 +232,14 @@ def test_run_end_to_end(e2edata_path, e2eoutput_path):
...
             os.remove(file_path)
     
+    # Initialize a connection to the calibration database
+    tmp_caldb_csv = os.path.join(corgidrp.config_folder, 'tmp_e2e_test_caldb.csv')
+    corgidrp.caldb_filepath = tmp_caldb_csv
+    # remove any existing caldb file so that CalDB() creates a new one
+    if os.path.exists(corgidrp.caldb_filepath):
+        os.remove(tmp_caldb_csv)
+    this_caldb = caldb.CalDB() # connection to cal DB
+
```
maxwellmb: Are we going to have to do this for all the e2e tests, or for some reason just this one?

juliamilton: Some e2e tests do use this already if they need to scan for existing calibrations (like astrom_e2e.py and others), and some don't. I just copied this block of code from an existing test but I think this file is fine without it, so I've now removed it. Idk why this block of code isn't showing as outdated here!

maxwellmb: (maybe it doesn't show as outdated because I just added a single comment, not a full review).

PR #597, opened 2025-10-21
pol-flatfield

ramya-anche:
## Type of change

- [x] I have linted my code

### DIFF
--- a/corgidrp/flat.py
+++ b/corgidrp/flat.py
@@ -1,19 +1,21 @@
 # Place to put flatfield-related utility functions
-
+import warnings
 import numpy as np
 from scipy import interpolate
 from scipy.ndimage import median_filter
 from scipy.ndimage import gaussian_filter as gauss
 from scipy import ndimage
 from scipy.signal import convolve2d
+from scipy.ndimage import zoom
 import astropy.io.fits as fits
 from astropy.convolution import convolve_fft
+from astropy.utils.exceptions import AstropyUserWarning
 import photutils.centroids as centr
 from photutils.aperture import CircularAperture
-import corgidrp.mocks as mocks
 from photutils.detection import DAOStarFinder
 import corgidrp.data as data
-from scipy.ndimage import zoom
+import corgidrp.mocks as mocks
+
 
 def create_flatfield(flat_dataset):
 
@@ -223,7 +225,9 @@ def combine_flatfield_rasters(resi_images_dataset,cent=None,planet=None,band=Non
     nx = np.arange(0,full_residuals_resel.shape[1])
     ny = np.arange(0,full_residuals_resel.shape[0])
     nxx,nyy = np.meshgrid(ny,nx)
-    cent_n=centr.centroid_com(p_flat)
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", category=AstropyUserWarning) 
+        cent_n=centr.centroid_com(p_flat)
     nrr = np.sqrt((nxx-cent_n[0])**2 + (nyy-cent_n[1])**2)
     p_flat[nrr>n_pix//2]= 1
     p_flat_err[nrr>n_pix//2]=0


corgidrp/flat.py
```diff
@@ -73,37 +73,93 @@ def raster_kernel(width, image, hard=True):
...
+        current_filter_smooth = matched_filters_smooth[n_images]
+        #calculate the residual
+        residual = current_image / current_filter_smooth
+        matched_residuals.append(residual)
+        
+        #error propagation for division (Z = X/Y) ---
+        # Sigma_Z = |Z| * sqrt( (Sigma_X/X)^2 + (Sigma_Y/Y)^2 )
+        
+        #sigma_x: Image error (Assumed Poisson noise: Sigma_X = sqrt(X))
+        sigma_image = np.sqrt(np.maximum(current_image, 1e-6))
```
semaphoreP: instead of just assuming poisson noise, why don't we just pass in the err maps of each frame as one of the inputs to the function?

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+        Args:
+            dataset (corgidrp.data.Dataset): a dataset of image frames that are raster scanned (L2a-level)
+            nr (int): planet radius
+            pfov_size (int): size of the image created for the polarization variation, typically ~ 2 x nr
+            image_center_x (int): x coordinate of the center pixel of the final image with two orthogonal pol components
+            image_center_y (int): y coordinate of the center pixel of the final image with two orthogonal pol components
+            separation_diameter_arcsec (float): separation between the two orthogonal pol components in arcsecs
+            alignment_angle_WP1 (int): wollaston prism angle for Pol0 - 0
+            alignment_angle_WP2 (int): wollaston prism angle for Pol45- 45
+            planet (str): neptune or uranus
```
semaphoreP: should the code throw an error if the user does not supply neptune/uranus and the band 1/4, or should the code automatically determine neptune/uranus and band 1/4 using the header keywords in the dataset object?

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+
+                                     
+def create_spatial_pol(dataset,nr=60,pfov_size=174,image_center_x=512,image_center_y=512,separation_diameter_arcsec=7.5,alignment_angle_WP1=0,alignment_angle_WP2=45,planet=None,band=None,prism='POL0'):
+    """Turns a dataset of neptune or uranus images with single planet images into the images observed through the wollaston prisms also incorporates the spatial variation of polarization on the 
+        surface of the planet
+
+    
+        Args:
+            dataset (corgidrp.data.Dataset): a dataset of image frames that are raster scanned (L2a-level)
+            nr (int): planet radius
```
semaphoreP: should nr default to None as the value? since you have code further down in lines 73-78 to guess the nr based on the target name.

corgidrp/flat.py
```diff
@@ -73,37 +73,93 @@ def raster_kernel(width, image, hard=True):
 
     return kernel
 
+# def flatfield_residuals(images, N=None):
```
semaphoreP: delete the commented out code

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+
+
+    if prism == 'POL0':
+    #place image according to specified angle
+        angle_rad = (alignment_angle_WP1 * np.pi) / 180
+    else:
+        angle_rad = (alignment_angle_WP2 * np.pi) / 180
+     
+    
+    displacement_x = int(round((separation_diameter_arcsec * np.cos(angle_rad)) / (2 * 0.0218)))
```
semaphoreP: we're going to be measuring the real platescale on-sky with the Astrometric Calibration. Should we be passing in the astrometric calibration and getting the platesclae from it, instead of hard coding it to an approximate number?

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+    y_left = y + start_left[1]
+    x_right = x + start_right[0]
+    y_right = y + start_right[1]
+    
+    WP_pol=WP_image.copy() 
+    WP_pol[start_left[1]:start_left[1]+pfov, start_left[0]:start_left[0]+pfov]=I_1[yc_init - (pfov//2):yc_init + (pfov//2),xc_init - (pfov//2):xc_init + (pfov//2)]* 0.5 * (1+(2*r_xy)) 
+    WP_pol[start_right[1]:start_right[1]+pfov, start_right[0]:start_right[0]+pfov]=I_2[yc_init - (pfov//2):yc_init + (pfov//2),xc_init - (pfov//2):xc_init + (pfov//2)]* 0.5 * (1-(2*r_xy))
+
+
+    prihdr, exthdr = mocks.create_default_L1_headers()
```
semaphoreP: Shouldn't we use the existing headers from the input dataset instead of making mock ones?

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+    frame = data.Image(WP_pol, pri_hdr=prihdr, ext_hdr=exthdr)
+    frame.pri_hdr.set('TARGET', planet)
```
semaphoreP: do we need to be overriding this?

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+from photutils.aperture import CircularAperture
+import corgidrp.mocks as mocks
+import corgidrp.data as data
+from astropy.stats import sigma_clipped_stats
+from photutils.detection import DAOStarFinder
+import corgidrp.flat as flat
+from skimage.transform import downscale_local_mean
+
+                                     
+def create_spatial_pol(dataset,nr=60,pfov_size=174,image_center_x=512,image_center_y=512,separation_diameter_arcsec=7.5,alignment_angle_WP1=0,alignment_angle_WP2=45,planet=None,band=None,prism='POL0'):
```
semaphoreP: what is the purpose of this function compared to the one that exists in mocks.py already? should we just delete it? if so, ignore all my comments on thsi function.

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+            rad_mask = 1.25
+         elif band[0] == "4":
+            rad_mask = 1.75
+
+    smooth_images=[]; raster_images_cent_pol1=[]; cent_pol1=[]; act_cents_1=[]; act_cents_2=[];raster_images_cent_pol2=[];cent_pol2=[];
+    for j in range(len(dataset)):
+        planet_image=dataset[j].data
+        prihdr=dataset[j].pri_hdr
+        exthdr=dataset[j].ext_hdr
+        planet_image_downsampled = downscale_local_mean(planet_image, (10,10))
```
semaphoreP: why do we need to do this? Alternatively, you could use scipy.ndimage.gaussian_filter or median_filter to smooth the image. That way we 1) don't have to worry about downsmapling and upsampling coordinates and 2) don't need to add a new package to the requirements just for this step.

Additioanlly, the amount we smooth/downsample should be a optional keyword for the function.

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+    raster_com_pol2=combine_pol_flatfield_rasters(resi_images_pol2,resi_images_pol2_err,planet=planet,band=band,cent=cent_pol2, im_size=im_size, rad_mask=rad_mask,planet_rad=planet_rad,n_pix=n_pix, n_pad=n_pad,image_center_x=512,image_center_y=512)
+    
+    if prism == 'POL0':
+    #place image according to specified angle
+        angle_rad = (alignment_angle_WP1 * np.pi) / 180
+    else:
+        angle_rad = (alignment_angle_WP2 * np.pi) / 180
+     
+    
+    displacement_x = int(round((separation_diameter_arcsec * np.cos(angle_rad)) / (2 * 0.0218)))
```
semaphoreP: Since the real platescale of the data won't probably be exaclty 0.0218, should we be passing in the Astrometric Calibration as an input to use the real platescale we measure?

corgidrp/mocks.py
```diff
@@ -4375,6 +4375,115 @@ def get_formatted_filename(dt, visitid):
...
+    else:
+        angle_rad = (alignment_angle_WP2 * np.pi) / 180
+     
+    
+    displacement_x = int(round((separation_diameter_arcsec * np.cos(angle_rad)) / (2 * 0.0218)))
+    displacement_y = int(round((separation_diameter_arcsec * np.sin(angle_rad)) / (2 * 0.0218)))
+    center_left = (image_center_x - displacement_x, image_center_y + displacement_y)
+    center_right = (image_center_x + displacement_x, image_center_y - displacement_y)
+
+    WP_image=np.random.poisson(lam=0.199, size=(1024, 1024)).astype(np.float64)
```
semaphoreP: Is this the true flat field? if so, should we be returning it? so we can use it to check if the derived flat field matches the true flat field?

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+    		data.Dataset: dataset of uranus or neptune with spatial variation of polarization corresponding to specific wollaston prism
+    		
+	"""
+    pfov = pfov_size
+    polar_fov = np.ones((pfov,pfov))
+
+     # default to unvignetted polarimetry FOV diameter of 3.8"
+    radius_arcsec = 3.8 / 2
+    # convert to pixel: 0.0218" = 1 pixel
+    radius_pix = int(round(radius_arcsec / 0.0218))
```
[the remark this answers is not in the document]
ramya-anche: I don't think so. I used this part of code from one of Eric's step function, I should remove this

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+    y = np.arange(0,pfov)
+    xx, yy = np.meshgrid(x,y)
+    nrr = np.sqrt((xx-(pfov//2))**2 + (yy-(pfov//2))**2)
+
+    polar_filter = np.logical_and(yy<.99*xx,yy<.99*(pfov-xx)) + np.logical_and(yy>.99*xx,yy>.99*(pfov-xx))
+    
+    for i in range(len(dataset)):
+        target=dataset[i].pri_hdr['TARGET']
+        filter=dataset[i].pri_hdr['FILTER']
+        if planet==target and band==filter:
```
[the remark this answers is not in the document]
ramya-anche: in this step, we are selecting the corresponding HST image for the planet and filter and modifying it to be a image through POL0 or POL45 before generating rasters.

corgidrp/polflat.py
ramya-anche: This is required since the input image is HST image that doesn't have L1 headers

corgidrp/polflat.py
ramya-anche: I agree, I will make it a key word

corgidrp/polflat.py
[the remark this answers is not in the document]
semaphoreP: for clarity, you should include a `break` afterwards, so you automatically break out of the for loop once you have what you need.

corgidrp/polflat.py
```diff
@@ -0,0 +1,377 @@
...
+    else:
+        angle_rad = (alignment_angle_WP2 * np.pi) / 180
+     
+    
+    displacement_x = int(round((separation_diameter_arcsec * np.cos(angle_rad)) / (2 * 0.0218)))
+    displacement_y = int(round((separation_diameter_arcsec * np.sin(angle_rad)) / (2 * 0.0218)))
+    center_left = (image_center_x - displacement_x, image_center_y + displacement_y)
+    center_right = (image_center_x + displacement_x, image_center_y - displacement_y)
+
+    WP_image=np.random.poisson(lam=0.199, size=(1024, 1024)).astype(np.float64)
```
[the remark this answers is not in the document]
semaphoreP: but why are you generating an array of random numbers to put the HST images into, rather than an array of zeros?

Issue #555, opened 2025-09-17
photon counting function to allow synthesized master dark for input


kjl0025:
**Is your feature request related to a problem? Please describe.**
Currently, the PC function only allows for a photon-counted master dark to be fed in, and such a master dark would require thousands of darks to be created.  A more RAM-efficient option is to use the synthesized master dark created from the detector noise maps.

**Describe the solution you'd like**
Adjust function to allow synthesized master dark for input.

kjl0025:
Closing this issue since @semaphoreP approved PR #613.

PR #613, opened 2025-10-30
any master dark can be used for dark-subtracting PC frames

kjl0025:
## Describe your changes
Changed PC recipes and relevant code to allow for dark subtraction using non-PC master dark.

## Type of change

#555 

- [x] I have linted my code


corgidrp/recipe_templates/l2a_to_l2b_pc.json
semaphoreP: same comments here as the other recipe

corgidrp/darks.py
```diff
@@ -249,9 +249,13 @@ def build_trad_dark(dataset, detector_params, detector_regions=None, full_frame=
     exthdr['DATATYPE'] = 'Dark'
 
     master_dark = Dark(data, prihdr, exthdr, dataset, err, dq, errhdr)
+    master_dark.ext_hdr['DRPNFILE'] = int(np.round(np.nanmean(unmasked_num)))
```
semaphoreP: can you run zz_dataformat_e2e and update the data format doc in docs/source/data_format?

corgidrp/l2a_to_l2b.py
```diff
@@ -61,24 +61,30 @@ def add_shot_noise_to_err(input_dataset, kgain, detector_params):
     return phot_noise_dataset
 
 
-def dark_subtraction(input_dataset, dark, detector_regions=None, outputdir=None):
+def dark_subtraction(input_dataset, noisemaps=None, dark=None, detector_regions=None, outputdir=None):
```
semaphoreP: why do we need to pass in both? It seems to me that we can only use one regardless.

corgidrp/photon_counting.py
```diff
@@ -150,7 +150,14 @@ def get_pc_mean(input_dataset, pc_master_dark=None, T_factor=None, pc_ecount_max
...
         
     num_bins = len(input_dataset)//bin_size 
-    
+
+    lines = []
+    for line in input_dataset[0].ext_hdr['HISTORY']:
+        lines += [line]
+    msg = 'Dark subtracted using dark'
+    if msg in lines:
+        pc_master_dark = None # dark subtraction was already done, so override any input pc_master_dark
```
semaphoreP: maybe also add a print statement to let the user know that we are skipping dark subtraction.

corgidrp/photon_counting.py
```diff
@@ -284,17 +291,18 @@ def get_pc_mean(input_dataset, pc_master_dark=None, T_factor=None, pc_ecount_max
         if pc_master_dark is not None:
             if type(pc_master_dark) is not data.Dark:
                 raise Exception('Input type for pc_master_dark must be a Dataset of corgidrp.data.Dark instances.')
+            if (pc_master_dark.ext_hdr['EXPTIME'], pc_master_dark.ext_hdr['EMGAIN_C']) != unique_vals[0][:2]:
+                raise PhotonCountException('Dark should have the same EXPTIME and EMGAIN_C as input_dataset.')
```
semaphoreP: it would be useful in the error message to state what exptime/emgain_c we got, vs what was expected.

corgidrp/photon_counting.py
```diff
@@ -284,17 +291,18 @@ def get_pc_mean(input_dataset, pc_master_dark=None, T_factor=None, pc_ecount_max
...
-            if pc_master_dark.ext_hdr['NUM_FR'] < len(sub_dataset):
-                raise PhotonCountException('Number of frames that created the photon-counted master dark must be greater than or equal to the number of illuminated frames in order for the result to be reliable.')
+            if pc_master_dark.ext_hdr['PC_STAT'] == 'photon-counted master dark':
+                if 'PCTHRESH' not in pc_master_dark.ext_hdr:
+                    raise PhotonCountException('Threshold should be stored under the header \'PCTHRESH\'.')
+                if pc_master_dark.ext_hdr['PCTHRESH'] != thresh:
+                    raise PhotonCountException('Threshold used for photon-counted master dark should match the threshold to be used for the illuminated frames.')
+                if pc_master_dark.ext_hdr['NUM_FR'] < len(sub_dataset):
+                    raise Warning('Number of frames that created the photon-counted master dark should be greater than or equal to the number of illuminated frames in order for the result to be reliable.')
```
semaphoreP: do you mean warnings.warn? 

I'm also thinking whether this should be a print statement instead of a warning.

semaphoreP (commented):
A few comments. Mainly on defining the default recipe use case. 

I'm running all the e2e tests now, and will raise further comments if needed, but wanted to get these out before I forget (I started this PR last week and forgot to finish..)

corgidrp/darks.py
```diff
@@ -758,6 +762,7 @@ def calibrate_darks_lsq(dataset, detector_params, weighting=True, detector_regio
     noise_maps = DetectorNoiseMaps(input_stack, prihdr.copy(), exthdr.copy(), dataset,
                            input_err, input_dq, err_hdr=err_hdr)
     
+    noise_maps.ext_hdr['DRPNFILE'] = int(np.round(np.sum(mean_num_good_fr)))
```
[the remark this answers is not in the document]
semaphoreP: I ran it and this seems fine.

corgidrp/recipe_templates/l2a_to_l2b_pc.json
kjl0025: My intent with the recipe adjustments was to make the default recipes able to handle any input master dark, with the photon-counted master dark trumping any other type of master dark the walker might find.  If one is not found, it goes withe synthesized master dark.  If a traditional dark is found, it also prefers that over the synthesized master dark.  My e2e adjustment tests that these recipes pick out the best dark.  Is that cool?

corgidrp/mocks.py
```diff
@@ -2746,7 +2751,7 @@ def create_photon_countable_frames(Nbrights=30, Ndarks=40, EMgain=5000, kgain=7,
         frame.ext_hdr['EXPTIME'] = exptime
         frame.ext_hdr['RN'] = 100
         frame.ext_hdr['KGAINPAR'] = kgain
-        frame.pri_hdr['PHTCNT'] = True
+        frame.pri_hdr['PHTCNT'] = 0
```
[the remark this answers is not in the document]
kjl0025: I updated the .rst file.  I just checked here (https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/173184343/L1+Draft+-+SSC+Release+3.3.2), though, and I see that the role of 0 and 1 swapped (to what makes more sense to me, 1 for True and 0 for False).

corgidrp/photon_counting.py
semaphoreP: yes, unless this is specifically important enough for a warning (maybe not)

corgidrp/recipe_templates/l2a_to_l2b_pc.json
semaphoreP: What if the traditional dark is older than the synthesized dark? Shouldn't we pick the closer-in-time one? I worry that there are other considerations here. Because of that, I don't really like this is that it is ambiguous from reading the recipe what is the important calibration file. I'd rather have a recipe that's clear in intent, and we can always adjust the recipe to be different (use dark instead of detectornoisemap) if our calibration strategy changes.

kjl0025: Sure, that makes sense.  I was expecting closer control of the calibration folder, but I see your point.  I'll change the recipes.

corgidrp/l2a_to_l2b.py
semaphoreP: Ah yeah, you're right.

semaphoreP (approved):

PR #616, opened 2025-10-30
Unit tests for replace_bad_pixels on pol data

clarissardoo:
## Describe your changes

Introduced two unit tests to replace_bad_pixels to check that it works on pol data (one with a single frame, another with two frames). Verified replace bad pixel returns occur as expected and shapes behave according to pol data expectations.

## Type of change

- [X ] I have linted my code

### DIFF
--- a/tests/test_badpixel_correction.py
+++ b/tests/test_badpixel_correction.py
@@ -293,7 +293,6 @@ def test_replace_bps_pol_4frames():
     if not cleaned_dataset.all_dq == pytest.approx(input_dataset_bad.all_dq):
         raise Exception("Output DQ array does not match input DQ array for pol uniform data.")
 
-    print("UT passed for pol data")
 
 
 if __name__ == '__main__':


tests/test_badpixel_correction.py
```diff
@@ -241,14 +239,156 @@ def test_replace_bps_nonuniform():
...
+        raise Exception("Cleaned data array does not match input data array for pol uniform data.")
+    if not cleaned_dataset.all_err == pytest.approx(input_dataset_clean.all_err):
+        raise Exception("Cleaned error array does not match input error array for pol uniform data.")
+    if not cleaned_dataset.all_dq == pytest.approx(input_dataset_bad.all_dq):
+        raise Exception("Output DQ array does not match input DQ array for pol uniform data.")
+
+    print("UT passed for pol data")
+
+
+def test_replace_bps_pol_2frames():
```
maxwellmb: if it works with 2 frames doesn't that imply that it works with one frame? If so I think we could get rid of the first test

maxwellmb (changes requested):
smallest of small: please remove the print statement at the end, then we're good to go!

maxwellmb (approved):

Issue #542, opened 2025-09-15
Spectroscopy VAP: L3->L4 processing


juliamilton:
**Reqts:** CGI-REQT-5474, CGI-REQT-5464, CGI-REQT-5482, CGI-REQT-5485
**Purpose:** Produce L4 images/spectra with valid wavelength zero-point and wavelength map; perform PSF-sub and/or non-PSF combination and extraction.

**Inputs:**
Zero-point step (5474):

- L3 FITS image of a prism-dispersed DM satellite spot observed through the FSAM slit mask and a narrowband calibration filter
- DRP-provided CSV of relative image shifts (wedge) to correct x,y offsets between filters (e.g., 3D and 3F)

Wavelength map step (5464):
- L3 FITS image with zero-point solution
- Prism dispersion scale and orientation calibration product

Combine/extract (5482 / 5485):
- For PSF-sub: N ≥ 2 data sets across target/reference and/or orientations; star/source locations from satellite spots and/or CGI telemetry
- For non-PSF: Unocculted star sequences in a given configuration

**Procedure (high-level):**
See issues for details: https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/173180475/R3.1+VAP+Planning

1. Validate inputs (levels, data type, headers, filenames).
2. Zero-point: compute WAVLEN0 using sat-spot narrowband + wedge CSV; write to header.
3. Wavelength map: compute 2-D wavelength-to-pixel map using zero-point + dispersion CAL; write as HDU5.
4. Combine/extract:
- PSF-sub path: perform RDI (allow column-dependent rescaling for target/reference SED ratio), align (do not rotate to North-up/East-left), combine (mean/median per user), extract spectrum at estimated source position; record throughput at source position.
- Non-PSF path: align (no rotation), combine to L4, extract spectrum.
5. Log checks and validate outputs (levels, data type, headers, filename) using functions in check.py.

**Expected Outputs:**
L4 spectroscopy PSF-subtracted and/ or non-PSF-subtracted image

semaphoreP:
We just have to associate a fluxcal factor cal file to the dataset. The baseline is to use the fluxcal factor from imaging mode (i.e., in the simple case, we do not worry about flux cal factor vs wavelength). We have tested that the fluxcal code we wrote works in Bands 2 and 3, so it should be good to use.

PR #622, opened 2025-11-04
Spectroscopy l3 to l4 e2e test with and without psf subtraction

JuergenSchreiber:
## Describe your changes
added 2 spectroscopy e2e VAP tests. The test without psf subtraction is done using the old sim files from Neil in test_data, the test with psf subtraction is done using the new simulation files from sharepoint.
Additionally
1) the final product of extract_spec is changed to still contain the 2D spectral and wavelength data.
2) add the excam STARLOC star position header keywords
3) A new step function to align and combine the single images in the dataset to a single output image had to be written

## Type of change

## Reference any relevant issues
#542

- [x] I have linted my code


semaphoreP:
> For second psf sub e2e test:
> 
> 1. I wonder why it runs without the satspot data.

That is curious. We might need to look into this.

> 2. extracted spectra seems to have negative values -> still some investigations to do

For the PSF subtraction e2e test, the PSF subtraction could create negative values, so that's ok.

corgidrp/recipe_templates/l3_to_l4_psfsub_spec.json
```diff
@@ -0,0 +1,36 @@
...
+        {
+            "name" : "add_wavelength_map",
+                "calibs" : {
+                    "DispersionModel" : "AUTOMATIC"
+            }
+        },
+        {
+            "name" : "spec_psf_subtraction"
+        },
+        {
```
semaphoreP: we also need step functions to:

- align PSF-subtracted images; do no rotate to North Up / East Left
- Combine all PSF-subtracted target frames (mean or median as per user preference) to produce single final image

For the second step, we can use the `combine.combine_subexposures()` out of the box. For the first step, we need a new step function written. Writing this separate step function probably makes sense as a separate PR.

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
```diff
@@ -0,0 +1,304 @@
...
+    
+    # ================================================================================
+    # (3) Run Processing Pipeline
+    # ================================================================================
+    logger.info('='*80)
+    logger.info('Running processing pipeline')
+    logger.info('='*80)
+
+    logger.info('Running e2e recipe...')
+    recipe = walk_corgidrp(
```
semaphoreP: just like for the other e2e test, we should modify the walker to guess the recipe, so we don't need to manually specify the template here. Again, can use the logic already used in walker.guess_recipe to identify spec and coronagraphic data.

corgidrp/l3_to_l4.py
semaphoreP: I was reading over the spec FDD and we need to do this a bit differently actually. We need to return both the 2-D image, the 2-D wavelength map, and the extracted spectrum in the L4 data product. I thnk the best way to do this is to keep the 2-D image as the image.data field, and make a new HDU to store the extracted spectrum. I think this just requires you to change a bit of bookkeeping. Let me know what you think.

semaphoreP (commented):
I haven't gotten your psfsub e2e test running, but here are some comments on the implementation. I can try to help you write the missing align function.

corgidrp/recipe_templates/l3_to_l4_psfsub_spec.json
```diff
@@ -0,0 +1,36 @@
...
+    "template" : true,
+    "drpconfig" : {
+        "track_individual_errors" : false
+    },
+    "inputs" : [],
+    "outputdir" : "",
+    "steps" : [
+        {
+            "name" : "determine_wave_zeropoint"
+        },
```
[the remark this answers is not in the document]
JuergenSchreiber: shouldn't we do this in l3_to_l4.find_star()? Polarimetry does it there, I guess. Otherwise I guess we need an issue for this.

corgidrp/recipe_templates/l3_to_l4_psfsub_spec.json
JuergenSchreiber: This is already done for reference frames in spec_psf_subtraction. Do you mean do it also for the resulting subtracted dataset in spec_psf_subtraction? By the way, where is this done for the "normal" image frames?

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
```diff
@@ -0,0 +1,304 @@
...
+
+    # Validate all input images
+    for i, frame in enumerate(l3_dataset):
+        frame_info = f"L3 Frame {i}"
+        
+        check_filename_convention(getattr(frame, 'filename', None), 'cgi_*_l3_.fits', frame_info, logger, data_level = 'l3_')
+        check_dimensions(frame.data, (125, 125), frame_info, logger)
+        verify_header_keywords(frame.ext_hdr, {'DPAMNAME', 'CFAMNAME', 'FSAMNAME'}, frame_info, logger)
+        verify_header_keywords(frame.ext_hdr, {'DATALVL': 'L3'}, frame_info, logger)
+        verify_header_keywords(frame.pri_hdr, {'PSFREF'}, frame_info, logger)
```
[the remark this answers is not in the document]
JuergenSchreiber: OK, I guess I should use one satspot = 1 ref frame and one satspot = 1 target frame? Still to be downloaded from sharepoint

corgidrp/recipe_templates/l3_to_l4_psfsub_spec.json
semaphoreP: For normal TTR5 imaging frames, it is done inside the PSF subtraction function

corgidrp/l3_to_l4.py
```diff
@@ -981,6 +980,41 @@ def add_wavelength_map(input_dataset, disp_model, pixel_pitch_um = 13.0, ntrials
...
+        the occulted star in degrees, with 0 degrees meaning +X and 90 degrees
+        meaning +Y.
+
+    Returns:
+        corgidrp.data.Dataset: Dataset with updated keywords recording the satellite position
+          in EXCAM pixels.
+    """
+    
+    dataset = input_dataset.copy()
+    dataset = star_pos_spec(dataset, r_lamD = r_lamD, phi_deg = phi_deg)
```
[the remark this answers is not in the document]
JuergenSchreiber: I used the FSMLOS keyword, or is there another,  maybe better, keyword that indicates the coronagraphic mode?

Issue #490, opened 2025-08-14
reduce RAM footprint for master dark for photon counting and for analog (synthesized and traditional)


kjl0025:
**Is your feature request related to a problem? Please describe.**
The DRP will be limited to ~100GB of RAM before delivering processed products.  We want to reduce the RAM footprint for photon counting, especially in the case of a photon-counted master dark, which requires ~> 1000 frames.

**Describe the solution you'd like**
-- Read in files at float32 level (the bit depth of the on-sky data).
-- Cropping frames for the photon-counting part of the pipeline if necessary? 

**Additional context**
Can we rely on additional virtual memory via swap?

PR #652, opened 2025-11-11
RAM-heavy processes alleviated

kjl0025:
## Describe your changes
RAM-heavy processes as planned in flight (making detector noise maps, traditional master darks, PC master darks, PC illuminated) are now much less intense in RAM usage.  Huge numbers of frames (like the 25k planned for the high-gain frames for making detector noise maps) should be < 100 GB of RAM now.  Also, sort_pupilimg_frames() acts on a dataset containing metadata but no data to reduce RAM for this function since in flight the input dataset may contain ~1000 frames since the corresponding CAR is for k gain, nonlin, and EM gain cal files altogether.  

Trimmed memory usage, made processes involving large stacks to be combined instead iterate a running counter, deleting each addition after it has been added in.  Also prevented the walker from loading in the whole stack for these RAM-heavy processes, and corresponding recipes were broken up and chained together.  The input dataset for these processes are broken up into smaller datasets of 200 frames each, run one at a time, saved to hard disk, and written over by the next one.  The parts of these processes requiring all the frames in the dataset were isolated into one recipe in the chain, and those functions were edited so that they accumulate a running total, adding in each next frame one at a time a removing it from RAM before adding on the next one.  (It turned out I didn't need to resort to a cropping strategy after all, despite the name of the branch.)

## Type of change

#490 

- [x] I have linted my code

### DIFF
--- a/corgidrp/walker.py
+++ b/corgidrp/walker.py
@@ -348,7 +348,7 @@ def guess_template(dataset):
             
             if is_spectroscopy:
                 if image.ext_hdr['ISPC'] in (True, 1):
-                    recipe_filename = "l2a_to_l2b_pc_spec.json"
+                    recipe_filename = ["l2a_to_l2b_pc_spec_1.json", "l2a_to_l2b_pc_spec_2.json", "l2a_to_l2b_pc_spec_3.json"] #"l2a_to_l2b_pc_spec.json"
                 else:
                     recipe_filename = "l2a_to_l2b_spec.json"
             else:
@@ -488,8 +488,6 @@ def run_recipe(recipe, save_recipe_file=True):
     if not recipe["inputs"]:
         curr_dataset = []
         ram_heavy_bool = False
-        # total_dset_length = 1 #for loop purposes later
-        # ram_increment = 1
         filelist_chunks = [0] #anything of length 1
     else:
         filelist = recipe["inputs"]
@@ -512,6 +510,11 @@ def run_recipe(recipe, save_recipe_file=True):
             for frame in curr_dataset:
                 frame.ext_hdr["RECIPE"] = json.dumps(recipe)
         # execute each pipeline step
+        print('Executing recipe: {0}'.format(recipe['name']))
+        if ram_increment_bool and len(filelist_chunks) > 1:
+            print('Processing frames in chunks of {0} frames'.format(corgidrp.chunk_size))
+        if ram_heavy_bool:
+            print('Processing frames in RAM-heavy mode (data not loaded into memory until necessary, one frame at a time)')
         for i, step in enumerate(recipe["steps"]):
             print("Walker step {0}/{1}: {2}".format(i+1, tot_steps, step["name"]))
             if step["name"].lower() == "save":
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -62,6 +62,12 @@ class Dataset():
         else:
             # list of frames
             self.frames = frames_or_filepaths
+            # if one of the input frames has data, set data_loaded to True
+            self.data_loaded = False
+            for frame in self.frames:
+                if frame.data is not None:
+                    self.data_loaded = True
+
 
         # turn lists into np.array for indexing behavior
         if isinstance(self.frames, list):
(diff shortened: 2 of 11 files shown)


corgidrp/walker.py
```diff
@@ -462,75 +460,120 @@ def run_recipe(recipe, save_recipe_file=True):
         with open(recipe_filepath, "w") as json_file:
             json.dump(recipe, json_file, indent=4)
 
+    # determine if this is a RAM-heavy recipe which needs crop-stack processing
+    #sort_pupilimg_frames included here b/c it sorts through a large number of frame (>700) b/c 
+    # EM gain cal files (sorted here and excluded, processed by SSC) are included in the visits
+    ram_heavy_bool = (recipe['name'] == 'noisemap_generation_2' or
```
semaphoreP: Instead of having this big branching logic, let's just add a new field in the recipe .json file that's like "ram_heavy", and a pipeline parameter for ram_increment that defaults to 200, but can be changed (like how track_indiviudal_errors works). We can also make an issue and address this after this release.

corgidrp/darks.py
```diff
@@ -50,61 +51,84 @@ def mean_combine(image_list, bpmap_list, err=False):
...
+            if bp.shape != s0:
+                raise TypeError('all input list elements must be the same shape')
+            pass
 
     # Add non masked elements
-    sum_im = np.zeros_like(image_list[0])
-    map_im = np.zeros_like(image_list[0], dtype=int)
-    for im_m in ims_m:
+    if isinstance(dataset_or_image_list, data.Dataset):
+        with fits.open(dataset_or_image_list[0].filepath, 'readonly') as temp_fits:
```
semaphoreP: can you just use the Image() class to load in the file? This way, we are reusing code to open/close files. You can access the data and err arrays from the Image attribute.

corgidrp/darks.py
```diff
@@ -197,28 +230,59 @@ def build_trad_dark(dataset, detector_params, detector_regions=None, full_frame=
...
-        err = fr.err[0]
-        frame[telem_rows] = np.nan
-        i0 = slice_section(frame, 'SCI', 'image', detector_regions)
-        if np.isnan(i0).any():
-            raise ValueError('telem_rows cannot be in image area.')
-        frame[telem_rows] = 0
-        frames.append(frame)
-        bpmaps.append(b1)
-        errs.append(err)
+    if dataset[0].data is not None:
```
semaphoreP: because of how many times we're reading it data in this function, I suggest we use one function to do the loading each time (just to improve code reuse and code readability). I would suggest just doing `Image(filepath)`, but open to alternatives.

corgidrp/darks.py
semaphoreP: can we add some comments to this docstring that is the dataset has no data, is function processes one file at a time, if the dataset has data, this function processes all the data at the same time.

semaphoreP (commented):
I'm waiting for the e2e tests to run locally on my computer to verify. I think all my previous comments are resolved. I have a quick follow-up question below.

corgidrp/walker.py
```diff
@@ -472,75 +470,115 @@ def run_recipe(recipe, save_recipe_file=True):
...
+    # read in data, if not doing bp map
+    if not recipe["inputs"]:
+        curr_dataset = []
+        ram_heavy_bool = False
+        filelist_chunks = [0] #anything of length 1
+    else:
+        filelist = recipe["inputs"]
+        if ram_increment_bool and not ram_heavy_bool: #ram_heavy_bool supercedes ram_increment_bool
+            # how many frames to process at a time (before getting the RAM-heaviest function in the recipe) if RAM-heavy
+            filelist_chunks = [filelist[n:n+corgidrp.chunk_size] for n in range(0, len(filelist), corgidrp.chunk_size)]
```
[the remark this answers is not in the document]
kjl0025: This should keep all the files.  The remainder gets lumped into its own chunk.  For example:
>>> filelist=np.arange(0,13)
>>> filelist_chunks = [filelist[n:n+10] for n in range(0, len(filelist), 10)] 
>>> filelist_chunks
[array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9]), array([10, 11, 12])]
>>>

semaphoreP (approved):

Issue #641, opened 2025-11-10
Polarimetry L4->TDA VAP Test 5: Extended Source (Disk) Polarized and Total Intensity Test


juliamilton:
Implement a VAP test with logging according to the requirements for L4->TDA processing, based on the Polarimetry FDD.

### **Test 5: Extended Source (Disk) Polarized and Total Intensity Test**
**Use/adapt**: `test_polarimetry.py::test_calc_pol_p_and_pa_image()`  
**Function**: `l4_to_tda.calc_pol_p_and_pa_image()`  
**Input data**: L4 Stokes cube [I, Q, U] with known (p, theta) polarization

**5.1 Data format:**
- [ ] Check/log that L4 data input complies with cgi format
- [ ] Check/log that DATALVL = L4
- [ ] Check/log that BUNIT = photoelectron/s

**5.2 Polarized and total intensity:**
Calculate and provide polarized and total intensity:
- [ ] Check/log that P computed correctly from Q, U
- [ ] Check/log that p = P/I matches input polarization fraction
- [ ] Check/log that EVPA matches input polarization angle
- [ ] Check/log that Output shape is [3, H, W]
- [ ] Check/log error propagation through sqrt and division
- [ ] Check/log handling I ≈ 0 (avoid divide by zero)
- [ ] Check/log that DQ flags propagate correctly

maxwellmb:
Closed with PR #657

Issue #638, opened 2025-11-10
Polarimetry L4->TDA VAP Test 2: Companion Polarization Properties


juliamilton:
Implement a VAP test with logging according to the requirements for L4->TDA processing, based on the Polarimetry FDD.

### **Test 2: Companion Polarization Properties**
**Use/adapt**: `test_polarimetry.py::test_calc_pol_p_and_pa_image()`  
**Function**: `l4_to_tda.calc_pol_p_and_pa_image()`  
**Input data**: L4 Stokes cube [I, Q, U, ...] with known (p, theta) polarization

**2.1 Data format:**
- [ ] Check/log that L4 data input complies with cgi format
- [x] Check/log that DATALVL = L4
- [x] Check/log that BUNIT = photoelectron/s

**2.2 Calculate and log polarization fraction and polarization position angle of the companion:**
- [ ] Check/log that polarized intensity P is computed correctly
- [ ] Check/log that fractional polarization p matches input (within error)
- [ ] Check/log that polarization position angle matches input angle (within error)
- [ ] Check/log that error propagation from Q, U is correct
- [ ] Check/log that flags propagate from I, Q, U to P, p, polarization angle

maxwellmb:
Closed with PR #657 and #667

Issue #630, opened 2025-11-07
Polarimetry L4->TDA processing VAP tests


juliamilton:
Implement VAP tests with logging according to the requirements for L4->TDA processing, based on the Polarimetry FDD. 

Specifics of each VAP test are described in the sub-issues linked to this task.

PR #657, opened 2025-11-12
pol l4_to_tda VAP test2 #638

toshiyukimizuki:
## Describe your changes
The logger option has been added for the VAP test. 
A fixed seed has been introduced to ensure stable and reproducible simulations. 
Minor changes were implemented in 

- test_polarimetry.py:test_calc_pol_p_and_pa_image(), 
- l4_to_tda.calc_pol_p_and_pa_image(), and 
- mocks.py:create_mock_stokes_image_l4(). 

In the mock function, a point source is generated with a Gaussian profile. While the underlying signal is correlated across pixels due to the PSF, the noise is generated independently for each pixel. For the unit test, the fractional polarization is evaluated by averaging over all pixels, which succeeds. Therefore, in principle, the test works for any input, whether point sources like planets or extended sources like disks, and meets the VAP test requirements.

## Type of change

#630
#638 
#641 

- [ ] I have linted my code

### DIFF
--- a/tests/test_polarimetry.py
+++ b/tests/test_polarimetry.py
@@ -242,21 +242,27 @@ def test_calc_pol_p_and_pa_image(n_sim=100, nsigma_tol=3., seed=0,
     evpa_chi_mean, evpa_chi_std = [], []
 
     for i, p, theta in zip(range(n_sim), p_input, theta_input):
-        
-        # Generate mock Stokes cube 
+
+        if i != n_sim-1:
+            I0 = 1e10
+        else:
+            # I ~ 0 scenario for a sanity check
+            I0 = 0
+            
+        # Generate mock Stokes cube
         common_kwargs = dict(
             fwhm=1e2,
-            I0=1e10,
+            I0=I0,
             p=p,
             theta_deg=theta,
             rng=rng,
-        )
+        )   
         
         # Generate mock Stokes cube
         Image_input = mocks.create_mock_stokes_image_l4(badpixel_fraction=1e-3, **common_kwargs)
     
-        Image_input_noerr = mocks.create_mock_stokes_image_l4(badpixel_fraction=0.0, add_noise=False, **common_kwargs)
-        P_input = np.sqrt(Image_input_noerr.data[1]**2. + Image_input_noerr.data[2]**2.)
+        Image_input_true = mocks.create_mock_stokes_image_l4(badpixel_fraction=0.0, add_noise=False, **common_kwargs)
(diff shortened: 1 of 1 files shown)


juliamilton:
> I don't think there's a need to really create a new test file. We could just add the new tests for the input formats to the existing tests, no?

Sure, you can add the logging and new data format tests as described above to the `test_polarimetry.py` file if that is easier.

maxwellmb:
Ok, great, then we'll keep all the current assert statements, and then @toshiyukimizuki you can add in the logging and the specific checks that Julia has specified as well.

toshiyukimizuki:
I added a test_polarimetry_vap.py file containing 
def setup_logger and
def test_calc_pol_p_and_pa_image.
This function performs the following checks with logging:
    1. Validates that the input Image is in the expected CGI format
       and that its headers (DATALVL, BUNIT) are correct.
    2. Computes the polarization products: 
       - Polarized intensity (P)
       - Fractional polarization (p)
       - Electric-vector position angle (EVPA)
    3. Validates the output images:
       - Shape is [3, H, W]
       - Data Quality (DQ) flags propagate correctly from input to output
    4. Performs statistical consistency checks using chi:
       - (measured - true) / sigma for P, p, and EVPA
       - Checks that median ~ 0 and standard deviation ~ 1 within tolerance

This test verifies entire pixels by averaging all of them, meaning it works for any input, whether point sources, such as planets, or extended sources, such as disks. Therefore, I believe that it meets the VAP test requirements for both #638 and #641. If you notice anything that I have misunderstood, please let me know by specifying the differences between #638 and #641 and the requirements for each.

juliamilton:
Thank you so much @toshiyukimizuki, this looks good to me. Would you be able to add 2 additional things to the log just so we explicitly meet the logging requirements?

1. You are already calculating P statistically, but can you please add a quick residual check (for example, compare P_map to sqrt(Q^2+U^2) and log whether that is PASS/FAIL) in test_calc_pol_p_and_pa_image so that requirement is satisfied exactly?
2. Can you add an “I ≈ 0” scenario inside test_calc_pol_p_and_pa_image, run calc_pol_p_and_pa_image on it, and then log PASS/FAIL to confirming no NaN/Inf values show up in p?

toshiyukimizuki:
Thank you very much for the comments, @juliamilton 
I revised the test as follows: 

1. Added a check to verify that P = sqrt(Q^2 + U^2) in one of the runs. 
2. In the final run (i = n_sim-1), implemented a sanity check with I = 0. This confirms that no NaN/Inf values appear in the computed P, p, or EVPA.

PR #658, opened 2025-11-12
Flux calibration step functions: spectroscopy L4->TDA

juliamilton:
## Describe your changes

- Added `convert_spec_to_flux()` in `l4_to_tda.py`, which does flux calibration on 1‑D spectroscopy spectra. Function applies slit-throughput, color correction, and flux-factor scaling directly to the SPEC/SPEC_ERR extensions in L4 files.
- Added `compute_spec_flux_ratio()` to calculate companion/host star spectra flux comparison. Function handles wavelength interpolation, and exposure-time weighting/ averaging if multiple roll angles are present. 
- Added additional unit tests of the above functions to test_fluxcal.py.

## Type of change

- [ ] I have linted my code

### DIFF
--- a/corgidrp/l4_to_tda.py
+++ b/corgidrp/l4_to_tda.py
@@ -307,7 +307,7 @@ def apply_core_throughput_correction(frame,
         fpam_fsam_cal,
         logr=logr,
     )
-    ct_value = float(np.atleast_1d(ct_values)[0])
+    ct_value = float(np.atleast_1d(ct_values).ravel()[0])
     if not np.isfinite(ct_value) or ct_value <= 0:
         raise ValueError(f"Invalid core throughput value {ct_value}.")
 


corgidrp/l4_to_tda.py
```diff
@@ -137,6 +137,267 @@ def determine_color_cor(input_dataset, ref_star, source_star):
...
+    if history_messages:
+        spec_dataset.update_after_processing_step(
+            " ".join(history_messages),
+            header_entries={"SPECUNIT": "erg/(s*cm^2*AA)", "FLUXFAC": fluxcal_factor.fluxcal_fac}
+        )
+
+    return spec_dataset
+
+
+def compute_spec_flux_ratio(host_dataset, companion_dataset, fluxcal_factor,
```
semaphoreP: I'm not sure there is always a single corresponding host star file to each companion file. It may be cleaner to just pass in a single L4 host spectrum image and a single L4 companion spectrum image.

corgidrp/l4_to_tda.py
```diff
@@ -137,6 +137,267 @@ def determine_color_cor(input_dataset, ref_star, source_star):
...
+        slit_vals = slit_per_frame[idx]
+        slit_applied = False
+        if slit_vals is not None:
+            slit_vals = np.asarray(slit_vals, dtype=float)
+            if slit_vals.shape != spec.shape:
+                raise ValueError(f"slit_transmission array shape {slit_vals.shape} must match SPEC data shape {spec.shape}.")
+            
+            # Divide by slit transmission, handling division by zero
+            nonzero_mask = slit_vals != 0
+            spec = np.divide(spec, slit_vals, out=np.full_like(spec, np.nan), where=nonzero_mask)
```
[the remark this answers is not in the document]
juliamilton: Ok, I have tried to implement that. One point of clarification is what to do with the 2D grid option for slit map that is specified in spec.slit_transmission. How do we reduce that to a single factor after interpolating x and y? The 1D option for slit map seems more straightforward.

corgidrp/l4_to_tda.py
```diff
@@ -137,6 +137,267 @@ def determine_color_cor(input_dataset, ref_star, source_star):
     return color_dataset
 
 
+def convert_spec_to_flux(input_dataset, fluxcal_factor, slit_transmission=None):
+    """
+    Flux calibrate 1-D spectroscopy spectra stored in the L4 SPEC extension.
+    The function applies COL_COR when present, propagates calibration
+    uncertainties, and records whether slit correction was applied.
+
+    Args:
```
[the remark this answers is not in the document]
juliamilton: Now have this in a different step function and have convert_spec_to_flux() check to make sure that core throughput correction has been applied first.

JuergenSchreiber:
I just looked a bit into this. You apply the FluxcalFactor of a band (which is determined by the aperture photometry of an image integrated over the spectral band) to a spectrum. This might be a first estimate of the flux, but for a spectral flux calibration (per wavelength in a band) I think we would need new cal files that can be determined from extracted spectra of calspec standards.

corgidrp/l4_to_tda.py
```diff
@@ -137,6 +137,516 @@ def determine_color_cor(input_dataset, ref_star, source_star):
...
+    else:
+        raise TypeError("slit_transmission must be None, a (slit_map, slit_x, slit_y) tuple, or a list of such tuples.")
+
+    history_messages = []
+
+    for idx, frame in enumerate(spec_dataset):
+        if 'SPEC' not in frame.hdu_list:
+            raise ValueError("Input dataset does not contain a 'SPEC' extension.")
+
+        mask_loc_x = frame.ext_hdr.get('MASKLOCX') # using MASKLOCX to check if the image is coronagraphic
```
semaphoreP: I don't think we write this in spec mode. You may want to use `FSMLOS = 1` or just skip this check altogether for now.

corgidrp/l4_to_tda.py
```diff
@@ -137,6 +137,516 @@ def determine_color_cor(input_dataset, ref_star, source_star):
...
+                                     fpam_fsam_cal,
+                                     logr=False):
+    """
+    Apply a core-throughput correction to a single L4 spectroscopy frame.
+
+    Args:
+        frame (corgidrp.data.Image): L4 spectroscopy frame containing SPEC/SPEC_ERR HDUs.
+        core_throughput_cal (corgidrp.data.CoreThroughputCalibration): calibration product
+            providing InterpolateCT().
+        coron_dataset (corgidrp.data.Dataset): dataset describing the coronagraphic
```
semaphoreP: why can't this just be the input `frame`?

corgidrp/l4_to_tda.py
```diff
@@ -137,6 +137,516 @@ def determine_color_cor(input_dataset, ref_star, source_star):
...
+        'companion_roll': companion_image.pri_hdr.get('ROLL'),
+        'exp_time': host_image.pri_hdr.get('EXP_TIME', 1.0),
+        'weighted': False,
+        'ratio_err': ratio_unc,
+    }
+
+    return flux_ratio, host_wave, metadata
+
+
+def compute_weighted_spec_flux_ratio(host_datasets, companion_datasets, fluxcal_factor,
```
semaphoreP: I think I just meant a function that given a single input Dataset, combines them together into a single spectrum (weighted based on errors, and interpolated to a common wavelength grid). The idea is that we could use this to separately run on any dataset of spectrum (Whether host star or companion, whether in flux ratio units or not), independent of the other functions above. 

My previous concern with doing everything in a single function is that I'm not sure we are guaranteed to get one host star spectrum per roll. So there may not be this 1:1 mapping.

semaphoreP (commented):
Oops, sorry, we should also fix these warnings from the new unit tests that you made:

```
tests/test_fluxcal.py::test_apply_core_throughput_correction
tests/test_fluxcal.py::test_compute_spec_flux_ratio_single_roll
tests/test_fluxcal.py::test_compute_spec_flux_ratio_weighted
  /home/runner/work/corgidrp/corgidrp/corgidrp/l4_to_tda.py:296: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    ct_value = float(np.atleast_1d(ct_values)[0])

tests/test_fluxcal.py::test_l4_companion_photometry
  /home/runner/work/corgidrp/corgidrp/corgidrp/l4_to_tda.py:778: UserWarning: There is no COL_COR keyword in the header, color correction was not done, it is set to 1
    warnings.warn("There is no COL_COR keyword in the header, color correction was not done, it is set to 1")
```

Issue #648, opened 2025-11-10
Spectroscopy L4->TDA VAP Test 5: Slit Transmission Factor


juliamilton:
Implement a VAP test with logging according to the requirements for L4->TDA spectroscopy processing, based on the Spectroscopy FDD. Test should be once each for the following types of input data: 

- Analog non-PSF-subtracted data (eg: unocculted star)
- Analog PSF-subtracted data
- Photon-counted PSF-subtracted data
- Photon-counted non-PSF-subtracted data

---------------------------------------------------------

### **Test 5: Slit Transmission Factor**
**Use/adapt**: `tests/test_spec.py::test_slit_transmission_map()` *(add if missing)*  
**Function**: `spec.slit_transmission()`  
**Input data**:
- Averaged spectra with the slit in (FSAMNAME = Slit) for each FSM position.
- Averaged spectra with the slit open (FSAMNAME = Open) for the same FSM grid.
- Target-pixel grid for interpolation

**Test:**
- [ ] Check data format and header values per L4 documentation (DATALVL=L4, DPAMNAME = PRISM*, BUNIT=photoelectron/s/bin)
- [ ] For each FSM position, compute the transmission ratio T(λ)=S_slit(λ)/S_open(λ).
- [ ] Interpolate the ratios onto the grid
- [ ] Check and log that the returned transmission map shape matches the requested grid and that all values are (0, 1]. If NaNs exist, they should match locations of masked pixels (if any).
- [ ] Check and log that interpolation handles 1-D and 2-D FSM grids.

juliamilton:
Completed in PR #461 `test_spec.py`

Issue #647, opened 2025-11-10
Spectroscopy L4->TDA VAP Test 4: Line Spread Function Characterization


juliamilton:
### **Test 4: Line Spread Function Characterization**
**Use/adapt**: `tests/test_spec.py::test_fit_line_spread_function()`  
**Function**: `spec.fit_line_spread_function()`  
**Input data**: Narrowband + prism calibration dataset with `WAVE` map and slit metadata

**Test:**
- [ ] Check and log that the LSF fit uses the correct narrowband subset (`CFAMNAME=` 3D or 2C) 
- [ ] Check and log that returned `LineSpread` object reports mean wavelength, FWHM, and amplitude with uncertainties
- [ ] Check and log that FWHM values stay within algorithm bounds (5–30 pixels?) or are flagged

JuergenSchreiber:
Isn't that already done in spec_linespread_e2e?

Issue #646, opened 2025-11-10
Spectroscopy L4->TDA VAP Test 3: Companion‑to‑Host Flux‑Ratio ( PSF‑subtracted)


juliamilton:
Implement a VAP test with logging according to the requirements for L4->TDA spectroscopy processing, based on the Spectroscopy FDD. 

---------------------------------------------------------

### **Test 3: Companion‑to‑Host Flux‑Ratio ( PSF‑subtracted)**
**Use/adapt**: `tests/test_fluxcal.py::test_convert_to_flux()`  
**Functions**: `l4_to_tda.convert_to_flux()`, `l4_to_tda.determine_flux()`  
**Inputs:**:
- L4 PSF‑subtracted cube that contains the companion data
- L4 non‑PSF‑subtracted (un‑occulted) cube of the host star (the reference star).
- Absolute flux‑calibration factor (FluxcalFactor).
- Slit‑transmission map (wavelength-dependent transmission for the FSAM slit)

**Test:**  
- [ ] Check data format and header values per L4 documentation (DATALVL=L4, DPAMNAME = PRISM*, BUNIT=photoelectron/s/bin, ROLL)
- [ ] Confirm both L4 cubes contain a WAVE extension (? confirm how this is stored) with a monotonic wavelength array and that the arrays are identical in length.
- [ ] Extract 1‑D spectra from the PSF‑subtracted cube (companion(λ)) and from the non‑PSF‑subtracted cube (host(λ)).
- [ ] Convert host and companion spectra to physical units - call convert_to_flux() on host(λ) and companion(λ).
- [ ] Apply slit‑transmission loss to host and companion fluxes.
- [ ] Apply color‑correction if present to host and companion fluxes.
- [ ] Compute flux‑ratio per roll. For each roll A and B): R(λ)= companion_corrected(λ)/host_corrected(λ)
- [ ] Roll‑averaging. If both rolls are present, weight the flux ratio by exposure time.

Issue #645, opened 2025-11-10
Spectroscopy L4->TDA VAP Test 2: Unocculted Star in Astrophysical Units


juliamilton:
Implement a VAP test with logging according to the requirements for L4->TDA spectroscopy processing, based on the Spectroscopy FDD. 

---------------------------------------------------------
### **Test 2: Unocculted Star in Astrophysical Units**
**Use/adapt**: `tests/test_fluxcal.py::test_convert_to_flux()` (adapted for 1-D spectra)  
**Functions**: `l4_to_tda.convert_spec_to_flux()`  
**Inputs:**  
- L4 spectroscopy FITS cube of the unocculted host star (per roll used in Test 1)  
- Absolute flux calibration factor (`FluxcalFactor`)  
- Slit-transmission map from `spec.slit_transmission()`  

**Test:**  
- [x] Verify spectroscopy headers (`DATALVL=L4`, `DPAMNAME=PRISM*`, `BUNIT=photoelectron/s/bin`, `ROLL`) and confirm `SPEC`, `SPEC_ERR`, `SPEC_WAVE` extensions exist.  
- [ ] Interpolate the slit-transmission map onto the `SPEC_WAVE` grid if necessary and log the throughput applied per roll.  
- [x] Run `convert_spec_to_flux()` with the fluxcal factor and slit-transmission vector. Ensure `SPEC`/`SPEC_ERR` units update to `erg/(s*cm^2*Å)` and uncertainties propagate.  
- [ ] Log whether `COL_COR` was applied and confirm the wavelength array remains monotonic.  
- [x] Log the value

PR #659, opened 2025-11-12
Spectroscopy L4->TDA VAP test

juliamilton:
## Describe your changes
Implements a VAP test with logging that addresses:

- Unocculted star in astrophysical units (#645)
- Companion to host star flux ratio with roll/ exposure time weighting (#646)
- LSF characterization from narrowband data (#647)
- Slit-transmission map generation (#648)
- L4 spectroscopy input data validation 

## Type of change

#643 

- [ ] I have linted my code

### DIFF
--- a/corgidrp/l4_to_tda.py
+++ b/corgidrp/l4_to_tda.py
@@ -208,6 +208,20 @@ def convert_spec_to_flux(input_dataset, fluxcal_factor, slit_transmission=None):
         if spec_header.get('BUNIT', '').strip().lower() != "photoelectron/s/bin":
             raise ValueError("SPEC extension must have BUNIT 'photoelectron/s/bin' before flux calibration.")
 
+        # Apply algorithm throughput correction (ALGO_THRU) if present (PSF-subtracted frames)
+        if 'ALGO_THRU' in frame.hdu_list:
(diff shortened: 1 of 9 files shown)


juliamilton:
I updated the paths to work with pytest. Two tests print FAIL, due to the L4 files lacking a CT_THRU extension. I believe there is a PR in progress (#666 ) that will resolve this, but then I will need to re-generate the L4 files.

juliamilton:
> @juliamilton I just merged #666. Spec coronagraphic now has an ALGO_THRU extension (throughput value between 0 and 1 per spectral channel). In addition to re-generationg the L4 file for spec coronagraphic, can you also make a small addition in l4_to_tda to correct for algo_throughput using the ALGO_THRU extension on the 1-D extracted spectra (just for coronagraphic obs)?
> 
> If this doesn't make sense, I'll think about how to do it somewhere else.

I implemented that and put it in l4_to_tda.convert_spec_to_flux() since other throughput correction is already also done there (slit transmission) and it needs to happen before flux calibration. Feel free to move it into its own function though if you think that is more appropriate. 

Also, I made new L4 files and put them on the sharepoint under SPEC_sims.

PR #679, opened 2025-11-21
Data format documentation updates for read the docs page

juliamilton:
## Describe your changes
Updates the rst file formatting and index.rst page so that the data format pages will be properly displayed on read the docs.

## Type of change

- [ ] I have linted my code


maxwellmb:
I'm pretty sure the pages are generated in zz_dataformat_e2e with the function `generate_template`, which is built off the template file `data_format_template.rst`.

Looking at the `generate_template` function it seems like one can pass in a `dtype_name` header, which seems to be one of the things you're updating manually. 

Perhaps by adjusting `data_format_template` and passing in `dtype_name` you can make sure you don't have to manually change all these things in the future.

juliamilton:
Things should probably be named in a less confusing way, but the `generate_template` in zz_dataformat_e2e.py doesn't create the ReadTheDocs documentation. It generates a temporary .rst file from actual FITS output which is used later to compare against the specification .rst files in docs/source/data_formats/. The files in docs/source/data_formats/ are intended to be separately generated and maintained so that we're not just comparing the actual output with itself.

semaphoreP:
Overall, the PR looks good to me, just one question: does having separate pages for polarimetry versions of ND filter and abs flux mean they are separate data types? or is it just to document they exist, but otherwise look the same?

PR #686, opened 2025-12-04
Huge dataset accommodation in walker.py

kjl0025:
## Describe your changes
walker.autogen_recipe() now loads in the dataset, regardless of DATALVL value, with no data, err, or dq for any frame so that huge datasets (like the one for noise maps, ~26000 frames) does not exceed 100 GB of RAM.   A few other minor changes were made as well: 
--the mean value of the image-area FPN, dark current, and CIC are now attributes of DetectorNoiseMaps (@agreenbaum , for SSC's purposes)
--the HiSTORY of DetectorNoiseMaps now shows what sets of exposure time, gain, and number of frames were used for creating the calibration.
-- Two calls to "save" were in the l1_to_l2a_noisemap_1.json, which made the number of output files twice as long according to run_recipe().  Fixed now.

## Type of change

N/A

- [x] I have linted my code

### DIFF
--- a/tests/e2e_tests/flatfield_e2e.py
+++ b/tests/e2e_tests/flatfield_e2e.py
@@ -59,9 +59,19 @@ def test_flat_creation_neptune(e2edata_path, e2eoutput_path):
     flat_mock_inputdir = os.path.join(flat_outputdir, "input_l1")
     if not os.path.exists(flat_mock_inputdir):
         os.makedirs(flat_mock_inputdir)    
+    # Empty out flat_mock_inputdir if it contains any files
+    for filename in os.listdir(flat_mock_inputdir):
+        file_path = os.path.join(flat_mock_inputdir, filename)
+        if os.path.isfile(file_path):
+            os.remove(file_path)
     calibrations_dir = os.path.join(flat_outputdir, "calibrations")
     if not os.path.exists(calibrations_dir):
         os.makedirs(calibrations_dir)
+    # Empty out flat_mock_inputdir if it contains any files
+    for filename in os.listdir(calibrations_dir):
+        file_path = os.path.join(calibrations_dir, filename)
+        if os.path.isfile(file_path):
+            os.remove(file_path)
 
     # assume all cals are in the same directory
     nonlin_path = os.path.join(processed_cal_path, "nonlin_table_240322.txt")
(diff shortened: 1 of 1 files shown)


corgidrp/darks.py
```diff
@@ -898,7 +899,15 @@ def calibrate_darks_lsq(dataset, detector_params, weighting=True, detector_regio
     l2a_data_filename = dataset.copy()[-1].filename.split('.fits')[0]
     noise_maps.filename =  l2a_data_filename + '_dnm_cal.fits'
     noise_maps.filename = re.sub('_l[0-9].', '', noise_maps.filename)
-    
+    noise_maps.FPN_im_mean = FPN_image_mean
+    noise_maps.CIC_im_mean = CIC_image_mean
+    noise_maps.DC_im_mean = DC_image_mean
+    noise_maps.FPN_im_median = FPN_image_median
```
semaphoreP: These new attributes dont' seem to get saved into the FITS file when the file gets saved, unless I missed something. Is this correct?

semaphoreP (commented):
I'm running all the e2e tests just in case we missed something. Otherwise I just have one comment about how these new attributes are being saved to disk.

corgidrp/darks.py
kjl0025: Oops, right.  I'll make these header values instead.

corgidrp/walker.py
```diff
@@ -526,11 +525,17 @@ def run_recipe(recipe, save_recipe_file=True):
         if recipe["inputs"]:
             if ram_heavy_bool:
                 curr_dataset = data.Dataset(filelist, no_data=True)
+                recipe_temp = recipe.copy()
+                # don't want to keep all ~26000 filepaths in all ~26000 ext headers b/c that's a lot of memory
+                recipe_temp["inputs"] = "See RECIPE header value in {0}".format(curr_dataset[-1].filepath)
             else:
                 curr_dataset = data.Dataset(filelist)
+                recipe_temp = recipe
+            curr_dataset[-1].ext_hdr["RECIPE"] = json.dumps(recipe)
```
semaphoreP: minor: just for code clarity sake, can you put this line below the comment in the next line. It's also part of the "write the recipe into header" chunk.

PR #732, opened 2026-01-29
Corgisim dataset

helloWorlds0:
## Type of change
 added some option so that WFOV band 4 data can be processed

### DIFF
--- a/corgidrp/data.py
+++ b/corgidrp/data.py
@@ -2672,7 +2672,7 @@ class PyKLIPDataset(pyKLIP_Data):
     """
     A pyKLIP instrument class for Roman Coronagraph Instrument data.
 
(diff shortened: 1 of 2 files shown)


semaphoreP (commented):
Thanks for finding these fixes. I have one comment below and one suggestion for a unit test:

We could copy `test_klipthrupt.test_psfsub_withklipandctmeas_adi()` and do it for Band4 WFOV. For the unit tests, we don't care what the data values look like, so we just want to adjust the unit test so the image shape is correct, and modify the header keywords (e.g., after  the function runs `create_psfsub_dataset()`, manually modify some of the header values). Hopefully this is a relatively straightforward adaptation of an existing unit test.

PR #772, opened 2026-02-25
NORTHANG to replace PA_APER

### LINKED ISSUE #728: Update North angle determination with new PA_APER keyword
ROLL has always been a temporary keyword until we got a more formal definition. We are now replacing it with PA_APER, which is defined to be the CCW angle from Celestial North to +y on the detector (based on spacecraft metrology). This value is not as accurate as the North angle we calculate from our astrometric calibration, but it also factors in the telescope ROLL, so it is still something we need to use. I think for ADI, we can just use PA_APER straight, but for rotating images North, we need to do something more complicated.

To determine the rotation angle for any frame, we should do the following (I think should be done in create_wcs)
 1. Compute `ROLL_OFFSET = PA_APER_frame -PA_APER_astromcal`. This is the roll angle offset between the target frame and the astrometric calibration frame.
 2. Compute a more accurate North angle rotation `NORTHANG = astromcal.northangle + ROLL_OFFSET` . (there may be some negative signs here that I'm missing. we should think about it)
 3. Use the more accurate north angle to rotate each value North up. We should also store our computed NORTHANG into the header (like how we also store PLTSCALE)
 4. Determine if any modifications need to be made in L3 -> L4 processing when rotating north up (for both coron and non-coron modes) to use new NORTHANG keyword instead of ROLL and new PA_APER. I believe this involves changing `combine.derotate_arr()`, `combine.prop_err_dq()` and `data.PyKLIPDataset.readdata()`

aneeshb97:
Introducing a NORTHANG keyword in the ext_hdr to use instead of PA_APER. Corresponding changes to create_wcs_headers(), combine.derotate_arr(), combine.prop_err_dq(), data.PyKLIPDataset.readdata(), the .rst docs and some unit tests. Also initializing a NORTHANG keyword in create_psfsub_dataset() in mocks.py

## Type of change


corgidrp/l2b_to_l3.py
```diff
@@ -38,10 +39,11 @@ def create_wcs(input_dataset, astrom_calibration, offset=None):
         target_ra, target_dec = image.pri_hdr['RA'], image.pri_hdr['DEC']
         corrected_ra, corrected_dec = target_ra - ra_offset, target_dec - dec_offset
         pa_aper_deg = image.pri_hdr['PA_APER']
+        roll_offset_deg = pa_aper_deg - pa_aper_deg_astrom  #Define a roll offset using the pa_aper for the image frame and astrom cal frame
 
         #TO DO: double check this. northangle may be defined as the full rotation angle, 
         # not north offset, in which case, adding the two below would be adding two absolute rotation angles from north
-        vert_ang = np.radians(northangle + pa_aper_deg)  ## might be -pa_aper_deg
+        vert_ang = np.radians(northangle + roll_offset_deg)   ##might be -roll_offset_deg
```
semaphoreP: I look at the diagrams again, and based on the definition of PA_APER (CCW from +N to +y) and North angle (CCW from +y to N), I think we may need the negative sign (so subtract roll_offset_deg, instead of add). If true, also should be done on line 70

semaphoreP (commented):
Changes generally look good. Did you think about whether it should be a + or - addition? I think it may be a negative sign.

corgidrp/l2b_to_l3.py
aneeshb97: Negative sign makes coron imaging break and I have out of ideas how to fix it

semaphoreP: which test are you referring to? l2b_to_l4_e2e?

aneeshb97: yup, thats the one. I can give it one last shot tonight to fix it while using the -ve sign

semaphoreP: maybe for that e2e test, do you also need to negative the value of PA_APER that it gets set to? There's an negative sign there, but perhaps it is unncessary (that the roll angle is in the same direction as the PA_APER keyword)

aneeshb97: oh yes, that's it! If PA_APER is never defined as negative, we should be good

semaphoreP: Have you visually checked images at each roll, and confirm whether the image was rotated CCW, and that is consistent with a +10 change in PA_APER? If not, could you do this manual visual check, to make sure things are consistent?

aneeshb97: I do not think we ever actually rotate the images, but maybe I am missing something...

semaphoreP (approved):

PR #800, opened 2026-03-24
Kgain nonlin ram

kjl0025:
## Describe your changes
Reduces RAM spike for nonlin and k gain calibrations to < 100 GB for float64 data.  The auto-selected recipe is currently [kgain recipe, nonlin recipe] on develop branch.  On this PR, it is a nonlin recipe chain.  So the k gain recipe cannot be auto-selected; it needs to be provided (e.g., see k gain e2e test).  Hence "breaking change" is selected below.

## Type of change


semaphoreP:
There are a lot of changes in this PR that don't seem relevant to this PR: e.g., why are so many recipes modified to have copy_dataset=False, why is the satellite spot XML code in the walker removed, etc?

kjl0025:
Ahh, sorry about that.  That was a copy error.  I had changes on a lab computer to some scripts that were tweaked for RAM testing that weren't synced up right, and I copied and pasted to fix things b/c I wasn't sure if everything should be synced up.  Anyways, I went through line by line and restored what should be unchanged (the walker.py stuff and test_calibrate_darks_lsq.py stuff).  The change to all those recipes:  I added an option to detect_cosmic_rays() to forego copying the input dataset, just as did in the past for prescan_biassub(), to make some e2e tests a little faster.  So the default for that function is to copy the dataset so that the output dataset is a different object than the input (as I did for prescan_biassub()), so I went through and added the keyword in the recipes so that the dataset wouldn't be copied in the function.

semaphoreP:
e2e tests pass for me too. I'm slowly reviewing this, but going to priortize reviewing the higher priority ones first.

corgidrp/recipe_templates/build_trad_dark_full_frame_1.json
semaphoreP, line 1: is this recipe being used anywhere? are the _1 and _2 templates needed?

corgidrp/recipe_templates/build_trad_dark_full_frame_2.json
semaphoreP, line 1: this also seems unused in all the code base

corgidrp/calibrate_nonlin.py
```diff
@@ -966,7 +1020,10 @@ def nonlin_kgain_dataset_2_stack(dataset, apply_dq = True, cal_type='nonlin'):
             if cal_type == 'nonlin':
                 smallest_set_length = None # for nonlin, don't need to truncate exptime sets to be same length for a given EM gain
             for i, exptime_dset_list in enumerate(exptime_dsets):
-                sub = np.stack([dset.frames[0].data for dset in exptime_dset_list[:smallest_set_length]])
+                if ram_heavy:
+                    sub = [dset.frames[0].filepath for dset in exptime_dset_list[:smallest_set_length]]
+                else:
+                    sub = np.stack([dset.frames[0].data for dset in exptime_dset_list[:smallest_set_length]])
```
semaphoreP: to keep the code as similar as possible between ram_heavy and not ram_heavy (and reduce the number of `if` staetements), can both just be lists? and then you cast it to a np.array() at the very end? This way, I think you don't need the two `if` statements below this one.

corgidrp/calibrate_nonlin.py
```diff
@@ -400,26 +405,38 @@ def calibrate_nonlin(dataset_nl,
...
+    if ram_heavy:
+        temp_frame = data.Image(mean_frame_arr[0]) # just to get shape
+        frame_shape = temp_frame.data.shape
+        good_mean_frame = np.zeros(frame_shape)
+        for filepath in mean_frame_arr:
+            temp_frame = data.Image(filepath)
+            if apply_dq:
+                bad = np.where(temp_frame.dq > 0)
+                temp_frame.data[bad] = np.nan
+            good_mean_frame += temp_frame.data
```
semaphoreP, line 417: In both the ram_heavy and non_ram_heavy, we're just adding np.nan to the good_mean_frame. This means that in the final mean frame, any pixel where a single frame was nan will result in nan. Do we want this?

corgidrp/walker.py
```diff
@@ -122,15 +123,16 @@ def walk_corgidrp(filelist, CPGS_XML_filepath, outputdir, template=None):
...
         recipes = [recipes]
-    
+
     # process recipes
     output_filelist = None
     for i, recipe in enumerate(recipes):
         # check for recipe chaining
-        if i > 0 and  len(recipe['inputs']) == 0:
+        if i > 0: 
+            recipe["inputs"] = []
```
semaphoreP: This breaks the previous behavior of the walker which supports both chaining and non-chained sets of recipes. In addition to recipe chaining, we can also process 2 separate recipes in the walker in series. Can we keep the original logic?

corgidrp/walker.py
```diff
@@ -336,7 +338,7 @@ def guess_template(dataset):
                 recipe_filename = ["build_trad_dark_image_1.json", "build_trad_dark_image_2.json"] #"build_trad_dark_image.json"
                 chained = True
         elif image.pri_hdr['VISTYPE'] == "CGIVST_CAL_PUPIL_IMAGING":
-            recipe_filename = ["l1_to_l2a_nonlin.json", "l1_to_kgain.json"]
```
semaphoreP, line 345: don't we want to also produce kgain calibration from this same input dataset? It seems like we have some issues here..

corgidrp/walker.py
```diff
@@ -561,6 +565,8 @@ def run_recipe(recipe, save_recipe_file=True):
                     frame.ext_hdr["RECIPE"] = json.dumps(recipe_temp)
         # execute each pipeline step
         print('Executing recipe: {0}'.format(recipe['name']))
+        if type(filelist) == list:
```
semaphoreP: is this a debug statement or is it intentional?

tests/e2e_tests/nonlin_kgain_e2e.py
semaphoreP, line 1: I think it is problematic that we cannot generate both the kgain and nonline recipes automatically before given an l1 pupil imaging dataset.

tests/e2e_tests/nonlin_RAM_testing.py
```diff
@@ -0,0 +1,313 @@
...
+            prihdr['VISTYPE'] = 'CGIVST_CAL_PUPIL_IMAGING'
+            prihdr['VISTYPE'] = 'CGIVST_CAL_PUPIL_IMAGING'
+        exthdr = fits_file[1].header
+        if exthdr['EMGAIN_A'] == 1:
+            exthdr['EMGAIN_A'] = -1 #for new SSC-updated TVAC files which have EMGAIN_A by default as 1 regardless of the commanded EM gain
+        # Update FITS file
+        fits_file.writeto(file, overwrite=True)
+
+@profile
+@pytest.mark.e2e
```
semaphoreP: if this is not part of the e2e tests, why is this marked e2e?

tests/RAM_testing/sorting_file_generator.py
semaphoreP, line 1: what is this file for, and why is it in the e2e tests folder?

semaphoreP (changes requested):
I think there's a fundamental problem with doing this, as now a l1 pupil imaging dataset cannot automatically produce kgain and nonlin calibrations by default. I think we definitely want that. It seems like splitting the recipe into multiple ram-friendly recipes prevents this from happening, which is a problem. We'd need to either add features into walker.py to handle multiple chained recipes that come from a single l1 dataset, or find another way to reduce RAM.

corgidrp/walker.py
kjl0025: Hmm, is there ever a case where i>0 in the list of recipes in the recipe chain for which the "inputs" is already populated?  I've never encountered that, so I took out the requirement about the 0-length of the inputs.  And doing that allowed my ram_heavy mode in Dataset.save to work.

corgidrp/walker.py
kjl0025: Yes, so I was wondering how you might want to handle this.  Since guess_template doesn't actually decide on a recipe in this case (since the user has to manually choose k gain or nonlin here), I figured I could just choose the nonlin case since it was a chain.  So you prefer having both recipe names/chains available to the user via guess_template?  Alternatively, I could make `recipe_filename = ["l1_to_kgain_1.json", "l1_to_kgain_2.json", "l1_to_l2a_nonlin_1.json", "l1_to_l2a_nonlin_2.json", "l1_to_l2a_nonlin_3.json"]`, and then the user would specify k gain or nonlin from there.

semaphoreP: I would like the walker produce both a kgain and nonlin calibration file if given a list of L1 pupil imaging data. This is what has been implemented and I would like to keep it this way. I don't want manual user specification. It makes me think that this approach to dealing with the RAM issue (is it really an issue?) may not be suitable for Kgain/nonlin.

corgidrp/recipe_templates/build_trad_dark_full_frame_1.json
kjl0025: I added these in for completeness; there's a build_trad_dark_image.json and image_1.json and image_2.json, so I figured I'd put in the RAM-heavy version for the full_ version.  But there's no official plan to use build_trad_dark_full_frame_1.json or _2.json.  I can remove them if you want?

semaphoreP: I'm not a fan of having too many recipes as it makes it hard to go through them. (and it implicitly states that we support this feature) Did you test if they work?

corgidrp/walker.py
semaphoreP: yes in the old kgain/nonlin processing that you changed here. if you go to the develop branch and run the nonlin_kgain_e2e, it will have both recipes populated with input data.

corgidrp/recipe_templates/build_trad_dark_full_frame_1.json
kjl0025: I didn't explicitly test them.  At this point, I'll just remove them.

corgidrp/calibrate_nonlin.py
kjl0025: This function is used for k gain and nonlin calibration, and to make that kind of change would require a structural change to how the k gain function expects the data, so I think it would be easier to leave it like it is.

corgidrp/calibrate_nonlin.py
kjl0025: Yes, this was purposeful when it was implemented prior to this PR.  For k gain, the goal for the mean frame is to have many sample pixels which are each averaged over the full range of frames, so the pixels which have any dq flag are disregarded.

tests/RAM_testing/sorting_file_generator.py
kjl0025: This was for generating the planned files for k gain/nonlin according to CAR-121 (the full 2502 files).  I'll move this into tests folder somewhere.

tests/e2e_tests/nonlin_RAM_testing.py
kjl0025: Oops, will remove.  And I'll move this out of e2e_tests folder

tests/RAM_testing/sorting_file_generator.py
semaphoreP: Is this for CAR-121, and not for DRP e2e testing? Would it be better to put on sharepoint?

kjl0025: Wherever you want to put it is fine with me.  I used it to generate all the frames as specified by CAR-121 needed for the test I did in nonlin_RAM_testing.py, so I was thinking we would include sorting_file_generator.py somewhere in the DRP since nonlin_RAM_testing.py is also in the DRP.

semaphoreP: Ok we can keep it here for now for this PR. But I would consider making all these RAM testing stuff a folder on the sharepoint (both simulation scripts and the data you generated), so that we have it all nicely centralized somewhere.

tests/test_data/nonlin_table_TVAC.fits
semaphoreP, line 1: should this file have been modified?

semaphoreP (commented):
one more addition minor comment on re-review. The changes to the walker look good to me.

tests/test_data/nonlin_table_TVAC.fits
kjl0025: Oops, i missed that one.  That usually slips through .gitignore.   I'll just undo the change for this PR.

semaphoreP (approved):

Issue #795, opened 2026-03-24
additional dark selection criterion


nasavbailey:
Is your feature request related to a problem? Please describe.
caldb.py has an incomplete set of selection criteria that can cause an incorrect dark to be applied. The current implementation matches on exposure time only.

Describe the solution you'd like
darks should match both EXCAM EM gain and exposure time.

juliamilton:
Fixed in #801

Issue #794, opened 2026-03-24
additional criteria for flat selection


nasavbailey:
**Is your feature request related to a problem? Please describe.**
caldb.py has an incomplete set of selection criteria that can cause an incorrect flat to be applied. 

**Describe the solution you'd like**
flats should match on:
- **DPAM**: IMAGING, POL0, or POL45 only. All other DPAM settings, including PUPIL, use flat = ones.
- **CFAM**: detector fringing is wavelength dependent, so I believe the plan is to collect flats for both bands 1F and 4F. 
  - The narrowband sub-filters of bands 1 and 4 that are used for acquisition and HOWFSC have relaxed accuracy needs so can use broadband flats, if available. ie: 1 A, B, C can use 1F. 4 A, B, C can use 4F.  If recent broadband flats are not available, these sub-bands can use ones.
  - No flats are planned for 2F and 3F or their sub-bands (2A, B, ... 3A, ...), because we do not flatfield spectroscopy data.  For all these cases use flat = ones.
- SPAM shouldn't matter b/c it's a pupil plane optic. Do not check.
- LSAM shouldn't matter b/c it's a pupil plane optic. Do not check.
- FSAM will often not match b/c flats are taken in OPEN, while observations are taken with various FSAMs. Do not check.
- **FPAM** needs some thought. Here's my suggestion. I believe baseline is to take flats with and without FPAM in the beam, to allow checking for contamination on the FPM substrate itself. 
  - For no-FPM data (FPAM = OPEN_12 or OPEN_34), the FPAM should match
  - For FPM-in data, we have two options:
    - option 1: use the appropriate no-FPM flat (for all band 1 & 2 FPMs, use the OPEN_12 flat, for all band 3 & 4 FPMs use the OPEN_34 flat).  
    - option 2: since the FPM could in principle move between the flat data collection and the observation data collection, we would need to check that the FPAM H&V matches within a certain tolerance. If it's within tolerance, then use it. Otherwise, synthesize it by using the FPM-in and FPM-out data to decompose the flatfield into contributions from the static optics+detector vs. on the FPM substrate, then translate the FPM substrate components to match observation H&V and then add that to the static component. Using an FPM-in flat would also have the odd effects under the FPM itself b/c you're dividing two tiny numbers, and I don't know how DRP would handle it?
    - In my opinion, we should start with option 1, and only implement option 2 if we find that we have contamination on the FPM substrate that we must account for.

nasavbailey:
[...]

PR #801, opened 2026-03-25
Update caldb.py automatic calibration product selection criteria

juliamilton:
## Describe your changes
Updates the selection criteria for calibration products in caldb.py for when a recipe uses automatic selection to choose a calibration product. Selection logic in this PR:

- Dark (#795): 
        - EXPTIME
        - EMGAIN_C
        - PC_STAT (exclude PC master dark for analog processing, prefer PC for PC processing)
- FlatField (#794): 
         - If DPAM not in [IMAGING, POL0, POL45] or CFAM in spectroscopy/CLEAR use ones_flat
         - If CFAMNAME not in imaging set, error; else CFAMNAME (sub-band maps to corresponding broadband)
         - DPAMNAME selection
         - FPAMNAME starts with OPEN
- FluxcalFactor: 
         - CFAMNAME
         - DPAMNAME
- CoreThroughputCalibration:
        - FPAMNAME
        - CFAMNAME
- MuellerMatrix: 
         - CFAMNAME 
- SlitTransmission: 
         - FSAMNAME
         - DPAMNAME
         - CFAMNAME
- LineSpread:
         - DPAMNAME
         - CFAMNAME
- SpectroscopyCentroidPSF:
         - DPAMNAME
         - CFAMNAME

Calibration products which are needed for datasets with mixed PAM settings will be addressed in #820 . (NDFilterSweetSpot, NDSpectroscopy, NDMuellerMatrix, SpecFluxCal, DispersionModel.)

All other calibration products use closest-in-time selection.

## Type of change

#795 #794 

- [ ] I have checked that I am merging into the right branch


nasavbailey:
Just checking flatfield: The CFAM criterion appears to be more strict than what was suggested in https://github.com/roman-corgi/corgidrp/issues/794. Was that intentional?

juliamilton:
> Just checking flatfield: The CFAM criterion appears to be more strict than what was suggested in #794. Was that intentional?

Ah ok - I think I misinterpreted what was written. A follow up question for you: 
If a subband flat is available in either band 1 or band 4, is that still preferable to use instead of the broadband flat? Should the logic be: try subband, if not available: try broadband, if not available: use all ones?

nasavbailey:
For sub-bands, I think pick the most recent of [matching sub-band, broadband]. I think we will rarely (if ever) take sub-band flats, so what I'd want to avoid is picking a year-old sub-band flat over a week-old broadband flat. tagging @ecady-jpl since HOWFSC is currently the only sub-band customer I'm aware of

ecady-jpl:
HOWFSC will not have different flats for different subbands because the HOWFSC processing is onboard using a single large data structure for the flats.  So HOWFSC will either use all ones as the onboard default (which historically has been fine) or upload a single representative flat that covers all subbands, but we won't use individual subband flats.

juliamilton:
Sounds good. Since there are no other known customers for a sub-band flat at the moment and Eric's case explicitly only wants the broadband flat, I will implement it to always choose the closest-in-time matching broadband flat. If there is a need in the future for a sub-band flat we could use a custom recipe. Does that work?

corgidrp/caldb.py
```diff
@@ -383,13 +381,38 @@ def get_calib(self, frame, dtype, to_disk=True):
...
-            options = self.filter_calib(calibdf, "CFAMNAME", frame_dict['CFAMNAME'], err_if_none=False)
-            if frame_dict['DPAMNAME'] in ['IMAGING', 'POL0', 'POL45']:
+            # DPAM: IMAGING, POL0, or POL45 only. All other DPAM settings, including PUPIL, use flat = ones
+            # CFAM: spectroscopy bands (2F/3F and their sub-filters) use flat = ones
+            # Sub-bands 1A/1B/1C and 4A/4B/4C are mapped to their corresponding broadband (1F, 4F)
+            # FPAM: FPM-in data uses the open-substrate flat for the appropriate band pair (OPEN_12 or OPEN_34)
+            spectroscopy_cfams = {'2F', '3F', '2A', '2B', '2C', '3A', '3B', '3C', '3D', '3E', '3G'}
+            cfam_subband_map = {'1A': '1F', '1B': '1F', '1C': '1F',
+                                '4A': '4F', '4B': '4F', '4C': '4F'}
+            if frame_dict['DPAMNAME'] not in ['IMAGING', 'POL0', 'POL45'] or frame_dict['CFAMNAME'] in spectroscopy_cfams:
```
nasavbailey: CFAM has other named positions that this logic wouldn't catch. I'm most worried about CLEAR (should be flat = ones?). For other positions (DARK, LAUNCH, etc), do you want to throw an error b/c those types shouldn't be flatfielded or do you want to return all ones to enable no-op processing (in which case change logic should be if not in [good set])?

corgidrp/caldb.py
```diff
@@ -383,13 +381,38 @@ def get_calib(self, frame, dtype, to_disk=True):
...
+                options = calibdf[calibdf["Filepath"].str.endswith("ones_flat.fits")]
+                if len(options) == 0:
+                    raise ValueError("No ones_flat.fits found in caldb at {0}".format(self.filepath))
+                result_index = options["MJD"].argmax()
+            else:
+                # sub-bands map to their corresponding broadband flat
+                cfam_lookup = cfam_subband_map.get(frame_dict['CFAMNAME'], frame_dict['CFAMNAME'])
```
nasavbailey, line 427: I think this will return nothing for 1F and 4F, resulting in no flat returned for those filters on the next line. should 1F and 4F be in the dictionary?  Also, why is frame_dict['CFAMNAME'] passed in twice to .get?

corgidrp/caldb.py
```diff
@@ -383,13 +381,38 @@ def get_calib(self, frame, dtype, to_disk=True):
...
+                    fpam_lookup = fpam
+                elif fpam.startswith('SPC12') or fpam.startswith('HLC12'):
+                    # FPM-in data uses the open substrate flat for the appropriate band 
+                    fpam_lookup = 'OPEN_12'
+                elif fpam.startswith('SPC34') or fpam.startswith('HLC34'):
+                    fpam_lookup = 'OPEN_34'
+                else:
+                    # no dedicated flats for other FPAM positions (eg ND225, ND475)
+                    # fall back to the open substrate flat for the appropriate band
+                    fpam_lookup = 'OPEN_12' if cfam_lookup == '1F' else 'OPEN_34'
```
nasavbailey: future proofing: if bands 2 or if subbands of 1 and 2 were added in the future, defaulting all CFAM except 1F to OPEN_34 would produce incorrect results for those caess. should this explicitly check for 1f and 4f and return ones otherwise?  

Alternatively, the key point is that FPAMNAME needs to start with OPEN. We should never configure FPAM=OPEN_12 with CFAM=3/4  or vice versa b/c FPAMs are AR coated for either bands 1/2 or bands 3/4. So could this entire section be simplified to just select on "starts with OPEN"?

corgidrp/caldb.py
```diff
@@ -357,11 +357,9 @@ def get_calib(self, frame, dtype, to_disk=True):
             result_index = np.abs(options["MJD"] - frame_dict["MJD"]).argmin()
             calib_filepath = options.iloc[result_index, 0]
         elif dtype_label in ['NDFilterSweetSpot']:
-            # filter by color filter (CFAM) and ND filter (FPAM)
             # filter_calib() is configured to not throw an error if no matches are found, so that
             # no existing e2e tests breaks, if in the future we want to strictly only use the calibration
             # files with matching headers, then set err_if_none to True
-            options = self.filter_calib(calibdf, "CFAMNAME", frame_dict['CFAMNAME'], err_if_none=False)
```
nasavbailey, line 365: why remove CFAMNAME check? ND has significant variation in OD with wavelength. I thought that ND would be calibrated in each broadband?

corgidrp/caldb.py
```diff
@@ -357,11 +357,9 @@ def get_calib(self, frame, dtype, to_disk=True):
             result_index = np.abs(options["MJD"] - frame_dict["MJD"]).argmin()
             calib_filepath = options.iloc[result_index, 0]
         elif dtype_label in ['NDFilterSweetSpot']:
-            # filter by color filter (CFAM) and ND filter (FPAM)
             # filter_calib() is configured to not throw an error if no matches are found, so that
             # no existing e2e tests breaks, if in the future we want to strictly only use the calibration
             # files with matching headers, then set err_if_none to True
```
nasavbailey: is DPAM check missing? how to ensure spectroscopy wavelength-dependent ND filter cal is differentiated from imaging scalar cal? Are pol planning to measure this with each WP as well (so default to match POL# but fall back to IMAGING in same CFAM?)

corgidrp/caldb.py
```diff
@@ -400,15 +423,14 @@ def get_calib(self, frame, dtype, to_disk=True):
...
         elif dtype_label in ['MuellerMatrix']:
-            # filter by color filter and Wollaston
+            # filter by color filter
             options = self.filter_calib(calibdf, "CFAMNAME", frame_dict['CFAMNAME'], err_if_none=False)
-            options = self.filter_calib(options, "DPAMNAME", frame_dict['DPAMNAME'], err_if_none=False)
 
             # select the one closest in time
             result_index = np.abs(options["MJD"] - frame_dict["MJD"]).argmin()
             calib_filepath = options.iloc[result_index, 0]
         elif dtype_label in ['NDMuellerMatrix']:
```
nasavbailey: is this missing a match on CFAM?

corgidrp/caldb.py
nasavbailey, line 447: is this missing a CFAM check?

neilzim, line 447: We should also match this according to DPAM (either PRISM2 or PRISM3)

neilzim: For flux calibration, we may need to add a filter for SPAM to differentiate between SPEC and SPECROT, since the shaped pupil masks have different transmission factors.

corgidrp/caldb.py
juliamilton: I think this one should work. The syntax is `.get(key, default)` and the second argument is the default value returned when the key isn't in the dict. So this logic is: look up the CFAMNAME in the sub-band map and if it's not there (ie it's already a broadband filter like 1F or 4F), just use the CFAMNAME name itself.

But, to your above point, if there are other named positions that wont have their own flats, the logic still needs to be modified to deal with that.

corgidrp/caldb.py
juliamilton: Hmm good question. If those other named positions won't have their own flats, would we use a custom recipe to process them? If so, perhaps we could throw an error here with the intent that we shouldn't be trying to automatically process them.

corgidrp/caldb.py
kjl0025, line 364: If the frame is analog (ext_hdr['ISPC'] = 0), we should check that we aren't getting a PC master dark (so ext_hdr['PC_STAT'] != 'photon-counted master dark') that matches in EMGAIN_C and EXPTIME (even though the chances of an analog-frame match are low).  And if the frame is PC ((ext_hdr['ISPC'] = 1), we should give preference to the latest PC master dark (ext_hdr['PC_STAT'] = 'photon-counted master dark').

Rumpelstil:
My (generic) comment: do we or should we care about DM1 and DM2 settings for the calibration data base? Are any calibrations apart from astrometry potentially affected by the DM settings (flat DMs, etc.)?

nasavbailey:
> My (generic) comment: do we or should we care about DM1 and DM2 settings for the calibration data base? Are any calibrations apart from astrometry potentially affected by the DM settings (flat DMs, etc.)?

I don't think DM settings are accessible in caldb (others should correct me if that's not true!).  I think the assumption is that the closest in time match is the one most likely to have matching DMs. Generally true, probably, and we can manually edit recipes when it's not true?

corgidrp/caldb.py
mygouf: I assume if SPEC or SPECROT are selected for SPAM the default is to have same value for LSAM? Or should we check both to be sure?

mygouf: More generally, what about WFOV vs OPEN for SPAM or WFOV vs NFOV for LSAM? No risk of mixing those?

corgidrp/caldb.py
```diff
@@ -350,47 +351,107 @@ def get_calib(self, frame, dtype, to_disk=True):
...
-            options = self.filter_calib(calibdf, "CFAMNAME", frame_dict['CFAMNAME'], err_if_none=False)
-            if frame_dict['DPAMNAME'] in ['POL0', 'POL45']:
-                options = self.filter_calib(options, "DPAMNAME", frame_dict['DPAMNAME'], err_if_none=False)
+            options = self.filter_calib(calibdf, "CFAMNAME", frame_dict['CFAMNAME'], err_if_none=True)
+            options = self.filter_calib(options, "DPAMNAME", frame_dict['DPAMNAME'], err_if_none=True)
 
             # select the one closest in time
             result_index = np.abs(options["MJD"] - frame_dict["MJD"]).argmin()
             calib_filepath = options.iloc[result_index, 0]
         elif dtype_label in ['CoreThroughputCalibration']:
```
mygouf, line 395: Does that rule encompasses the 2 core throughput calibration products?

corgidrp/caldb.py
```diff
@@ -436,7 +497,7 @@ def scan_dir_for_new_entries(self, filedir, look_in_subfolders=True, to_disk=Tru
         for calib_frame in calib_frames:
             self.create_entry(calib_frame, to_disk=to_disk)
 
-    def filter_calib(self, calibdf, col_name, value, err_if_none=False):
+    def filter_calib(self, calibdf, col_name, value, err_if_none=True):
         '''
         Takes in a calibration dataframe, filters them so that
         only the files with matching header values are returned. If none is found,
```
mygouf, line 520: Just to cross-check, no special selection needed for:
- astrometric calibration (same for all configs?)
- line spread function (same for nominal and rotated?)  
- wavelength zeropoint (same for nominal and rotated?)  
- registered prism image of target and ref stars (same for nominal and rotated?)

corgidrp/caldb.py
juliamilton: Updated to make CLEAR use a flat of ones, and other named positions outside of bands 1-4 to error

corgidrp/caldb.py
juliamilton: Thanks - implemented the logic which selects on FPAMNAME starts with "OPEN"

corgidrp/caldb.py
juliamilton: Per slack discussion, ND filter selection now uses DPAM, FPAM, and CFAM (with sub-bands falling back to matching broadband).

corgidrp/caldb.py
juliamilton: I don't think spectroscopy or polarimetry modes will use this NDFilterSweetSpot calibration product. Spectroscopy ND filter cal is a new and separate datatype which I've now added in. Polarimetry has a separate ND filter Mueller matrix product which is the only ND filter-related product used in those calculations.

corgidrp/caldb.py
juliamilton: Added

juliamilton: Added

juliamilton: added SPAMNAME to column_dtypes and updated the SpecFluxCal filter to use it

juliamilton: added PC_STAT to column_dtypes and updated the dark logic to use it as described above

corgidrp/caldb.py
juliamilton: Added

corgidrp/caldb.py
juliamilton: Core throughput map is only created as an intermediate product and never gets read in from the caldb, so it should be ok.

corgidrp/caldb.py
juliamilton: I think astrometric calibration is the same for all configs (at least, that is how it is currently set up in the code). 

I've added logic for line spread function to select on DPAM and CFAM because the code doesn't have any specific FPAM/SPAM processing differences, but perhaps @neilzim can clarify if either of those are needed in the selection logic. 

Wavelength zero point and registered prism image of target/ ref stars are intermediate processing steps and aren't saved as separate calibration products.

juliamilton:
@semaphoreP @maxwellmb This is ready for review, however, after switching the logic in caldb to error if no matching calibration found (`error_if_none=true`), two of the e2e tests fail. I've yet to resolve those, but think it may be better for a different PR.

semaphoreP:
@juliamilton is it just these two errors:

```
FAILED tests/e2e_tests/spec_l3_to_l4_noncoron_e2e.py::test_run_end_to_end - ValueError: No valid calibration with CFAMNAME=3)
FAILED tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py::test_run_end_to_end - ValueError: No valid calibration with CFAMNAME=3F)
```

If so, I think the fixes are probably pretty straightforward (one of the calibration files just needs its CFAMNAME updated I think..). If it is indeed not huge fixes, I'd like to fix them here.

corgidrp/caldb.py
```diff
@@ -351,47 +355,150 @@ def get_calib(self, frame, dtype, to_disk=True):
...
+            if is_pc:
+                # prefer PC master dark if available, otherwise fall back to any matching dark
+                pc_options = options[options['PC_STAT'] == 'photon-counted master dark']
+                if len(pc_options) > 0:
+                    options = pc_options
+            else:
+                # analog frame: do not select a matching PC master dark
+                options = options[options['PC_STAT'] != 'photon-counted master dark']
+                if len(options) == 0:
+                    raise ValueError("No valid analog Dark calibration in caldb located at {0}".format(self.filepath))
```
semaphoreP: a nice to have: specify that there were no analog master darks with this EXPTIME and EMGAIN_C found.

corgidrp/caldb.py
```diff
@@ -351,47 +355,150 @@ def get_calib(self, frame, dtype, to_disk=True):
...
+            #       all other non-imaging CFAM positions should use a custom recipe, not auto-selection
+            # Sub-bands 1A/1B/1C and 4A/4B/4C are mapped to their corresponding broadband (1F, 4F)
+            # FPAM: FPM-in data uses the open-substrate flat for the appropriate band pair (OPEN_12 or OPEN_34)
+            spectroscopy_cfams = {'2F', '3F', '2A', '2B', '2C', '3A', '3B', '3C', '3D', '3E', '3G'}
+            imaging_cfams = {'1F', '1A', '1B', '1C', '4F', '4A', '4B', '4C'}
+            cfam_subband_map = {'1A': '1F', '1B': '1F', '1C': '1F',
+                                '4A': '4F', '4B': '4F', '4C': '4F'}
+            ones_flat_cfams = spectroscopy_cfams | {'CLEAR'}
+            if frame_dict['DPAMNAME'] not in ['IMAGING', 'POL0', 'POL45'] or frame_dict['CFAMNAME'] in ones_flat_cfams:
+                options = calibdf[calibdf["Filepath"].str.endswith("ones_flat.fits")]
```
semaphoreP: is this a file that exists? should we create a default flat calib in caldb.initiliaze() do include this?

corgidrp/caldb.py
```diff
@@ -351,47 +355,150 @@ def get_calib(self, frame, dtype, to_disk=True):
...
+                if len(options) == 0:
+                    raise ValueError("No ones_flat.fits found in caldb at {0}".format(self.filepath))
+                result_index = options["MJD"].argmax()
```
semaphoreP, line 420: it doesn't matter which one we grab right? maybe just a comment to this effect that it doesn't matter so we're just grabbing the lattest one.

semaphoreP:
Seems like there are more e2e tests failing than those two now that we added these changes. @juliamilton can you or someone make a PR to this PR to fix those e2e tests? that way we can keep the PRs separate, but I don't want to merge this to develop with many e2e tests breaking, given there's a bunch of other development that's going on.

semaphoreP (commented):
main thing is do we need to add a ones_flat default calibration file? otherwise, 2 other very minor things. 

We can fix e2e tests in a separate PR that goes into this PR, but I don't want to merge until we fix the e2e tests.

juliamilton:
> Seems like there are more e2e tests failing than those two now that we added these changes. @juliamilton can you or someone make a PR to this PR to fix those e2e tests? that way we can keep the PRs separate, but I don't want to merge this to develop with many e2e tests breaking, given there's a bunch of other development that's going on.

I'll work on it and try to get it in today

corgidrp/caldb.py
juliamilton: added in a default flat calib that is made during caldb initialization (it follows normal cgi filename conventions but has FPAMNAME=ONES for identification)

corgidrp/caldb.py
juliamilton: Added this to the message

corgidrp/caldb.py
juliamilton: Added

corgidrp/caldb.py
```diff
@@ -543,6 +622,29 @@ def initialize():
         disp_model.save(output_dir, disp_model.filename)
         rescan_needed = True
 
+    # Add default ones-flat FlatField calibration file
+    ones_flat_exists = False
```
semaphoreP: could the filename be static/fixed so we don't have to do this search of the directory?

corgidrp/data.py
```diff
@@ -602,11 +602,21 @@ def _record_parent_filenames(self, input_dataset):
...
+            # Some synthetic products may not have a filename yet, prefer `filename`, fall back to `filepath'
+            if getattr(img, "filename", None):
+                parent_filenames.add(img.filename)
+            elif getattr(img, "filepath", None):
+                parent_filenames.add(img.filepath)
             # also check if this frame has parent frames. keep trakc of them too
             if 'DRPNFILE' in img.ext_hdr:
                 for j in range(img.ext_hdr['DRPNFILE']):
-                    parent_filenames.add(img.ext_hdr['FILE{0}'.format(j)])
+                    # Support FILE0- and FILE1- conventions, and deal with missing cards (for synthetic calibration products)
```
semaphoreP, line 628: it seems like a bug that some synthetic calibrations use the file convention instead of the file0 convention. Should we be filing a bug and do you know which synthetic calibrations this is?

corgidrp/data.py
```diff
@@ -1619,14 +1629,32 @@ class BadPixelMap(Image):
...
+                    if getattr(f, "ext_hdr", {}).get("DATATYPE") in ("Dark", "MasterDark")
+                    or "_drk" in str(getattr(f, "filename", "")).lower()
+                ]
+                if not dark_frames:
+                    raise ValueError(
+                        "BadPixelMap input_dataset must contain at least one dark(-like) frame "
+                        "(DATATYPE='Dark'/'MasterDark' or filename containing '_drk'). "
+                    )
+                base_dataset = Dataset(dark_frames)
+            except Exception:
```
semaphoreP: why do you need try/exception if you just re-throw the exception?

corgidrp/data.py
```diff
@@ -2214,8 +2242,11 @@ def __init__(self, data_or_filepath, err = None, pri_hdr=None, ext_hdr=None, err
             # use the start date for the filename by default
             self.filedir = "."
             # slight hack for old mocks not in the standard filename format
-            self.filename = "{0}_abf_cal.fits".format(orig_input_filename)
-            self.filename = re.sub('_l[0-9].', '', self.filename)
+            if input_dataset is not None:
```
semaphoreP, line 2257: this also seems like a bug? we should be throwing an error if we're making a new fluxcalfactor without an input dataset? maybe an issue to be made as well?

corgidrp/flat.py
```diff
@@ -42,6 +42,39 @@ def create_flatfield(flat_dataset):
...
+
+    # ERR headers
+    if getattr(flat_field, "err_hdr", None) is not None:
+        if "BUNIT" not in flat_field.err_hdr:
+            flat_field.err_hdr["BUNIT"] = "photoelectron"
+        for k in ("KGAIN_ER", "RN_ERR", "DESMEAR"):
+            if k in flat_field.ext_hdr and k not in flat_field.err_hdr:
+                flat_field.err_hdr[k] = flat_field.ext_hdr[k]
+        if "LAYER_1" not in flat_field.err_hdr:
+            flat_field.err_hdr["LAYER_1"] = "combined_error"
```
semaphoreP, line 76: this seems like a hack we should make an issue to fix? maybe the right thing is to fix the mocking code or the e2e tests, rather than putting it here.

tests/e2e_tests/l1_to_l4_pol_e2e.py
```diff
@@ -407,6 +416,8 @@ def run_l1_to_l4_e2e_test(l1_datadir, l4_outputdir, processed_cal_path, logger):
     ##############################################
     # Step 1/2 b: Do the stap spot data. 
     logger.info('Running L1 to L2a on the polarimetry satspots...')
+    # Ensure subsequent walker calls use this test's CalDB
+    corgidrp.caldb_filepath = tmp_caldb_csv
```
semaphoreP: why do we need to set this again? Is this a symptom of a different issue?

semaphoreP (commented):
I have one small request on the naming of the ONES flat field. Otherwise, there are several questions, which the action item may be to raise issues to figure out later. They seem suspicious and I don't really want to permanently cover them up with some code that suppresses the symptom of a possibly underlying issue.

corgidrp/data.py
juliamilton: The FILE0 vs FILE_0 difference comes from two separate places. FILE0 comes from the DRP: if DRPNFILE = N, then the header has FILE0..FILE(N). FILE_0 style headers come from PYKLIP outputs. I agree this is confusing and worth filing a bug over.

corgidrp/data.py
juliamilton: I added this because some places make mock fluxcal factors without passing in input datasets when doing that. Passing in a dummy frame as an input dataset means you need to be diligent about making sure the frame also has the correct headers, which had been causing issues.

corgidrp/flat.py
juliamilton: We could make an issue about it - the better fix would be to make sure all the mocks and synthetic input flats have consistent headers. The synthetic flat fields (like the all ones flat) don't have those frame select headers or the other headers, so this code adds them in if they don't exist.

corgidrp/data.py
juliamilton: removed it

tests/e2e_tests/l1_to_l4_pol_e2e.py
juliamilton: looks like a holdover from the JIT issue debugging - removed

corgidrp/caldb.py
juliamilton: Ok - I kept the cgi filename convention (as opposed to naming it something like ones_flat.fits) since there are other places in the code that search for a flt_cal.fits. But it has a static filename now.

corgidrp/data.py
semaphoreP: Ok so the issue is only for data products of do_psf_subtraction()?

corgidrp/caldb.py
semaphoreP: ok because we have a static filename, can we remove the entire try/except where we search the directory for flats, and just check for the existence of the specific filename?

corgidrp/data.py
semaphoreP: But it sounds like we should do this properly in the future, so we should make an issue to track this

corgidrp/flat.py
semaphoreP: Sounds like an issue to be made so we can track this

corgidrp/data.py
juliamilton: Yes. I've made an issue that tracks this topic here: #827

corgidrp/data.py
juliamilton: Tracked in #825

corgidrp/flat.py
juliamilton: Tracked in #826

semaphoreP (approved):

Issue #798, opened 2026-03-24
L1 to SpecFluxCal E2E Test


semaphoreP:
**Is your feature request related to a problem? Please describe.**
We want an e2e test to generate a SpecFluxCal data from L1 data. 

**Describe the solution you'd like**
A separate e2e test for this. L1 data should exist on the Sharepoint. We probably need to modify the walker and update the data format pages as well.

semaphoreP:
Completed in #817 with data format page added in #819

PR #817, opened 2026-03-31
L1 to SpecFluxCal E2E Test #798

JuergenSchreiber:
## Describe your changes
added an L1 to SpecFluxCal e2e test to test the generation of spectral flux calibration factors. Changed the walker to run l2b_to_specflux.json recipe if vistype is CGIVST_CAL_ABSFLUX... and no ND filter in FPAM

## Type of change

#798


corgidrp/walker.py
semaphoreP, line 1: can you also add some code to guess_template to handle the l1 case with recipe chaining? (basically do the l1_to_l2a, l2a_to_l2b, and then l2b_to_spec_flux recipes in a chain). This will make the e2e test simpler too.

tests/e2e_tests/l1_to_spec_flux_e2e.py
```diff
@@ -0,0 +1,534 @@
...
+    # ------------------------------------------------------------------ 
+    # 1. Caldb                                             
+    # ------------------------------------------------------------------ 
+    calibrations_dir = os.path.join(outputdir, 'calibrations')
+    os.makedirs(calibrations_dir, exist_ok=True)
+
+    this_caldb, is_pc_data = setup_caldb(
+        l1_datadir, processed_cal_path, calibrations_dir)
+
+    # ------------------------------------------------------------------
```
semaphoreP, line 287: question: do you have the same issue here as in ND Spectroscopy where the L1 data have different exposure times, so we aren't able to process all of them together for L1 -> L2b? If not, could we process L1 -> SpecFluxcal in one walker call?

tests/e2e_tests/l1_to_spec_flux_e2e.py
```diff
@@ -0,0 +1,534 @@
...
+
+    l1_datadir         = os.path.join(args.e2edata_dir, "ND_SPEC", "L1")
+    processed_cal_path = os.path.join(args.e2edata_dir, "ND_SPEC", "Cals")
+    outputdir = os.path.join(args.outputdir, "l1_to_spec_fluxcal_e2e")
+
+    if os.path.exists(outputdir):
+        shutil.rmtree(outputdir)
+    os.makedirs(outputdir)
+
+    run_spec_flux_e2e(l1_datadir, processed_cal_path, outputdir)
```
semaphoreP: you should just call test_spec_fluxcal_e2e.py so that we don't need to have all this code. is there a reason why you have this?

semaphoreP (commented):
a few small comments/questions. Otherwise looks good.

tests/e2e_tests/l1_to_spec_flux_e2e.py
JuergenSchreiber: Aeh, no, I copied it from Julias l1_to_nd_filter_spec_e2e, maybe should be changed there, too.

corgidrp/walker.py
JuergenSchreiber: Ok, I did. There is a thing I am not sure of, since the nd filter calibration is done, if the ND filter is set, and the flux cal is done if not. But I guess for bright sources we also want an abs flux calibration with ND filter set. But how to distinguish?

tests/e2e_tests/l1_to_spec_flux_e2e.py
JuergenSchreiber: No, I only use the faint star with the same exposure time, so we can indeed shorten the script.

tests/e2e_tests/l1_to_spec_flux_e2e.py
semaphoreP: good to note. we should do that, but not this PR.

corgidrp/walker.py
semaphoreP: yeah I think we can't process calibration data of just the ND Filter without the abs flux cal. The spec ND filter step function will error if this happens, so I think we don't need to worry about it here in the walker right now.

semaphoreP (approved):

Issue #824, opened 2026-04-02
Polarimetry L4: record both POL0 and POL45 abs flux calibrations in headers


juliamilton:
Per the FDD, L4 images must be associated with an absolute flux calibration (FluxcalFactor).

Currently,` l3_to_l4_pol` only uses FluxcalFactor in one place: to put the name of the associated FluxcalFactor into the headers in `update_to_l4`. The walker finds one cal using filelist[0] as the reference frame and matches on DPAMNAME, so you get either the associated POL0 or POL45 abf_cal product written in to the header, depending on input file order. 

However, both POL0 and POL45 abf_cal products should be written into the header for completeness.

semaphoreP:
Fixed in #828

PR #828, opened 2026-04-03
Record both POL0 and POL45 flux calibrations in L4 pol headers

john-livingston:
## Describe your changes

- Added `FluxcalFactorPOL0` and `FluxcalFactorPOL45` wrapper classes in `data.py`, following the existing `FlatFieldPOL0`/`FlatFieldPOL45` pattern
- Extended `caldb.py` to handle POL0/POL45 flux cal lookups by forcing `DPAMNAME` during calibration selection
- Added walker special-case loading for the new types in `walker.py`
- Created new `update_to_l4_pol` step function in `l3_to_l4.py` that writes both `FLXCALF0` (POL0) and `FLXCALF1` (POL45) header keywords
- Updated `l3_to_l4_pol.json` recipe template to use `update_to_l4_pol` with both flux cal types
- Updated `l4_pol.rst` data format docs to reflect the new header keywords
- Updated `l3_to_l4_pol_e2e.py` to create and verify both flux cal entries
- Added `test_update_to_l4_pol` unit test in `test_polarimetry.py`

## Type of change

This PR addresses issue #824

- [x] I have linted my code


corgidrp/l3_to_l4.py
```diff
@@ -1867,3 +1867,41 @@ def update_to_l4(input_dataset, corethroughput_cal, flux_cal):
...
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (L3-level)
+        corethroughput_cal (corgidrp.data.CoreThroughputCalibration): a CoreThroughputCalibration calibration file. Can be None
+        flux_cal_pol0 (corgidrp.data.FluxcalFactor): the POL0 FluxcalFactor calibration file
+        flux_cal_pol45 (corgidrp.data.FluxcalFactor): the POL45 FluxcalFactor calibration file
+
+    Returns:
+        corgidrp.data.Dataset: same dataset now at L4-level
+    """
+    for orig_frame in input_dataset:
```
semaphoreP: for the implementation of this function, could we call `update_to_l4()` inside this function, as to maximize code reuse? Then, all we need to do in this function is to update some header keywords.

docs/source/data_formats/l4_pol.rst
```diff
@@ -462,7 +462,9 @@ Image Header (HDU 1)
 +------------+------------+--------------------------------+--------------------------------------------------------------------------------------------------------------------------+
 | CTCALFN    | str        | cgi_0200001001001001001_202... | Core throughput calibration file name                                                                                    |
 +------------+------------+--------------------------------+--------------------------------------------------------------------------------------------------------------------------+
-| FLXCALFN   | str        | cgi_0200001001001001001_202... | Flux calibration file name                                                                                               |
```
semaphoreP, line 471: Max suggests FLXCLF0 and FLXCLF45 instead to be more descriptive. not sure if @agreenbaum or @juliamilton have any opinions on naming (we're splitting the FLXCALFN keyword for pol mode L4 data).

semaphoreP (commented):
Thanks for the quick fix on this! I think it generally looks good. I have one suggestion, and a question regarding header naming (that we should think about a little bit, but don't spend too much time thinking about )

corgidrp/l3_to_l4.py
john-livingston: Good point! Done - update_to_l4_pol now calls update_to_l4 internally, then replaces FLXCALFN with the two pol-specific headers. Will push once we settle on naming.

docs/source/data_formats/l4_pol.rst
john-livingston: Yes, I also considered e.g. FLXCAL00 and FLXCAL45, but opted to try to maximize consistency with the existing keywords. But FLXCLF0 and FLXCLF45 seem like a good middle ground

john-livingston: OK, sounds like there are no objections to FLXCLF0 and FLXCLF45, so I'll go with those.

semaphoreP (approved):

Issue #593, opened 2025-10-17
Improved mueller matrix cal errors with input errors


maxwellmb:
**Is your feature request related to a problem? Please describe.**
The error propagation for creating the mueller matrix cals only includes the errors from the measurements, not the errors from the knowledge of the source polarization. 

**Describe the solution you'd like**
Including the errors on the source polarization would be better.

maxwellmb:
Closed with PR #833

PR #833, opened 2026-04-04
Propagate reference star polarization errors into Mueller matrix calibration

john-livingston:
## Describe your changes

`generate_mueller_matrix_cal` had two related bugs:

1. The reference catalog columns `P_err` and `PA_err` (uncertainties on standard star polarization fraction and angle) were read in but never used. MM errors were therefore underestimated whenever reference star polarizations are imprecisely known. These are now propagated via first-order error propagation and combined in quadrature with the photometric measurement errors before the SVD-based MM inversion.

2. The computed `mueller_matrix_err` array was never passed to the `MuellerMatrix`/`NDMuellerMatrix` constructor (`err=` kwarg was unused), so `.err` was silently all-zeros on every returned object. Additionally, both constructors accepted `err=` but silently dropped it in the `super().__init__()` call. Both are now fixed.

### Changes
- `corgidrp/pol.py`: propagate P_err/PA_err in the stokes_matrix build loop; pass `err=mueller_matrix_err` to both constructors; update docstring
- `corgidrp/data.py`: fix `MuellerMatrix` and `NDMuellerMatrix` constructors to pass `err=` through to `Image.__init__()`
- `tests/test_polarimetry.py`: two new tests verifying errors are stored and that source polarization uncertainties increase MM errors

## Type of change

This PR addresses issue #593

- [x] I have linted my code


corgidrp/pol.py
```diff
@@ -305,12 +307,26 @@ def generate_mueller_matrix_cal(input_dataset,
...
+        # propagate reference star polarization uncertainties into stokes_vector_errs
+        # Q = P*cos(2*PA), U = P*sin(2*PA)
+        Q_ref_err = np.sqrt((np.cos(2*PA_rad) * P_err)**2 +
+                            (2 * P * np.sin(2*PA_rad) * PA_err_rad)**2)
+        U_ref_err = np.sqrt((np.sin(2*PA_rad) * P_err)**2 +
+                            (2 * P * np.cos(2*PA_rad) * PA_err_rad)**2)
+
+        # combine with existing measurement noise in quadrature
+        stokes_vector_errs[2*i] = np.sqrt(stokes_vector_errs[2*i]**2 + Q_ref_err**2)
+        stokes_vector_errs[2*i+1] = np.sqrt(stokes_vector_errs[2*i+1]**2 + U_ref_err**2)
```
maxwellmb: @john-livingston, do you mind writing out the algebra for why this is the correct approach. My first instinct is that it's slightly off, but I definitely could be wrong.

maxwellmb (changes requested):
Thanks for taking this on John, I appreciate it! 

I've got one question, see below.

corgidrp/pol.py
maxwellmb: (it could very well be that I'm being dense and that this is fine, but either way I'd like to understand it)

john-livingston: You're right - I was treating reference uncertainty as noise on `b`, which is only correct when the MM elements scaling those columns equal 1. I think this isn't catastrophic as it results in a slight overestimate of the errors, but computing it exactly turns out to be fairly trivial since `mueller_elements` is already available post-solve. Also including cov(Q,U) within each standard star since Q and U from the same catalog entry can be correlated. Pushing the fix now.

maxwellmb (approved):

PR #848, opened 2026-04-17
fixed legacy frame filter

kjl0025:
## Describe your changes
Rejection of legacy frames for trap-pumping was incorrect; it ignored bias and read noise when checking the frame values (due to a misunderstanding on my part about how legacy frames were actually made).  Fixed now so that if the mean of a 300x300 region in the center of the imaging area has a mean of 0 to within 10DN, reject the frame.  The code runs fine even without this "fix"; the fix just allows for more unnecessary frames to be rejected and reduce the amount of memory used.  At most, without this fix, at most 1200 1037x1056 frames would be held in memory and processed in the trap-pump function, whereas after this fix, at most 300 1037x1056 frames would be held in memory.  About 482 imaging-area frames is 100GB.  Tagging @mygouf for cognizance. 

## Type of change

N/A

### DIFF
--- a/corgidrp/pump_trap_calibration.py
+++ b/corgidrp/pump_trap_calibration.py
@@ -2380,7 +2380,8 @@ def tpump_analysis(input_dataset, time_head = 'TPTAU',
                 # Check if mean of a central 300x300 region is 0 (since read noise has mean of 0 and bias has already been subtracted).  If so, reject since it's a legacy frame.
                 # Don't reject it at beginning of recipe, though, since they may be used once TPUMP CAR is finalized.
                 #if np.abs(frame_data[0:1027]).max()==0:
-                central_mean_value = np.mean(frame_data[frame_data.shape[0]/2-150:frame_data.shape[0]/2+150, frame_data.shape[1]/2-150:frame_data.shape[1]/2+150])
+                halfway_row, halfway_col = int(frame_data.shape[0]/2), int(frame_data.shape[1]/2)
+                central_mean_value = np.mean(frame_data[halfway_row-150:halfway_row+150, halfway_col-150:halfway_col+150])
                 if np.abs(central_mean_value) < 10: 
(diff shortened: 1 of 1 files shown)


corgidrp/pump_trap_calibration.py
```diff
@@ -2376,9 +2376,14 @@ def tpump_analysis(input_dataset, time_head = 'TPTAU',
...
-                # ignore legacy frames, which are 0 except for rows 1027 through 1037 in imaging area (rows 0-1037, cols 1088-2144)
-                if np.abs(frame_data[0:1027]).max()==0:
-                    continue 
+                # ignore legacy frames, which have no illumination (just bias and read noise) except for rows 1027 through 1037 in imaging area (rows 0-1037, cols 1088-2144)
+                # Check if mean of a central 300x300 region is 0 (since read noise has mean of 0 and bias has already been subtracted).  If so, reject since it's a legacy frame.
+                # Don't reject it at beginning of recipe, though, since they may be used once TPUMP CAR is finalized.
+                #if np.abs(frame_data[0:1027]).max()==0:
+                halfway_row, halfway_col = int(frame_data.shape[0]/2), int(frame_data.shape[1]/2)
+                central_mean_value = np.mean(frame_data[halfway_row-150:halfway_row+150, halfway_col-150:halfway_col+150])
+                if np.abs(central_mean_value) < 10:
```
semaphoreP: should we make this threshold an optional keyword that defaults to 10? or is this robust enough that we don't think we'd ever tweak it.

semaphoreP (commented):
just one question. otherwise sounds reasonable.

corgidrp/pump_trap_calibration.py
kjl0025: Good idea.  I just pushed a change implementing that.

semaphoreP (approved):

PR #859, opened 2026-04-28
Improve spec_psf_subtraction routine

### LINKED ISSUE #683: Improve Spectroscopy PSF subtraction e2e output
The output of the L3->L4 spectrscopy PSF subtraction e2e test looks weird. See below. This doesn't affect the data format, but the values certainly could be improved.

Image

There's two possible things to investigate:
 * I think we may not be cropping the image properly in L2b->L3. This may be due to incorrect header keyword value for the star position
 * The PSF subtraction routine could use some more investigation and improvement.

### LINKED ISSUE #723: Improve the spectroscopy PSF subtraction routine
The output of the L3->L4 spectrscopy PSF subtraction e2e test looks weird. See below. This doesn't affect the data format, but the values certainly could be improved.

Image

Cropping issue is fixed (#683)

 * Improvements in the PSF subtraction routine needs more detailed refstar/targetstar sims

aneeshb97:
Algorithmic improvements to the spectroscopy psf subtraction routine

## Type of change

### DIFF
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -2,6 +2,7 @@
 import warnings
 import numpy as np
 import os
+import scipy.stats
 import pyklip.rdi
 from pyklip.klip import rotate, collapse_data
 import scipy.ndimage
@@ -1728,9 +1729,46 @@ def spec_psf_subtraction(input_dataset):
         # rescale wavelengh bands to match
         ref_col_mean = np.mean(shifted_ref,axis=1)
         ref_col_mean[ref_col_mean==0] = 1 # prevent div by 0
-        scale = np.mean(frame.data,axis=1)/ref_col_mean
-        shifted_scaled_ref = shifted_ref*scale[:,None]
 
+        frame_datcopy = np.copy(frame.data)
+
+        # Masking 6 pixels about planet location for dataset copy
+        planet_pos = int(frame.ext_hdr['WV0_X'])
+        frame_datcopy[:,planet_pos-6:planet_pos+6] = np.nan
+
+        # First, find slice where peak of ref star psf lies at each wav. Next, compute mode to find "best" slice. Then, calculate mean for 10-pixel centered on best slice for each wav for both ref and science.
+        # This mean is used to scale down ref star psf
+        row_peak_arr = np.argmax(shifted_ref, axis=1)
+        best_peak = int(scipy.stats.mode(row_peak_arr).mode)
+        if best_peak < 5 or best_peak > frame_datcopy.shape[1]-5:  ##Default to center of image if peak at edges and give warning
+            warnings.warn("Ref star PSF peak at edge of image. Please manually verify if reference is shifted properly.")
+            best_peak = int(frame.data.shape[1]/2)
+        avg_peak_ref = np.nanmean(shifted_ref[:,best_peak-5:best_peak+5],axis=1)
+        avg_peak_dat = np.nanmean(frame_datcopy[:,best_peak-5:best_peak+5],axis=1)
+
+        ##Find background median and std deviation from first 10 rows/wvs at top and bottom edge of image
+        bg_std = np.std(np.concatenate((avg_peak_ref[:10],avg_peak_ref[-10:])))
+        bg_median = np.median(np.concatenate((avg_peak_ref[:10],avg_peak_ref[-10:])))
+
+   
+        ##Use boxcar mean to identify wvs with the ref psf
+        cumulatsum = np.cumsum(np.insert(avg_peak_ref, 0, 0))
+        boxcar_mean = (cumulatsum[5:] - cumulatsum[:-5]) / 5
+        start = np.where(boxcar_mean > (bg_median+5 * bg_std))[0][0]; end = np.where(boxcar_mean > (bg_median+5 * bg_std))[0][-1]
+
+
+        # Fit a 5th order function to mean across wavelengths to scale down ref star psf. Only do this for wvs identified above
+        pixel_arr = np.arange(start,end)
+        polyfn_dat = np.polyfit(pixel_arr,avg_peak_dat[start:end],deg=5)
+        polyarr_dat = np.polyval(polyfn_dat, pixel_arr)
+
+        polyfn_ref = np.polyfit(pixel_arr,avg_peak_ref[start:end],deg=5)
+        polyarr_ref = np.polyval(polyfn_ref, pixel_arr)
+
+        scale = np.nanmean(frame_datcopy,axis=1)/ref_col_mean
+        scale[start:end] = polyarr_dat/polyarr_ref
+        shifted_scaled_ref = shifted_ref*scale[:,None]
+    
         shifted_refdq = np.roll(mean_ref.dq, (shift[0], shift[1]), axis=(0,1))
         # at this point in the pipeline, the err is mainly shot noise, so multiplying the err is appropriate
         # shifting may throw off err at the edges of the frame, but those pixels aren't used anyways
@@ -1740,12 +1778,16 @@ def spec_psf_subtraction(input_dataset):
         orig_frame = frame.data.copy()
         frame.data -= shifted_scaled_ref
 
+        # subtract reference from dataset copy 
+        frame_datcopy -= shifted_scaled_ref
+        frame_datcopy[:,planet_pos-6:planet_pos+6] = frame.data[:,planet_pos-6:planet_pos+6]
+
         # determine the throughput at the estimated source position
         # This is a rough guess at the throughput. Want to make this more accurate
         with warnings.catch_warnings():
             # catch divide by zero warnings
             warnings.filterwarnings('ignore', category=RuntimeWarning)
-            through = 1 - np.nansum(frame.data * shifted_scaled_ref, axis=1)/np.nansum(shifted_scaled_ref * shifted_scaled_ref, axis=1)**0.5/np.nansum(frame.data*frame.data, axis=1)**0.5
+            through = spec_throughput(orig_frame, shifted_ref, start_row=start, end_row=end)
         # Save algorithm throughput as an extension on the psf-subtracted Image
         frame.add_extension_hdu('ALGO_THRU', data = np.array(through), header = algothru_hdr)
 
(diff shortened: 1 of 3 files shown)


semaphoreP:
@aneeshb97 you can run `flake8 . --count --select=E9,F63,F7,F82,DCO020,DCO021,DCO022,DCO023,DCO024,DCO030,DCO031,DCO032,DCO060,DCO061,DCO062,DCO063,DCO064,DCO065 --show-source --statistics` locally to see if things are ok (see README for more details). Docstring formats are specific so that they can be parsed and turned into documentation pages automatically.

corgidrp/l3_to_l4.py
```diff
@@ -1728,9 +1729,50 @@ def spec_psf_subtraction(input_dataset):
...
         ref_col_mean[ref_col_mean==0] = 1 # prevent div by 0
-        scale = np.mean(frame.data,axis=1)/ref_col_mean
-        shifted_scaled_ref = shifted_ref*scale[:,None]
 
+        frame_datcopy = np.copy(frame.data)
+        # First, find slice where peak of ref star psf lies at each wav. Next, compute mode to find "best" slice.
+        row_peak_arr = np.argmax(shifted_ref, axis=1)
+        best_peak = int(scipy.stats.mode(row_peak_arr).mode)
+        if best_peak < 5 or best_peak > frame_datcopy.shape[1]-5:  ##Default to center of image if peak at edges and give warning
+            warnings.warn("Ref star PSF peak at edge of image. Please manually verify if reference is shifted properly.")
```
semaphoreP: this warning is not super accurate, because the code implicitly ignoring it and assuming it's at the middle of the image. I think we should do two things:

1) update the warning to indicate this
2) have an optional keyword specifying the mask size of the PSF. you have hard-coded 5's and 7's in this function. It's best practice to make these keywords so we don't implicitly assume the size of the PSF.

corgidrp/l3_to_l4.py
```diff
@@ -1728,9 +1729,50 @@ def spec_psf_subtraction(input_dataset):
         # rescale wavelengh bands to match
         ref_col_mean = np.mean(shifted_ref,axis=1)
         ref_col_mean[ref_col_mean==0] = 1 # prevent div by 0
-        scale = np.mean(frame.data,axis=1)/ref_col_mean
-        shifted_scaled_ref = shifted_ref*scale[:,None]
 
+        frame_datcopy = np.copy(frame.data)
+        # First, find slice where peak of ref star psf lies at each wav. Next, compute mode to find "best" slice.
```
semaphoreP: I don't understand why this is important. the reference star image is an image of the reference star's speckle filed (no on-axis ref star light). Why is it important to find the peak speckle flux location? Why not just sum up all the speckle flux across all columns?

corgidrp/l3_to_l4.py
```diff
@@ -1728,9 +1729,50 @@ def spec_psf_subtraction(input_dataset):
...
+
+        ##Find background median and std deviation from first 10 rows/wvs at top and bottom edge of image
+        bg_std = np.std(np.concatenate((avg_peak_ref[:10],avg_peak_ref[-10:])))
+        bg_median = np.median(np.concatenate((avg_peak_ref[:10],avg_peak_ref[-10:])))
+
+   
+        ##Use boxcar mean of row-by-row avg to identify wvs with the ref psf
+        cumulatsum = np.cumsum(np.insert(avg_peak_ref, 0, 0))
+        boxcar_mean = (cumulatsum[5:] - cumulatsum[:-5]) / 5
+        start = np.where(boxcar_mean > (bg_median+5 * bg_std))[0][0]; end = np.where(boxcar_mean > (bg_median+5 * bg_std))[0][-1]
```
semaphoreP: the background threshold 5 * std should be a optional keyword so that it can be an adjustable free parameter. for example, if we observe a really faint star, the speckles may not be 5x the background flux.

semaphoreP: Please also define start and end on different lines. it makes the code more readable.

semaphoreP: Also, what if start/end are empty lists, or start=end? should we fall back to using the entire frame (start =0 , end = image_size?)

corgidrp/l3_to_l4.py
```diff
@@ -1740,12 +1782,17 @@ def spec_psf_subtraction(input_dataset):
         orig_frame = frame.data.copy()
         frame.data -= shifted_scaled_ref
 
+        # subtract reference from dataset copy 
+        frame_datcopy -= shifted_scaled_ref
+        if mask_planet: #Unmasking if we masked before
+            frame_datcopy[:,planet_pos-5:planet_pos+5] = frame.data[:,planet_pos-5:planet_pos+5]
```
semaphoreP: does `frame_datcopy` get used for something? I don't think it gets saved into out_dataset.

corgidrp/l3_to_l4.py
```diff
@@ -1770,6 +1817,64 @@ def spec_psf_subtraction(input_dataset):
...
+        shifted_ref (corgidrp.data.Image): Shifted reference frame from spec_psf_subtraction.
+        start_row (int, optional): Start row location of ref star PSF.
+        end_row (int, optional): End row location of ref star PSF.
+        mask_planet (bool, optional): Mask 10 pixels about planet location. Do this only if planet psf is sufficiently separated (>7 pixels) from peak of reference psf.
+
+    Returns:
+        Algorithmic throughput as a function of wavelength.
+    '''
+
+    amplitude = 0.02; sigma = 1.5; x0 = 50
```
semaphoreP: these should be free parameters. presumably the amplitude should adapt to the local noise level? it shouldn't be fixed. similarly, the location x0 needs to depend on where the planet is not located?

corgidrp/l3_to_l4.py
```diff
@@ -1770,6 +1817,64 @@ def spec_psf_subtraction(input_dataset):
...
+    orig_frame[:,45:55] = orig_frame[:,45:55] + gaussian_planet
+    inj_planet = np.ones(orig_frame.shape[0]) * np.max(orig_frame[:,45:55],axis=1) #np.trapz(gaussian_planet[3:7],x[3:7])
+    
+    frame_w_injectedplanet = np.copy(orig_frame)
+    if mask_planet:
+        orig_frame[:,45:55] = np.nan
+
+    row_peak_arr = np.argmax(shifted_ref, axis=1)
+    best_peak = int(scipy.stats.mode(row_peak_arr).mode)
+    avg_peak_ref = np.nanmean(shifted_ref[:,best_peak-5:best_peak+5],axis=1)
```
semaphoreP: any time you have hard coded numbers, we should consider making them optional input keywords.

corgidrp/l3_to_l4.py
```diff
@@ -1770,6 +1817,64 @@ def spec_psf_subtraction(input_dataset):
...
+    # make same 
+    orig_frame-= shifted_scaled_ref
+    if mask_planet:
+        orig_frame[:,45:55] = frame_w_injectedplanet[:,45:55]
+    postimg_signal = np.max(orig_frame[:,45:55],axis=1) #np.trapz(orig_frame[:,48:52],x[3:7])
+        
+    orig_frame[:,45:55] = orig_frame[:,45:55] - gaussian_planet
+    speckle_std = np.nanstd(orig_frame[:,45:55],axis=1)
+
+    through = (postimg_signal-speckle_std)/inj_planet
```
semaphoreP: why do you subtract the standard deviation? 

Also did you check that your throughput values you get make sense?

semaphoreP (commented):
Some comments about making the step function more generalizable/flexible, and a few questions on the implementation.

semaphoreP:
note to self: check data format e2e test again

corgidrp/l3_to_l4.py
```diff
@@ -1728,9 +1731,63 @@ def spec_psf_subtraction(input_dataset):
...
+            raise Exception('Background threshold is too high. Algorithm could not identify rows/wavelengths with reference psf')
+        if start==end:
+            raise Exception('Found only one row/wavelength with reference psf. Please check your background threshold.')
+
+        # Find slice where peak of ref star psf lies between start and end wavelengths. Next, compute mode to find "best" slice. 
+        # Reference was masked earlier to ensure chosen slice is not on same slices as masked planet psf.
+        row_peak_arr = np.nanargmax(ref_copy[start:end,:], axis=1)
+        best_peak = int(scipy.stats.mode(row_peak_arr).mode)
+        if best_peak < 5 or best_peak > frame_datcopy.shape[1]-5:  #If reference star speckles peak at edge of image, raise exception
+            raise Exception("Ref star PSF peak at edge of image. Please manually verify if reference is shifted properly.")
```
semaphoreP: what happens if we verify that reference is shifted properly? This error would still happen? Should this be a warning, or a way to force the code to keep going even if this is true (to skip the exception)?

corgidrp/l3_to_l4.py
```diff
@@ -1770,6 +1847,66 @@ def spec_psf_subtraction(input_dataset):
...
+        start_row (int): Start row location of ref star speckles.
+        end_row (int): End row location of ref star speckles.
+        amplitude (float, optional): Amplitude of planet psf.
+        mask_pl_pixels (int, optional): No. of pixels to mask on either side of planet position.
+
+    Returns:
+        Algorithmic throughput as a function of wavelength.
+    '''
+
+    fwhm = 1.5 #mask_pl_pixels/2.
```
semaphoreP: did you mean to set this to mask_pl_pixels/2?

semaphoreP (commented):
just a couple small things. otherwise looks good.

corgidrp/l3_to_l4.py
aneeshb97: I am unsure about the best value. I'd like to mask at least a 2sigma region around the peak planet flux. Hence, the logic was sigma=mask_pl_pixels/2 (it should be sigma btw not fwhm, I'll fix that). Since mask_pl_pixels is pixels masked on either side, masked region = 2 x mask_pl_pixels = 4 x sigma = 1.7 x fwhm

But I can see an argument for masking a 3sigma region, in which case sigma=mask_pl_pixels/3. In that case, masked region = 2 x mask_pl_pixels = 6 x sigma = 2.5 x fwhm

semaphoreP: when in doubt, add more optional free parameters to the functions so we can pass this through and tweak it

corgidrp/l3_to_l4.py
aneeshb97: This exception means that brightest speckles are <5 columns away from the image edges. A possible idea is we look for the next brightest speckle col and so on, but the next brightest speckle col is probably just adjacent. We could still try that 1-2 times and see if that fixes stuff.

If delta(pixel) from edge is <5, we can just move the best_peak some pixels over such it is >5. Here, the issue is what if the planet is right next to the best_peak and the cols have been NaN'd, moving over might not aid with the subtraction. But it'll still run.

Another possibility is we decrease the number of pixels around the brightest peak we average over. But that comes with the issue of how good is a scaling that averages over just four pixels? For the last one especially, it'd be good to have a hard limit

semaphoreP: I think the error message indicates that if you come back after verification, it would still run. So either make it clear we absolutely can't run the code if something is 5 pixels from the edge, or add an override keyword that does what you suggest (move it over so the best_peak is >= 5 from the edge).

semaphoreP (approved):

PR #860, opened 2026-04-29
Enable dark subtraction on dataset with multiple detector settings

### LINKED ISSUE #803: handle multiple detector configurations for dark subtraction
Dark subtraction currently only supports a single EXCAM configurations. We should be able to pass multiple EMGAIN and Exposure times and dark subtraction (with detector noise maps) should be able to handle it. 

In dark subtraction, split dataset by EMGAIN and exposure time, and create synthetic darks for each combination of emgain and exposure times.

semaphoreP:
## Describe your changes
Enables dark subtraction with detector noise map on dataset with mutiple settings (exptime, gain). Saves each synthetic dark using the input science dataset filenames as the base (and change ending to `drk_cal.fits`). This naming convention is to ensure each synthetic dark as a unique filename. 

## Type of change

Closes #803 

### DIFF
--- a/tests/test_dark_sub.py
+++ b/tests/test_dark_sub.py
@@ -185,5 +185,71 @@ def test_dark_sub():
 
     corgidrp.track_individual_errors = old_err_tracking
 
+def test_dark_sub_multi_config():
+    """
+    Verify that dark_subtraction handles a dataset with multiple EMGAIN/EXPTIME
+    configurations when noisemaps is provided, and that the traditional-dark path
+    still rejects multi-config input.
+    """
+    calibdir = os.path.join(os.path.dirname(__file__), "testcalib")
+    if not os.path.exists(calibdir):
+        os.mkdir(calibdir)
+
+    configs = [
+        {"EMGAIN_C": 10, "EXPTIME": 4},
+        {"EMGAIN_C": 20, "EXPTIME": 8},
+    ]
+
+    frames = []
+    for cfg in configs:
+        em_gain = cfg["EMGAIN_C"]
+        exptime = cfg["EXPTIME"]
+        # Frame data equal to the synthesized dark so the result should be ~0
+        frame_data = (noise_maps.FPN_map + noise_maps.CIC_map * em_gain + noise_maps.DC_map * exptime * em_gain) / em_gain
+        prihdr, exthdr, errhdr, dqhdr, _ = mocks.create_default_L2a_headers()
+        img = data.Image(frame_data, pri_hdr=prihdr, ext_hdr=exthdr, err_hdr=errhdr, dq_hdr=dqhdr)
+        img.ext_hdr['EMGAIN_C'] = em_gain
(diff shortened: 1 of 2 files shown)


semaphoreP:
e2e tests pass

semaphoreP:
@ecady-jpl do you guys use the dark_subtraction() function in the DRP for FSW, and if so, do you care which filename is used as the base of the filename for the drk_cal.fits file?

ecady-jpl:
We don't call it directly in the ground software, no, if that's what you're asking.  For the flight software, we need the three synthetic dark components only as inputs to a tool to create the binaries for uplink.  I don't have strong feelings about the filename as long as I can find it unambiguously.

tests/test_dark_sub.py
```diff
@@ -185,5 +185,71 @@ def test_dark_sub():
...
+    assert saved_configs == expected_configs
+
+    # traditional dark path: must still raise for multi-config datasets
+    dark_dataset = mocks.create_dark_calib_files(filedir=calibdir)
+    detector_params_local = data.DetectorParams({})
+    from corgidrp.darks import build_trad_dark
+    traditional_dark = build_trad_dark(dark_dataset, detector_params_local, full_frame=True)
+    with pytest.raises(Exception):
+        l2a_to_l2b.dark_subtraction(multi_config_dataset, dark=traditional_dark, outputdir=calibdir)
+
```
maxwellmb, line 251: Should we delete `darkout` at the end?

maxwellmb (commented):
Overall this looks good to me and I think is a high priority to push though

I want to ping @mygouf and @ecady-jpl to get approval since this has implications for DIP processing and the default behavior of the pipeline <l2b.

tests/test_dark_sub.py
semaphoreP: generally in the tests, we clear out the output folders at the start of the test run, rather than the end. This has the slight benefit that we can always look back at output of tests, at the cost of some extra mess of folders in the testing dir.

semaphoreP:
I note that I don't think this changes any previous behavior except that dark processing now works for multiple detector settings whereas in the past it threw an exception.

ecady-jpl:
What are the implications for DIP processing?

maxwellmb:
> What are the implications for DIP processing?

Well, I have no idea how DIP works - never seen it. But the change in operations would be that you can now pass in more files at once to get processed L2a-L2b in one go, since you don't have to split them up by detector integration parameters. It also means that more than one dark could be produced for a given input dataset.

semaphoreP:
I think we can say the changes in general should be safe for DIP processing, as it expands the pipeline but doesn't break previous behavior. 

The only implication I could think of is the question I asked Eric above: that the synthetic dark created could have a different filename than before (still has the _drk_cal.fits, but is based off different filenames depending on the detector config).

maxwellmb (approved):
Since this only adds functionality, instead of breaking things, then I think I'm going to go ahead and approve it.

PR #866, opened 2026-05-01
Issue 799: add default calibrations

### LINKED ISSUE #799: Package a suite of default calibrations for corgidrp
In commissioning, we probably won't have calibrations taken for everything but we want to start processing data to l2b. We should provide a default suite of calibrations that are generated at the beginning (like how things like DetectorParams are initialized). 

Using a combintion of TVAC calibrations and "empty" calibrations (e.g., a flat field of all 1's for all the different flat field modes), I think we should be able to generate default calibration products for everything needed to do up to l2b processing. We should package some TVAC calibrations into corgidrp, and use the same way as the e2e tests to turn them into calibrations following corgidrp format.

semaphoreP:
## Describe your changes

- Added raw TVAC calibration source files to corgidrp/data/default_calibs/ (tracked via Git LFS): nonlinearity table, fixed-pattern noise, CIC, dark current, and bad pixel map
- caldb.initialize() now auto-generates a full suite of default calibration products from those source files on first import:
    - NonLinearityCalibration — from TVAC nonlinearity lookup table (March 2024)
    - KGain — 8.7 e/DN with TVAC read noise (RN=121.76 e-, RN_ERR=2.0 e-)
    - DetectorNoiseMaps — FPN, CIC, and dark current maps embedded in the full detector frame
    - FlatField (IMAGING) — all-ones array with FPAMNAME=OPEN_12, CFAMNAME=1F, DPAMNAME=IMAGING
    - FlatField (POL0 and POL45) — all-ones arrays for default polarimetric observations
    - BadPixelMap — from TVAC bad pixel data
- All synthetic default calibrations are stamped with MJDSRT=2026-01-01 to distinguish them from real instrument data
- Updated MANIFEST.in and .gitattributes to include the new packaged files
- Added test_l1_to_l2b_default_calibs in tests/test_walker.py: verifies that L1→L2b pipeline processing completes successfully using only the packaged default calibrations, without any user-supplied calibration files
- Fixed test_default_calibs to clean up a stale test caldb file at the start, preventing false failures when an earlier test leaves it behind

## Type of change

CLoses #799 


semaphoreP:
e2e tests pass

maxwellmb (commented):
This probably won't hold up implementing this PR, but since the flats are just 1s, could we perhaps have the walker automatically generate a default flat on the fly if it's not there? Part of my thinking is that currently we're selecting flats on a ton of different header keywords, so the flat here only fixes one case. Albeit it's likely the the most important case, so I'm keen to push this through, but I wanted to bring this up for discussion.

semaphoreP:
all e2e tests pass, so let me know if there's anything else @maxwellmb

maxwellmb (approved):

PR #869, opened 2026-05-04
Fix CT cal FPAMNAME corruption when pupil image is first in dataset

### LINKED ISSUE #624: File order matters when creating CT cal, but it shouldn't
**Describe the bug**
If you pass a Dataset in to `corethroughput.generate_ct_cal()` that is ordered such that the pupil image is first, then it gets saved to the caldb with a 'FPAMNAME'='OPEN_12' which is not what we want since we select for calibrations based on the FPAMNAME. 

**To Reproduce**
Run the l3_to_l4_pol_e2e.py test but with the order of the CT input Images swapped. Then you'll see that no CT file is selected. 

**Expected behavior**
We should make sure that the FPAMNAME isn't based on the pupil image. 

**corgidrp version:**
develop version as of Nov. 4 2025.

semaphoreP:
## Describe your changes
In generate_psf_cube(), the PSF cube header was always taken from dataset[0], which could be a pupil image with FPAMNAME='OPEN_12'. This caused the saved CoreThroughputCalibration to carry an incorrect FPAMNAME, preventing caldb selection by FPM name.

Fix: scan the dataset for the first off-axis (non-pupil) frame and use its ext_hdr/dq_hdr for the PSF cube, so the calibration file header reflects the actual FPM used during CT observations regardless of input frame ordering.

Adds test_ct_cal_order_independent() to guard against regression.

## Type of change

Closes #624

### DIFF
--- a/corgidrp/corethroughput.py
+++ b/corgidrp/corethroughput.py
@@ -300,13 +300,27 @@ def generate_psf_cube(
         raise Exception(('The number of PSFs does not match the number of PSF '+
             ' locations.'))
 
-    # PSF cube header
-    ext_hdr = dataset[0].ext_hdr
+    # PSF cube header - use first off-axis frame so pupil image headers
+    # (e.g. FPAMNAME='OPEN_12') do not propagate into the calibration file
+    first_offaxis_frame = None
+    for frame in dataset:
+        try:
+            exthd = frame.ext_hdr
+            if not (exthd['DPAMNAME'] == 'PUPIL' and exthd['LSAMNAME'] == 'OPEN' and
+                    exthd['FSAMNAME'] == 'OPEN' and exthd['FPAMNAME'] == 'OPEN_12'):
+                first_offaxis_frame = frame
+                break
+        except KeyError:
+            first_offaxis_frame = frame
+            break
+    if first_offaxis_frame is None:
+        raise Exception('No off-axis PSF frame found in dataset')
+    ext_hdr = first_offaxis_frame.ext_hdr
     # Add EXTNAME
     psf_hdu = fits.ImageHDU(data=psf_cube, header=ext_hdr, name='PSFCUBE')
 
     # Data quality cube
-    dq_hdr = dataset[0].dq_hdr
+    dq_hdr = first_offaxis_frame.dq_hdr
     # Add specific information
     dq_hdr['COMMENT'] = 'Data quality for each image' 
     # Add EXTNAME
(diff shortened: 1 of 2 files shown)


semaphoreP:
I can confirm that the l3_to_l4_pol_e2e.py e2e passes even if I swap and put the pupil image first now. Lines 139-140 of l3_to_l4_pol_e2e.py:
```
data_ct_interp = [data.Image(pupil_image,pri_hdr = prhd,
        ext_hdr = exthd_pupil, err = err), ct_dataset0]
```

tests/test_corethroughput.py
```diff
@@ -961,6 +961,45 @@ def test_psf_interp():
...
+    for frame in dataset_ct_syn:
+        try:
+            exthd = frame.ext_hdr
+            if (exthd['DPAMNAME'] == 'PUPIL' and exthd['LSAMNAME'] == 'OPEN' and
+                    exthd['FSAMNAME'] == 'OPEN' and exthd['FPAMNAME'] == 'OPEN_12'):
+                pupil_frames.append(frame)
+            else:
+                offaxis_frames.append(frame)
+        except KeyError:
+            offaxis_frames.append(frame)
```
maxwellmb, line 981: same question: what's up with the KeyError?

corgidrp/corethroughput.py
```diff
@@ -300,13 +300,27 @@ def generate_psf_cube(
...
+    for frame in dataset:
+        try:
+            exthd = frame.ext_hdr
+            if not (exthd['DPAMNAME'] == 'PUPIL' and exthd['LSAMNAME'] == 'OPEN' and
+                    exthd['FSAMNAME'] == 'OPEN' and exthd['FPAMNAME'] == 'OPEN_12'):
+                first_offaxis_frame = frame
+                break
+        except KeyError:
+            first_offaxis_frame = frame
+            break
```
maxwellmb, line 310: what is this Key error about? If there's no _PAMNAME I'd think we might want to error out instead of continuing on with this arbitrary frame, no?

maxwellmb (changes requested):

corgidrp/corethroughput.py
semaphoreP: I copied it from code blocks further above in the same code when determining what frames are pupil imaging. Upon closer inspection, I agree let's just throw an error instead of continuing, since it's very ambiguous.

maxwellmb (approved):

PR #873, opened 2026-05-06
fix how k gain/nonlin handles cosmic rays, fix effective FWC used for cosmic masking, no cosmic tails for EM gain =1

### LINKED ISSUE #847: corgidrp default CR threshold below the EETC threshold
The corgidrp default saturation threshold (0.7*Full Well) is different than the EETC's goal threshold (0.75*Full Well). Using this issue to record the conversation. 

TBD if any action is needed.

### LINKED ISSUE #867: False Positives in Cosmic Ray Flagging/Constant Saturation DN Bug?
**Describe the bug**
When processing corgisim-simulated HLC-NFOV coronagraphic datasets with EETC-driven EXCAM settings, the diffracted Arago spot from the simulated images will sometimes exceed the saturation threshold and be flagged as a cosmic ray that later gets masked in L2a to L2b (only consistently occurs for reference star images). In checking the DN of the L2a data and the saturation DN populated in the L2a header, it makes sense why the non-satellite spot image Arago spot is flagged (Arago spot is similar flux to 1e-7 satellite spots) as the EXCAM settings are only based on the speckle field brightness rather than the satellite spots.

Investigating this further, I found that the calculated Saturation DN populated in the L2a header was the same across every simulated image even when the EXCAM settings were different. I would expect saturation DN to change with EXCAM EM gain.

**To Reproduce**
Pass in an L1 dataset for processing through to L2b or L3, with different EM gain and EXPTIME settings for different flavors of the dataset (ref star with sat spots, ref star with no spots, sci star with sat spots, sci star with no spots). For sufficiently bright target stars and EXCAM settings optimized for just a high-contrast speckle field (3e-8 or better, likely would not see this issue if optimized for a 1e-7 speckle field), the central Arago spot will be flagged as a cosmic ray. The calculated Saturation DN would also be the same across all frames.

**Expected behavior**
More careful treatment of flagging cosmics may be needed, such as changing saturation thresholds. Ideally, false positives would not be flagged by the DRP. I am also uncertain about whether or not the saturation DN should vary as a function of frame EM gain.

**corgidrp version:**
 - 4.0

**Additional context**
These plots show representative L2a and L2b frames of a V=1.69 reference star observation (with and without satellite spots), with different EM gain settings and exposure times. The saturation DN is extracted from the L2a image headers.

Image

kjl0025:
## Describe your changes
The recently generated pupil plane images generated with corgisim used for k gain/nonlin calibration led to gross disagreement with what was simulated when processed through the DRP.  This was because cosmic ray flux at the level of what's expected in flight had not been tested before.  With the changes here, the DRP gives good results for the k gain and nonlin calibrations using the CAR-121 corgisim data.

--This PR fixes how k gain/nonlin calibration handles cosmic rays.  For k gain cal, EM gain is 1, and cosmic rays do not have a tail b/c the gain register is not used, so cosm_tail is set to 0 in this case, regardless of the input value of cosm_tail.  Also, the effective full-well capacity (FWC) used for cosmic ray detection was set to the FWC for the gain register, so now the effective FWC is actually used by the detection function.  
-- In both the k gain and nonlin calibration functions, frames are mean-combined with the DQs accounted for.  (Before, each frame's mask was applied as NaNs to the frame itself.)
-- Several pupil images called for by CAR-121 have almost the whole pupil image area saturated, and such frames corrupt the calibration results for nonlin.  A check is now so that an averaged frame is only used if the number of unmasked pixels in the pupil image area is percent_pupil * total number of pupil area pixels.  Default percent_pupil value is 0.9.  This allows for a good linear response fit to be made so that nonlinearity can be characterized correctly with that as a reference.  This gave better agreement with what was simulated even though extrapolation had to be used in some regimes of EM gain - DN parameter space.
-- sat_thresh is now 0.99, which allows for more nonlinearity data to be usable.  Cosmic rays in flight will most likely saturate completely, so 0.99 should be a safe default.  (We saw lower-energy cosmic rays in TVAC data which required sat_thresh = 0.7 most likely because the rays were scattered once before hitting a pixel.)
--DRP DetectorParams were updated to CBE values.

## Type of change

#867, #847,


corgidrp/l1_to_l2a.py
```diff
@@ -260,22 +264,25 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
+    # Mask 2:  captures cosmic rays.  If EM gain is 1, no cosmic tails made since 
+    # those are only made in gain register.  
     # Do a for loop since it's calling a for loop in the sub-routine anyway
     # and can't handle different 'FWC_EM's for different frames.
     m2 = np.zeros_like(crmasked_cube)
 
     for i in range(len(crmasked_cube)):
         arrtype = crmasked_dataset.frames[i].ext_hdr['ARRTYPE']
+        if emgain_list[i] == 1:
+            cosm_tail_i = 0
```
semaphoreP, line 282: there should be a notification or print statement that we're overriding the user provided value

corgidrp/calibrate_nonlin.py
```diff
@@ -597,20 +579,32 @@ def calibrate_nonlin(dataset_nl,
...
-                            frame_1 = image_1.data
-                            if apply_dq: # in ram_heavy case, this hasn't been applied yet
-                                bad = np.where(image_1.dq > 0)
-                                frame_1[bad] = np.nan
-                        else:
-                            frame_1 = selected_files[iframe]
-                        frame_1 = frame_1.astype(np.float64)
-        
+                    if ram_heavy:
+                        frame_1, bp, _, _ = mean_combine(data.Dataset(selected_files), None)
```
semaphoreP, line 586: why don't we pass in bad pixel / dq maps in ram heavy mode? we have the DQ extension don't we?

corgidrp/calibrate_kgain.py
```diff
@@ -545,7 +554,8 @@ def calibrate_kgain(dataset_kgain,
         # split each frame up into bins, take std and mean of each region
         with warnings.catch_warnings():
             warnings.filterwarnings('ignore', category=RuntimeWarning)
-            mean_frames_mean_curr0 = np.nanmean(frames, axis=0)
+            #mean_frames_mean_curr0 = np.nanmean(frames, axis=0) 
+            mean_frames_mean_curr0, _, _, _ = mean_combine(frames, fr_bp_maps)
```
semaphoreP, line 558: could you explain how this is now different from running np.nanmean with the bad pixels masked as nans?

corgidrp/data.py
```diff
@@ -4277,6 +4277,38 @@ def get_bit_to_flag_map():
     """
     return {bit: name for name, bit in get_flag_to_bit_map().items()}
 
+def selective_dq(dq, val=32):
```
semaphoreP: is this used anywhere? seems like a orphan function

semaphoreP (commented):
A few comments, in addition to the nonlin e2e test still failing for me. Did you check that the nonline e2e works for you?

semaphoreP:
@ecady-jpl and @ajeldorado: do you use the DRP detect_cosmic_rays() function? If so, the default keyword values have changed to have a higher cutoff to identify saturated pixels. Additionally, the code to produce the kgain and nonlin calibration products have been upgraded to handle cosmic rays better, is there anything in particular you need to verify for that?

corgidrp/data.py
kjl0025: I did in an earlier attempt for nonlin cal, but I ended up not needing it.  However, it is kind of handy at teasing out masks for specific reasons, and it took a little bit of work, so I left it in there.  I you prefer I remove it, though, I can do that.

semaphoreP: I don't really like untested code lying around. I also would put it as a class function of the image class, rather than a standalone function. I'd at least add unit tests if you were to keep it. otherwise, we have your commits to reference in the future if we wanted to revive it.

ecady-jpl:
@semaphoreP We don't use the function directly at present, no

corgidrp/calibrate_kgain.py
kjl0025: In a previous iteration, I was selecting DQ flags only for cosmic rays, and I handled that within mean_combine, but since I ended up not using that, there is no real difference between namean and mean_combine here.

corgidrp/calibrate_nonlin.py
kjl0025: The DQs aren't loaded into memory in ram_heavy mode; the DQ for each frame in the input dataset is None. The DQs are loaded into memory inside the mean_combine function.

corgidrp/data.py
kjl0025: Sounds good.  I'll remove it.

kjl0025:
Illustration of improved accuracy of this branch over the current implementation:  Calibrating the corgisim data for CAR-121 (pupil plane image data), k gain calibration gave a value of ~0.09 while 8.3 was the simulated value.  Using this branch, a value of 7.9 was the result.  The nonlin calibration from this same datat resulted in factors of > 10 (in the thousands if I remember correctly?) while no nonlinearity was actually simulated, so all the values in the calibration should have been 1.  Using this branch, the max value from the resulting nonlin table was a little less than 1.08.

semaphoreP (approved):

Issue #856, opened 2026-04-23
Bad pixel map cal products get an incorrectly formatted filename


juliamilton:
**Describe the bug**
When creating a BadPixelMap from an input dataset, the VISITID header keyword gets set to "-999” instead of preserving a valid visit ID. This happens because:

- BadPixelMap creation in data.py calls merge_headers() with invalid_keywords=typical_cal_invalid_keywords
- VISITID is in the invalid list because Bad Pixel maps are made from a flat and a dark and have different VISITIDs
- merge_headers sets VISITID to -999. This leads to filenames like: cgi_-000000000000000999_20260410t1909415_bpm_cal.fits

We should figure out how to handle this and what the VISITID and/ or filename convention should be for bad pixel maps. 

This issue is related to #760

nasavbailey:
My understanding is that the convention was agreed to be last visit id. see https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/136121588/Bad+Pixel+Map+Calibration+Current+DRP+Development+Version   (related: issue #462 to include filename convention in readthedocs)

Is this a symptom of a more general problem with multivisit data processing? Will this occur with any recipe that combines data from multiple visits?

juliamilton:
To reproduce: run l1_to_l2b_e2e.py and look at the bad pixel map product that is created and saved in l1_to_l2b_e2e/calibrations.

This issue will likely affect other combined frames made from different visits. I think the fix is just to remove “VISITID” from the list of keywords that get invalidated if they are mismatched among frames. That is in line 23 of data.py .

clarissardoo:
This turned out to be a more general issue, to be addressed by #879

PR #881, opened 2026-05-11
Use latest input UTC timestamp for calibration filenames

### LINKED ISSUE #879: Calibration product filenames should use documented last VISITID and lasttimeUTC instead of input_dataset[-1]
**Describe the bug**

Several calibration/data product classes in data.py generate output filenames by modifying input_dataset[-1].filename. This assumes the final item in the input dataset list is also the latest input file by observation time. That is not always guaranteed. As a result the calibration product filenames may use the wrong visit ID or timestamp if the input dataset is not already sorted.

This came up while investigating Issue #856 . BadPixelMap headers can correctly have `VISITID = -999` because the dark and flat inputs may have different visit IDs.

**To Reproduce**
For bad pixel map: run l1_to_l2b_e2e.py and look at the bad pixel map product that is created and saved in l1_to_l2b_e2e/calibrations.
This issue will likely affect other combined frames made from different visits. 

**Expected behavior**
For each calibration/data product, filename generation should follow the documented filename convention for that product as shown [here](https://collaboration.ipac.caltech.edu/pages/viewpage.action?pageId=136121588&spaceKey=romancoronagraph&title=Bad%2BPixel%2BMap%2BCalibration%2BCurrent%2BDRP%2BDevelopment%2BVersion).

**Additional context**
Found upon investigation of BMP issue #856 .

clarissardoo:
## Describe your changes

Updates calibration product filename generation to use the chronologically latest input frame rather than relying on input dataset order. When all inputs have SCTSRT, the latest frame is selected by SCTSRT. If SCTSRT unavailable, the selection falls back to the TimeUTC timestamp in the CGI filename. The selected frame’s filename is then used as the base for the calibration product filename. If an input Image does not have .filename set, the filename is pulled from the primary header instead. For multivisit products (e.g. BPM), the VISITID header is also copied from the selected latest input frame so the output header stays consistent with the filename convention.

## Type of change

This addresses issue #879 , which was created after investigating #856 

- [] I have checked the output of the latest Github Actions run associated with this PR and confirmed running `pytest` did not produce any warnings


corgidrp/data.py
```diff
@@ -856,11 +856,28 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, input_dataset=N
...
+                    latest_frame=max(enumerate(input_dataset),key=lambda item:(str(item[1].ext_hdr['SCTSRT']), item[0]))[1]
+                else:
+                    # if not, use timestamp from filename
+                    latest_frame=max(enumerate(input_dataset),key=lambda item:(next((name.split('_')[2] for name in [item[1].filename, item[1].pri_hdr.get('FILENAME')] if name and len(name.split('_')) > 2), ''), item[0]))[1]
+                # use latest frame filename to rename
+                latest_filename=latest_frame.filename or latest_frame.pri_hdr['FILENAME']
+                # get visitid
+                if latest_frame.pri_hdr.get('VISITID') is not None:
+                    # assign it to output product
+                    self.pri_hdr['VISITID']=latest_frame.pri_hdr['VISITID']
```
maxwellmb: I don't think you need to update the VISITID here. There are other parts of the code that update this where appropriate.

maxwellmb: that's true everywhere else too

corgidrp/data.py
```diff
@@ -2252,6 +2403,56 @@ def __init__(self, data_or_filepath, err = None, pri_hdr=None, ext_hdr=None, err
         self.fluxcal_fac = self.data[0]
         self.fluxcal_err =  self.err[0]
 
+        # if this is a new FluxcalFactors file, we need to bookkeep it in the header
```
maxwellmb, line 2372: Are all these changes on purpose? They seem beyond the scope of the PR description

corgidrp/data.py
```diff
@@ -2294,10 +2495,13 @@ def __init__(self, data_or_filepath, err = None, dq = None, pri_hdr=None, ext_hd
             else:
                 # log all the data that went into making this calibration file
                 self._record_parent_filenames(input_dataset)
-                # give it a default filename using the first input file as the base
-                # strip off everything starting at .fits
-                orig_input_filename = input_dataset[-1].filename.split(".fits")[0]
-  
+                # get latest timestamp frame name
```
maxwellmb: why doesn't this one have the SCRSRT logic?

tests/test_corethroughput.py
```diff
@@ -403,8 +403,8 @@ def test_cal_file():
     os.mkdir(test_dir) 
     ct_cal_file_in.save(filedir=test_dir)
 
-    # Check that the filename is what we expect
-    ct_cal_filename = dataset_ct[-1].filename.replace("_l2b", "_ctp_cal")
+    # Check that the filename is what we expect (not using latest item in list anymore)
+    ct_cal_filename = ct_cal_file_in.filename
```
maxwellmb: This change now just makes it so that we're checking that a file was created based on the filename in the object that was saved. The previous test was testing that a file was created with the expected filename, based on the input dataset. Now it was doing that incorrectly, since we've established that the list file in the list isn't necessarily the last file in time (the whole purpose of this PR), but I don't think we want to give up the nature of that test, so I think it needs to be slightly re-written so that it's still checking for a file based on the _expected_ filename based on the input list.

maxwellmb (commented):
A few comments! Mostly that we don't need to be modifying VISITID in the header.

corgidrp/data.py
clarissardoo: I think I just accidentally changed the line of a previous comment! I didn't add this (see main/develop branch files and you'll find these there)

maxwellmb: Ok, then it'd be useful for you to do a new merge of develop into your branch please

clarissardoo: ok! will do

maxwellmb (approved):

PR #871, opened 2026-05-05
Use SCTSRT for frame ordering and fix MJD in mocks

semaphoreP:
## Describe your changes

- SCTSRT/SCTEND are no longer stripped after L1 — they propagate to L2a+ products, with SCTSRT taken from the first frame and SCTEND from the last in merged/combined data
- Frame ordering and duplicate detection in merge_headers(), sorting.py, calibrate_kgain.py, and calibrate_nonlin.py now use SCTSRT instead of DATETIME/MJDSRT
- create_default_L1_headers() now sets SCTSRT, SCTEND, MJDSRT, and MJDEND dynamically from the current time (consistent with DATETIME/FTIMEUTC), replacing static placeholder values from l1.csv
- Tests updated to set SCTSRT alongside DATETIME wherever frame ordering depends on unique timestamps (test_sort_pupilimg_frames, test_nonlin_cal, test_kgain_cal)
- Updates all data format docs with the SCTSRT and SCTEND keywords, since they were previously removed. 
- Fixed an issue with the spec l3-l4 e2e test where old files that weren't deleting were messing up the test. 
- Due to some e2e test failures, I also went ahead and updated spec l3->l4 coron processing recipe to use SpecFluxCal instead of AbsFluxCal and updated the corresponding spec_l3_to_l4_psfsub_e2e.py test. This will close #823 and #837 

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- New feature (non-breaking change which adds functionality)

## Reference any relevant issues (don't forget the #)
Closes #850 and #853 and #823 and #837

## Checklist before requesting a review
- [x] I have verified that all unit tests pass in a clean environment and added new unit tests, as appropriate
- [x] I have checked that I am merging into the right branch
- [ ] I have checked the output of the latest Github Actions run associated with this PR and confirmed running `pytest` did not produce any warnings
- [x] I have checked if my code modifies an existing step function or modifies the data format docs. If it does, I've added my PR to the [list maintained here](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/212028061/Log+of+PRs+that+could+affect+existing+VAP+Tests).

PR #884, opened 2026-05-11
Change walker.py due to satspot convention change

### LINKED ISSUE #845: Update walker to handle new sat spot CPGS XML convention
For spectroscopy mode, we need to know the separation (in [lam/D]) and clocking angle (in deg) of the satellite spot in the slit for calibration. This information is parsed from the CPGS XML. We have an update to how we will do this from the CPGS XML. Walker.py and the appropriate tests (esp. e2e test) needs to be updated to reflect the updated CPGS XML convention. 

There is a new text field called "satellite_spots_command_info" that has manually interested text that should now be used instead. 

_(Edited with the latest specifications based on discussion below and what is documented on confluence)_: 
We are expecting satellite spot information here entered manually with the convention `(sep1={value in lam/D}, angle1={value in degrees}, fr1={flux ratio in #E# scientific notation}, filt1={filtername})`. For example, `(sep=5.5, angle=45, f1=1.2e-6, filt=1F)`. Sometimes there will be more satellite spot pairs and the code we write should not fail at handling this: e.g., `(sep1=5, angle1=30, fr1=1E-7, filt1=1F) (sep2=5, angle2=120, fr2=1E-7, filt2=1F)`. The walker needs to be updated to parse satellite spot information this way from the CPGS XML.

aneeshb97:
## Describe your changes
Wrote a new function to parse satspot info according to the new CPGS convention. Original function will be replaced if change is approved.

Change to l3_to_l4 spec psfsub e2e to take new CPGS file. Unsure if spec noncoron e2e also needs to be changed, the latter currently does not use any CPGS XML files.

## Type of change

#845 


corgidrp/walker.py
```diff
@@ -156,12 +156,12 @@ def walk_corgidrp(filelist, CPGS_XML_filepath, outputdir, template=None):
                     if read_cpgs: # if not already specified.
                         # need to populate satellite spot info from XML
                         cpgs_xml = ET.parse(CPGS_XML_filepath)
-                        sat_spot_info = _get_satellite_spot_info_from_xml(cpgs_xml)
+                        sat_spot_info = _get_satellite_spot_info_from_xml_new(cpgs_xml)
```
semaphoreP: we should just replace the old one

corgidrp/walker.py
```diff
@@ -759,4 +759,45 @@ def _get_satellite_spot_info_from_xml(xml_tree):
...
+    """
+    obs_specification = xml_tree.getroot()
+    cpgs_input_info = obs_specification.find("cpgs_input")
+    try:
+        sat_spot_info = cpgs_input_info.find("satellite_spots_command_info")
+    except:
+        raise Exception("Field with satellite spot info 'satellite_spots_command_info' not found in CPGS XML file. Please ensure that the CPGS file is formatted correctly.")
+    sat_spot_output = {}
+    sat_spot_output['num_spots'] = 0
+    sat_spots = sat_spot_info.text.split(") (")
```
semaphoreP: this is a bit clunky, since it assumes there is exactly 1 space between values, which can easily not be true due to an accidence (e.g., a user puts two spaces or a comma).  We should instead use regular expression to pattern match for `(sep[1-2]={value in lam/D}, angle[1-2]={value in degrees}, fr[1-2]={flux ratio in #E# scientific notation}, filt[1-2]={filtername})`. Do you know how to use "regular expressions"?

corgidrp/walker.py
```diff
@@ -759,4 +759,45 @@ def _get_satellite_spot_info_from_xml(xml_tree):
...
+    for sat_spot in sat_spots:
+        sat_spot = sat_spot.replace("(", "").replace(")", "")
+        fields = sat_spot.split(",")
+        sat_spot_output['num_spots'] += 1
+        for i, field in enumerate(fields):
+            key, value = field.split("=")
+            #key = (''.join([c for c in key if not c.isdigit()])).replace(" ","")
+            key = ['sep','angle','contrast','filt']
+            try:
```
semaphoreP: why is this in a try/except instead of just erroring if we can't convert to a float()

corgidrp/walker.py
```diff
@@ -759,4 +759,45 @@ def _get_satellite_spot_info_from_xml(xml_tree):
             sat_spot_output['spot2_sep'] = float(pair.find("radial_distance").text)
             sat_spot_output['spot2_angle'] = float(pair.find("clocking_angle").text)
 
+    return sat_spot_output
+
+def _get_satellite_spot_info_from_xml_new(xml_tree):
```
semaphoreP: can you add a unit test in test_walker.py to check the function works with only one pair of satellite spots in `satellite_spots_command_info` instead of two?

semaphoreP (commented):
Some comments. I'd like to suggest we try implementing this with regular expression instead.

corgidrp/walker.py
aneeshb97: done!

corgidrp/walker.py
aneeshb97: implemented using regular expression. However, I've been wondering if we might want to pattern match for each field instead of all four fields together

corgidrp/walker.py
aneeshb97: done!

corgidrp/walker.py
aneeshb97: Just me being obtuse

corgidrp/walker.py
semaphoreP: definitely some pros/cons: pro: less sensitive to things like how many spaces after a comma (alternatively, just allow variable number of spaces after a comma. cons: any risk in pattern matching something unintended? perhaps not. 

Pattern match each field might have a chance to be slightly more flexible, but not a huge gain I think.

tests/test_walker.py
```diff
@@ -644,7 +645,21 @@ def test_cpgs_satspots():
         assert 'phi_deg' in hdr_recipe['steps'][0]['keywords']
         assert hdr_recipe['steps'][0]['keywords']['phi_deg'] == 0.
 
+def test_cpgs_one_satspot():
```
semaphoreP, line 648: can we update this test (and the corresponding input file) to test a few more edge cases all at once? That decimals also work in the input data and that a lowercase `e` works for the exponent: 

e.g., : `(sep2=6.2, angle2=45.1, fr2=1.7e-6, filt2=3D)`

semaphoreP (commented):
just one comment to make sure we are testing regex edge cases. thanks!

tests/test_walker.py
aneeshb97: done!

semaphoreP (approved):
Looks good, but I'm going to wait to merge until #871 (we're going to do a hotfix release with it).

semaphoreP:
@aneeshb97 whoops, there's a conflict in test_walker.py that needs to be addresed now before we can merge.

aneeshb97:
looks like another new unit test was added to test_walker.py. We can just accept both changes and merge?

semaphoreP:
yes

PR #886, opened 2026-05-13
walker makes a different recipe if satspots ias analog and science pc

### LINKED ISSUE #872: How to handle l2a->l2b processing for satspot + science data if science data is PC
Sat spot data + science data are taken together in a same visit. If the science data is PC (note that we are not benchmarking sat spot data to ever be PC), then l2a -> l2b data processing of this visit requires two separate recipes, which we are not handling. 

For input data at the l2a level, the walker should check if there is a fix of PC and non-PC data. If so, the walker should create two recipes (not chained) to process each subset of data.

helloWorlds0:
## Describe your changes
in cases where we have both analog and photon_counting data (i.e. when the data is in photon-counting mode but the satellite_spots are in analog), guess_template returns an extra recipe just for the satellite spots in the form of a  list of lists of recipes (just for L2a data). 
When generating the recipe( autogen_recipe), in cases of L2a data with both pc and analog, the recipe's input is only the pc frames if  "pc" is in the recipe name, and only the analog frames if "pc" is not in the recipe name

## Type of change


corgidrp/walker.py
```diff
@@ -239,9 +239,24 @@ def autogen_recipe(filelist, outputdir, template=None):
             if i > 0 and chained:
                 pass
             else:
-                for filename in filelist:
-                    recipe["inputs"].append(filename)
+                # add pc and analog inputs separately at L2a level
+                if first_frame.ext_hdr['DATALVL'] == "L2a":
```
semaphoreP: I think this causes the bp_map_e2e to crash, because its input dataset is empty. So we should check that first_frame is not None first.

Can we also add a comment that this logic is only for science frame processing of L2a -> L2b? Just for future us trying to remember why this piece of logic is here.

semaphoreP (commented):
looks good. there's one e2e test that fails because the dataset/first_frame can be None.

corgidrp/walker.py
helloWorlds0: Done! 
I see that there are still e2e tests that fail but they seem to also fail on the develop branch so I don't think that it's related

semaphoreP: yes, these are the new tests and Juergen's PR #907 fixes the other one.

semaphoreP (approved):

PR #890, opened 2026-05-14
Fix satspot processing issue834

### LINKED ISSUE #834: Sat spot processing needs updates to match data collection proc
**Describe the bug**
The DM offset data collection proc (aka satspots) collects equal numbers of no-offset, + offset, and - offset data,  in that order, all with the same EXCAM settings. All of these data will be flagged in the L1 header as SATSPOTS = True, since they are all collected as part of the same proc. This behavior is now documented on the [CPP wiki.](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/251768409/CGI+Proc+Descriptions)

However, find_star in L3 to L4 processing takes a nan-median of all images with SATSPOTS = True. 

**Expected behavior**
No-offset (no sat spot) frames should not be included in the image combination. These are always the first 1/3 of the SATSPOT=True dataset. 

I also suggest that the no-offset frames could be subtracted from both the + and - offset datasets to remove speckles or astrophysical objects.

semaphoreP:
## Describe your changes
 * Addresses issue #834: that the satellite spots frames first come with 1/3 of the images as background frames with no satellite spots. `find_star` can now handle these appropriately (and also comes with an option to turn off background frames)
 * Fixes the imaging and pol e2e tests to mock-up background frames
 * In doing so, realized we were missing several header keywords in the error extension for l3 and l4 imaging-coron data products. Updated the e2e test and data format docs to reflect this. 

## Type of change

Closes #834


semaphoreP:
Tagging @juliamilton for cognizance on updates to the l3/l4 data format pages: we missed some err_hdr values due to not passing the err_hdr from mocks into the e2e test data inputs.

corgidrp/l3_to_l4.py
```diff
@@ -343,16 +369,37 @@ def find_star(input_dataset,
...
         # See if the satellite spot parameters are provided, if not used defaults
         if satellite_spot_parameters is not None:
             tuningParamDict = star_center.update_parameters(tuningParamDict, satellite_spot_parameters)
         # Compute median images
         img_ref = np.nanmedian(sci_dataset.all_data, axis=0)
-        img_sat_spot = np.nanmedian(sat_spot_dataset.all_data, axis=0)
+        img_sat_spot = np.nanmedian(offset_dataset.all_data, axis=0)
+        if subtract_no_offset_frames:
+            img_no_offset = np.nanmedian(data.Dataset(no_offset_frames).all_data, axis=0)
+            img_sat_spot = img_sat_spot - img_no_offset
```
maxwellmb: I think what we want to do is to just replace img_ref, which gets subtracted in the `star_center.star_center_from_satellite_spots()` function anyway, not explicitly subtract it here.

maxwellmb: I'm not sure what we should do with `img_ref` if `subtract_no_offset_frames==False`

tests/test_find_star.py
```diff
@@ -256,81 +279,172 @@ def test_find_star_dataset_split():
...
     assert np.isclose(measured_y_visit_0, measured_y_visit_1 - delta_position, atol=0.1), \
         f"Expected y position for visit 1 to be {delta_position} pixels different from visit 0, got {measured_y_visit_0} and {measured_y_visit_1}"
 
+
+def test_satspot_no_offset_excluded():
+    """
+    Verify that the no-offset frames (first third of SATSPOTS=True frames) are excluded
+    from the satellite spot median. The mock generates no-offset frames with amplitude=0
+    (no spots) and offset frames with spots. Star-center recovery should succeed using
+    only the offset frames; it would fail if the no-offset frames diluted the signal.
```
maxwellmb: I think this test is superfluous given other existing tests. I'm also not sure it actually tests what it thinks its testing. Since we're taking a median, I don't think the frames would actually really get diluted...

corgidrp/l3_to_l4.py
maxwellmb: I think all we really need to do is replace img_ref, since that already gets subtracted from img_sat_spot. 

I'm not sure what we want to do when `subtract_no_offset_frames` is false. The current behavior of subtracting img_reg (built from the science frames) is probably not what we want. Maybe we just pass zeros?

maxwellmb (changes requested):
A few comments.

corgidrp/l3_to_l4.py
semaphoreP: Oops, good catch. I forgot this is already done in star_center_from_satellite_spots(). 

I'm going to set img_ref = zeros for subtract_no_offset_frame=False

tests/test_find_star.py
semaphoreP: I think the idea is just to test the satellite spot finding algorithm with a realistic data sequence including no-offset frames. Maybe the docstring is not well written (I can rewrite)

corgidrp/l3_to_l4.py
semaphoreP: See comment above but agreed this was my oversight (I forgot the subtraction happens later). +1 to passing in zeros when no subtraction of offset.

tests/test_find_star.py
semaphoreP: Ok I realize now that there are other unit tests that effectively test this, so I'll just delete this unit test as superfluous.

maxwellmb (approved):

Issue #710, opened 2025-12-18
L1 to Spec Dispersion E2E Test


semaphoreP:
**Is your feature request related to a problem? Please describe.**
A new E2E test that shows we can go from L1 to Spec Dispersion calibration with a single walker.walk_corgidrp() call.

**Describe the solution you'd like**
Generate simulated L1 data (e.g., using corgisim or faking it with the TVAC data) and run it fully through the pipeline (L1 -> L2b -> cal). 

**Additional context**
Don't replace the current e2e test

semaphoreP:
Added in #894

Issue #521, opened 2025-09-05
Walker needs to autodetect spec dispersion calibration files


semaphoreP:
**Is your feature request related to a problem? Please describe.**
We have a recipe template and e2e test for spectroscopy dispersion calibration, but we need to also need to autodetect the input data in the walker using the walker.guess_template() file. 

**Describe the solution you'd like**
A few lines of code added to walker.guess_template() to automatically detect the spectroscopy dispersion data and to remove the hard coded recipe template in the e2e test.

semaphoreP:
I don't think the VISTYPES are defined yet, so we may need to mock one up for now.

semaphoreP:
Closing as "won't fix", because it turns out that these observations are rare and have the same header configuration as unocculted star obserations for flux ratio computation. We anticipate this will only need to be done in commissioning, so the number of times we need to manually identify these observations is small enough that we will proceed that way (there doesn't seem to be an easy way to distinguish them from unocculted star observations for flux ratio computaion otherwise).

PR #894, opened 2026-05-18
l1 to spec dispersion e2e and autodetect dispersion recipe

JuergenSchreiber:
## Describe your changes
add to walker.guess_template to autodetect the recipes for spec dispersion calibration
add e2e script for l1 to spec dispersion calibration and check format and content of the resulting DispersionModel calibration file. 
Better handling of broad band in spec.calibrate_dispersion_model

## Type of change

#521 and #710

### DIFF
--- a/corgidrp/spec.py
+++ b/corgidrp/spec.py
@@ -617,9 +617,7 @@ def calibrate_dispersion_model(centroid_psf, spec_filter_offset, band_center_fil
     yoff = []
     for band in filters:
         band_str = band.strip()
-        if band_str == ref_cfam:
-            pass
-        elif band_str not in subband_list:
+        if band_str != ref_cfam and band_str not in subband_list:
             warnings.warn("measured band {0} is not in the sub band list {1} of the used prism".format(band_str, subband_list))
         else:
             cen_wave = read_cent_wave(band_str, filter_file = band_center_file)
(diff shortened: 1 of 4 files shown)


semaphoreP:
@JuergenSchreiber apologies, but I discussed with @neilzim and we determined we can't really automatically distinguish spec dispersion datasets from unocculted host star observations. Because of that, we're not going to try to distinguish them. Can you undo the changes to the walker and manually pass the recipes in the e2e test?

JuergenSchreiber:
> @JuergenSchreiber apologies, but I discussed with @neilzim and we determined we can't really automatically distinguish spec dispersion datasets from unocculted host star observations. Because of that, we're not going to try to distinguish them. Can you undo the changes to the walker and manually pass the recipes in the e2e test?

Also the chained recipe for l1 to dispersion, I guess?
Ok, I already asked this myself. Otherwise could we have a special VISTYPE for Dispersion calibration?

semaphoreP:
I think instead of having an automatic chained recipe, we'll need to run the walker and specify each stage of the chained recipe separately. The issue is that we don't have a special VISTYPE for dispersion calibration, and because of that, there's basically no way to guess it.

corgidrp/spec.py
```diff
@@ -629,13 +627,13 @@ def calibrate_dispersion_model(centroid_psf, spec_filter_offset, band_center_fil
             yoff.append(offset_sub[1] - yoff_band)
     if len(center_wavel) < 4:
         raise ValueError ("number of measured sub-bands {0} is too small to model the dispersion".format(len(center_wavel)))
-    if len(center_wavel) != len(centroid_psf.xfit) -1:
```
semaphoreP, line 632: do you understand why we used to be dropping the last value? just trying to understand why that decision was made.

semaphoreP (commented):
other than the comment above about removing (reverting) the changes to the walker and no longer automatically detecting dispersion calibrations, just one question about one of your changes

corgidrp/spec.py
JuergenSchreiber: It was because the last frame of the filtersweep template dataset was the broadband "3" one. It seems that I somehow wanted to exclude this from the dispersion fit, but now I think it is ok to use it, since we have the parameters for it.

semaphoreP (approved):

Issue #889, opened 2026-05-13
Spectroscopy determine_wave_zeropoint needs to account for satellite spot background frames.


semaphoreP:
**Describe the bug**
Similar to #834 for imaging mode but for spec mode. In coronagraphic satellite spot images of determine_wave_zeropoint, the first 1/3 of the dataset is background frames that we should use for background subtraction and not try use to measure zeropoint. 

**To Reproduce**
The spectroscopy l3_to_l4_psfusb_e2e test

**Expected behavior**
Same implementation as for imaging mode. We need to check that the satellite spot data is divisible by 3 and use the first 1/3 for background subtraction

**corgidrp version:**
 - develop

**Additional context**
~TODO: confirm this is not the case for non-coronagraphy zeropoint observation.~ Confirmed not a problem

semaphoreP:
Implemented in #896

PR #896, opened 2026-05-19
Spec wave zeropoint accounts for satspot background

aneeshb97:
## Describe your changes
Satspot frames will always be in groups of 3, ordered as no offset/+offset/-offset frames. Determine_wavelength_zeropoint now considers first satspot frame (ordered by SCTSRT) as no_offset frame and uses it to subtract background from +/- offset frame. Implementing this in unit tests and e2e tests requires creation of a fake no_offset satspot frame with np.zeros

## Type of change

#889 

### DIFF
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -2011,7 +2011,7 @@ def combine_spec(input_dataset, collapse="mean", num_frames_scaling=True):
                     'Z10AVG', 'Z11AVG', 'Z12AVG', 'Z13AVG', 'Z14AVG',
                     'Z2RES', 'Z3RES', 'Z4RES', 'Z5RES', 'Z6RES', 'Z7RES', 'Z8RES', 'Z9RES',
                     'Z10RES', 'Z11RES',
-                    'Z2VAR', 'Z3VAR','WAVELEN0','WV0_X','WV0_Y','WV0_XERR','WV0_YERR','WV0_DIMX', 'WV0_DIMY']) 
+                    'Z2VAR', 'Z3VAR','WAVELEN0','WV0_X','WV0_Y','WV0_XERR','WV0_YERR']) 
     #combine frames                       
     dataset = combine_subexposures(dataset, collapse=collapse, num_frames_scaling=num_frames_scaling, combine_other_hdus=True)
     #certain headers are added in combine_subexposures, we manually add them in
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -939,6 +939,7 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
     if not dpamname.startswith("PRISM"):
         raise AttributeError("This is not a spectroscopic observation. but {0}").format(dpamname)
     if dataset.frames[0].ext_hdr["FPAMNAME"] == 'OPEN_34':
+        print("The dataset has FPAMNAME = OPEN_34, identicating that this is a non-coronagraphic spectroscopy observation, setting subtract_no_offset_frames = False")
         subtract_no_offset_frames = False
 
     # Assumed that only narrowband filter (includes sat spots) frames are taken to fit the zeropoint
@@ -1044,7 +1045,7 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
             frame.ext_hdr["WV0_DIMY"] = offset_dataset[0].ext_hdr['NAXIS2']
             science_frames.append(frame)
 
-        all_science_frames = np.concatenate([np.array(all_science_frames),np.array(science_frames)])
+        all_science_frames += science_frames
 
     sci_dataset = data.Dataset(all_science_frames)
 
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -2010,7 +2010,7 @@ def combine_spec(input_dataset, collapse="mean", num_frames_scaling=True):
                     'Z10AVG', 'Z11AVG', 'Z12AVG', 'Z13AVG', 'Z14AVG',
                     'Z2RES', 'Z3RES', 'Z4RES', 'Z5RES', 'Z6RES', 'Z7RES', 'Z8RES', 'Z9RES',
                     'Z10RES', 'Z11RES',
-                    'Z2VAR', 'Z3VAR']) 
+                    'Z2VAR', 'Z3VAR','WAVELEN0','WV0_X','WV0_Y','WV0_XERR','WV0_YERR','WV0_DIMX', 'WV0_DIMY']) 
     #combine frames                       
     dataset = combine_subexposures(dataset, collapse=collapse, num_frames_scaling=num_frames_scaling, combine_other_hdus=True)
     #certain headers are added in combine_subexposures, we manually add them in
--- a/corgidrp/l3_to_l4.py
+++ b/corgidrp/l3_to_l4.py
@@ -966,63 +966,40 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
     
 
     if subtract_no_offset_frames:
-        # First we split satspot dataset into ref/target satspot as their exposure times might be different
-        satspot_dataset, exptime = sat_dataset.split_dataset(exthdr_keywords=["EXPTIME"])
+        all_satspot_frames = []
 
-        ref_sat_dataset = satspot_dataset[int(np.nonzero(exptime == exptime[0])[0].item())]
-        refstar_satspot_frames = []
-        for frame in ref_sat_dataset:
-            refstar_satspot_frames.append(frame)
-        # Split sat spot frames into the three acquisition groups by SCTSRT order.
-        # Data collection order: N no-offset frames, N +offset frames, N -offset frames.
-        if len(refstar_satspot_frames) % 3 != 0:
-            raise ValueError(f"Expected the number of refstar SATSPOTS=1 frames to be divisible by 3 "
-                f"(no-offset / +offset / -offset groups), but got {len(refstar_satspot_frames)}.")
-        if all('SCTSRT' in f.ext_hdr for f in refstar_satspot_frames):
-            refstar_satspot_frames_sorted = sorted(refstar_satspot_frames, key=lambda f: f.ext_hdr['SCTSRT'])
-        else:
-            refstar_satspot_frames_sorted = sorted(refstar_satspot_frames, key=lambda f: f.filename)
-        n_per_group = len(refstar_satspot_frames_sorted) // 3
-        no_offset_frames = refstar_satspot_frames_sorted[:n_per_group]
-        refstar_offset_frames = refstar_satspot_frames_sorted[n_per_group:]
-        
-        img_no_offset = np.nanmedian(data.Dataset(no_offset_frames).all_data, axis=0)
-        for frame in data.Dataset(refstar_offset_frames):
-            frame.data = frame.data - img_no_offset
+        # First we split satspot dataset according to VISITID
+        satspot_dataset, visitid = sat_dataset.split_dataset(prihdr_keywords=["VISITID"])
+        visitid = np.array(visitid)
 
-        if len(exptime) == 1:
-            offset_dataset = data.Dataset(refstar_offset_frames)
-                
-        elif len(exptime) > 1:
-            target_sat_dataset = satspot_dataset[int(np.nonzero(exptime == exptime[1])[0].item())]
-            target_satspot_frames = []
-            for frame in target_sat_dataset:
-                target_satspot_frames.append(frame)
+        for visid in visitid:
+            satspot_subset = satspot_dataset[int(np.nonzero(visitid == visid)[0].item())]
+            satspot_frames = []
+            for frame in satspot_subset:
+                satspot_frames.append(frame)
             # Split sat spot frames into the three acquisition groups by SCTSRT order.
             # Data collection order: N no-offset frames, N +offset frames, N -offset frames.
-            if len(target_satspot_frames) % 3 != 0:
-                raise ValueError(f"Expected the number of targetstar SATSPOTS=1 frames to be divisible by 3 "
-                    f"(no-offset / +offset / -offset groups), but got {len(target_satspot_frames)}.")
-            if all('SCTSRT' in f.ext_hdr for f in target_satspot_frames):
-                target_satspot_frames_sorted = sorted(target_satspot_frames, key=lambda f: f.ext_hdr['SCTSRT'])
+            if len(satspot_frames) % 3 != 0:
+                raise ValueError(f"Expected the number of refstar SATSPOTS=1 frames to be divisible by 3 "
+                    f"(no-offset / +offset / -offset groups), but got {len(satspot_frames)}.")
+            if all('SCTSRT' in f.ext_hdr for f in satspot_frames):
+                satspot_frames_sorted = sorted(satspot_frames, key=lambda f: f.ext_hdr['SCTSRT'])
             else:
-                target_satspot_frames_sorted = sorted(target_satspot_frames, key=lambda f: f.filename)
-            n_per_group = len(target_satspot_frames_sorted) // 3
-            no_offset_frames = target_satspot_frames_sorted[:n_per_group]
-            target_offset_frames = target_satspot_frames_sorted[n_per_group:]
-                
+                satspot_frames_sorted = sorted(satspot_frames, key=lambda f: f.filename)
+            n_per_group = len(satspot_frames_sorted) // 3
+            no_offset_frames = satspot_frames_sorted[:n_per_group]
+            offset_frames = satspot_frames_sorted[n_per_group:]
+            
             img_no_offset = np.nanmedian(data.Dataset(no_offset_frames).all_data, axis=0)
-            for frame in data.Dataset(target_offset_frames):
+            for frame in data.Dataset(offset_frames):
                 frame.data = frame.data - img_no_offset
 
-            offset_dataset = data.Dataset(np.concatenate([np.array(refstar_offset_frames),np.array(target_offset_frames)]))
+            all_satspot_frames = np.concatenate([np.array(all_satspot_frames),np.array(offset_frames)])
+
+        offset_dataset = data.Dataset(all_satspot_frames)
 
     else:
-        sat_spot_frames = []
-        for frame in sat_dataset:
-            sat_spot_frames.append(frame)
-        offset_frames = sat_spot_frames
-        offset_dataset = data.Dataset(offset_frames)
+        offset_dataset = sat_dataset
 
     if xcent_guess is not None and ycent_guess is not None:
         n = len(offset_dataset)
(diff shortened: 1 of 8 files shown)


semaphoreP:
@aneeshb97 I get the following e2e test failures. do you get the same?

FAILED tests/e2e_tests/l1_to_linespread_e2e.py::test_l1_to_linespread - ValueError: diag requires an array of at least two dimensions
FAILED tests/e2e_tests/spec_l3_to_l4_noncoron_e2e.py::test_run_end_to_end - ValueError: Expected the number of refstar SATSPOTS=1 frames to be divisible by 3 (no-offset / +offset / -offs...
FAILED tests/e2e_tests/spec_linespread_e2e.py::test_run_end_to_end - TypeError: Improper input: func input vector length N=5 must not exceed func output vector length M=0

corgidrp/l3_to_l4.py
```diff
@@ -958,15 +964,75 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
     else:
         raise AttributeError("No narrowband frames found in input dataset")
     
+
+    if subtract_no_offset_frames:
+        # First we split satspot dataset into ref/target satspot as their exposure times might be different
+        satspot_dataset, exptime = sat_dataset.split_dataset(exthdr_keywords=["EXPTIME"])
+
+        ref_sat_dataset = satspot_dataset[int(np.nonzero(exptime == exptime[0])[0].item())]
```
semaphoreP: there's no guarantee ref star images come first. instead of splitting on exptime, why don't you split by VISITID, like is done in imaging mode (you could also split by PSFREF, but what if we had multiple PSFREF stars).

semaphoreP: I think VISITID makes sense, so that we process each VISIT's zeropoint separately. Since each visit the acquisition may change and cause the zeropoint to be at a slightly different position. So we probably just want to do the whole zeropoint step function in a for loop anyways, just like find_star() in imaging mode.

semaphoreP: Also the assumption that the first 1/3 of all SATSPOT frames being background frames is only true within a Visit (every time we switch stars, or switch ADI roll angles, we have a separate visit).

semaphoreP: this includes both doing the background subtraction and doing the zeropoint measurement in a for loop on VISITID

corgidrp/l3_to_l4.py
```diff
@@ -958,15 +964,75 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
...
+        refstar_offset_frames = refstar_satspot_frames_sorted[n_per_group:]
+        
+        img_no_offset = np.nanmedian(data.Dataset(no_offset_frames).all_data, axis=0)
+        for frame in data.Dataset(refstar_offset_frames):
+            frame.data = frame.data - img_no_offset
+
+        if len(exptime) == 1:
+            offset_dataset = data.Dataset(refstar_offset_frames)
+                
+        elif len(exptime) > 1:
```
semaphoreP: the above comment about splitting the entire zeropoint step to operate on individua visits separately will also remove the need to have a duplicate copy of this code as the ref and target stars will have separate VISITIDs.

corgidrp/l3_to_l4.py
```diff
@@ -958,15 +964,75 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
...
+                frame.data = frame.data - img_no_offset
+
+            offset_dataset = data.Dataset(np.concatenate([np.array(refstar_offset_frames),np.array(target_offset_frames)]))
+
+    else:
+        sat_spot_frames = []
+        for frame in sat_dataset:
+            sat_spot_frames.append(frame)
+        offset_frames = sat_spot_frames
+        offset_dataset = data.Dataset(offset_frames)
```
semaphoreP: how is this different from just doing `offset_dataset = sat_dataset`

tests/e2e_tests/l1_to_l3_spec_e2e.py
```diff
@@ -262,6 +263,53 @@ def run_l1_to_l3_e2e_test(l1_datadir, l3_outputdir, processed_cal_path, logger):
     if not input_files:
         raise FileNotFoundError(f"No files ending in 'l1_.fits' found in {l1_datadir}")
     
+    if fits.getheader(os.path.join(l1_datadir,input_files[0]), ext=1)['CFAMNAME'] == '3D':
```
semaphoreP: This check and all the code to mock up L1 satellite spot data should be done in the spec_l3_to_l4_psfusb_e2e.py function. It makes the tests hard to read and misleading to put in in the l1_to_l3 test.

tests/e2e_tests/l1_to_l3_spec_e2e.py
```diff
@@ -262,6 +263,53 @@ def run_l1_to_l3_e2e_test(l1_datadir, l3_outputdir, processed_cal_path, logger):
...
+        # frames (science background data, no DM probe) at L1 so they flow through
+        # the full pipeline and sort first (year-2000 timestamps) into the no-offset
+        # group at L3.  Frames are grouped by (TARGET, DPAMNAME) — the same split
+        # find_star uses — and one real frame is dropped when the count is odd to
+        # keep each group size even so that n_fake = K/2 gives a divisible-by-3 total.
+        # We do this for all satspots
+
+        fake_l1 = []
+        accepted_real_files = []
+        if 'refstar_zeropoint' in l1_datadir:
```
semaphoreP: are you giving the reference and science images different dates here? if so, can you add a comment to make that more clear

tests/e2e_tests/l1_to_l3_spec_e2e.py
```diff
@@ -262,6 +263,53 @@ def run_l1_to_l3_e2e_test(l1_datadir, l3_outputdir, processed_cal_path, logger):
...
+
+        for _ in range(len(group_files) // 2):
+            fake_frame = copy.deepcopy(template_img)
+            fake_frame.data[:] = np.zeros(np.shape(template_img.data))
+            if fake_frame.err is not None:
+                fake_frame.err[:] = np.zeros(np.shape(template_img.data))
+            fake_frame.ext_hdr['SCTSRT'] = fake_base_time.isoformat()
+            time_str = fake_base_time.strftime('%Y%m%dt%H%M%S%f')[:-5]
+            fake_satspot_filename = f"cgi_{visitid}_{time_str}_l1_.fits"
+            if os.path.exists(os.path.join(l1_datadir,fake_satspot_filename)): # If we already created a fake file earlier, prevents creation of new files
```
semaphoreP: why not overwrite? if it's from a old test, we don't want it.

tests/test_spec.py
```diff
@@ -482,16 +485,32 @@ def test_determine_zeropoint():
             dq=dq
         )
         psf_images.append(image)
+        if i == 11:       #With rule of threes, we need three satspot frames. We get it by creating a zeros frame as no_offset satspot frame and duplicating the original satspot frame in the test dataset
```
semaphoreP: I have two questions here, and maybe you can clarify with comments: 

1) shouldn't there be 3 satellite spot images, so one can be used as the background no-offset frame? I only see i=11 to be the background frame and i=12 being a with sat spot frame. 

2) wouldn't it be cleaner to merge this logic with the if i ==12 if statement above? Just seems hard to read.

tests/test_spec.py
```diff
@@ -482,16 +485,32 @@ def test_determine_zeropoint():
...
     # with respect to the broadband filter.
     (xoff_nb, yoff_nb) = (steps.read_cent_wave('3D')[2], steps.read_cent_wave('3D')[3])
     (xoff_bb, yoff_bb) = (steps.read_cent_wave('3')[2], steps.read_cent_wave('3')[3])
 
-    #test it with optional initial guess and with one satspot frame
+    #test it with optional initial guess and with three satspot frames
     spec_filter_offset = SpecFilterOffset({})
     input_dataset = Dataset(psf_images)
-    dataset_guess = l3_to_l4.determine_wave_zeropoint(input_dataset, spec_filter_offset, xcent_guess = 40., ycent_guess = 32.)
+    dataset_guess = l3_to_l4.determine_wave_zeropoint(input_dataset, spec_filter_offset, subtract_no_offset_frames=True, xcent_guess = 40., ycent_guess = 32.)
```
semaphoreP: seems like you don't need to pass in this new argument since by default it is True?

semaphoreP (commented):
Some comments below. The main thing I realize is that we should probably be running the zeropoint measurement in a for loop over all VISITIDs (since a sequence with multiple VISITIDs could be passed into L3 -> L4 processing).

tests/e2e_tests/l1_to_l3_spec_e2e.py
aneeshb97: I did it here as instead of doing it twice in spec_l3_to_l4_psfusb_e2e.py, we can just do it once in the l1_to_l3 test. But can alter this if its confusing

semaphoreP: doing it in l3_to_l4 would be highly preferred. you can always write a helper function to reduce code duplication.

aneeshb97:
> Some comments below. The main thing I realize is that we should probably be running the zeropoint measurement in a for loop over all VISITIDs (since a sequence with multiple VISITIDs could be passed into L3 -> L4 processing).

oversight on my part, I did not account for all the scenarios where subtract_no_offset_frame = False. Instead of putting if conditions in l3_to_l4.py for different modes, I directly change the recipe template for those specific modes

tests/test_spec.py
aneeshb97: there is already a satspot image in the mock dataset. We create a fake background frame, and duplicate the pre-existing satspot frame, giving us one background and two satspot frames.

The i==12 if statement earlier specifies some keywords for the satspot frame before creating an Image object out of it. The i==12 later duplicates the previously created Image object, so that we have two satspot Images

semaphoreP: can we make this code cleaner so these points are more obvious when reading the code? Looking at it, I wouldn't be able to come to these conclusions.

tests/test_spec.py
aneeshb97: we don't need to pass it, just specified for clarity.

tests/e2e_tests/l1_to_l3_spec_e2e.py
aneeshb97: While we can put the mock generation for L1 satellite spot data in l3_to_l4, the selection procedure for choosing the correct input satspot files to feed into the L1->L3 pathway would need to be in l1_to_l3

aneeshb97: this is because the e2e tests take the L1 data directory as input, and I imagine we want to keep this consistent across all e2e tests

tests/e2e_tests/l1_to_l3_spec_e2e.py
aneeshb97: ok!

corgidrp/l3_to_l4.py
aneeshb97: target/ref datasets from corgisim don't have different visitids, doing a hack change to the target dataset visitids

corgidrp/l3_to_l4.py
aneeshb97: it isn't, just me being dumb

tests/e2e_tests/l1_to_l3_spec_e2e.py
semaphoreP: doesn't l1 -> l3 not care about sat spot or not sat spot frames? can't they all just get processed indiscriminately?

tests/test_spec.py
semaphoreP: I'd prefer we not include it for clarity. Adding the keyword seems like it is intentional, that this is different from default behavior.

tests/e2e_tests/l1_to_l3_spec_e2e.py
aneeshb97: Why would we want to process a satspot frame from L1->L3 if we aren't using it in determine_wave_zeropt?

semaphoreP: I'm just saying the l3->l4 test should identify which frames are needed to for analysis, and just treat the function call to the l1_to_l3 e2e test as a processing step. The l1_to_l3_e2e tests shouldn't have any logic for sat spots, because the processing for sat spot and non-sat-spot data is the same.

corgidrp/l3_to_l4.py
```diff
@@ -934,6 +938,8 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
     dpamname = dataset.frames[0].ext_hdr["DPAMNAME"]
     if not dpamname.startswith("PRISM"):
         raise AttributeError("This is not a spectroscopic observation. but {0}").format(dpamname)
+    if dataset.frames[0].ext_hdr["FPAMNAME"] == 'OPEN_34':
```
semaphoreP, line 941: what is this for? should there be a print statement indicating that we've forced subtract_no_offset_frames = False?

corgidrp/l3_to_l4.py
```diff
@@ -958,41 +964,90 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
...
+            frame.ext_hdr["WAVLEN0"] = cen_wave
+            frame.ext_hdr["WV0_X"] = x0
+            frame.ext_hdr["WV0_XERR"] = x0err
+            frame.ext_hdr["WV0_Y"] = y0
+            frame.ext_hdr["WV0_YERR"] = y0err
+            frame.ext_hdr["WV0_DIMX"] = offset_dataset[0].ext_hdr['NAXIS1']
+            frame.ext_hdr["WV0_DIMY"] = offset_dataset[0].ext_hdr['NAXIS2']
+            science_frames.append(frame)
+
+        all_science_frames = np.concatenate([np.array(all_science_frames),np.array(science_frames)])
```
semaphoreP: very minor, but this is probably less clunky as just lists. You can do all_sceince_frames += science_frames to append.

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
```diff
@@ -42,6 +44,57 @@ def patch_eacq_to_center_if_missing(filelist):
...
+        list: List of L1 filenames for input.
+    """
+
+    if 'refstar_zeropoint' in l1_datadir:  # Different SCTSRT keyword for refstar and targetstar background satspot files
+        fake_base_time = datetime.datetime(2000, 1, 1, 0, 0, 0)
+    else:
+        fake_base_time = datetime.datetime(2000, 1, 1, 0, 0, 1)
+
+    group_files = sorted(glob.glob(os.path.join(l1_datadir, "cgi_*l1_.fits")))
+    if len(group_files) % 2 != 0:
```
semaphoreP, line 74: if len(group_files) % 2 == 0, do you not do anything? I'm guessing we never run into this case?

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
```diff
@@ -195,6 +282,13 @@ def run_spec_l3_to_l4_psfsub_e2e_test(e2edata_path, e2eoutput_path):
     # Scan for default calibrations
     this_caldb.scan_dir_for_new_entries(corgidrp.default_cal_dir)
     
+    # Reverting changed visitids to old values.
+    if changed_visitid:
+        for f in target_satspot_files:
+            fits.setval(f, 'VISITID', value = '0200001001001001001', ext=0)
```
semaphoreP: instead of assuming you know the old values, you should store them in two variables, and set them back here.

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
semaphoreP, line 1: Somehow this test is now not generating the appropriate number of satellite spot frames for me. I get the error "FAILED tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py::test_run_end_to_end - ValueError: Expected the number of refstar SATSPOTS=1 frames to be divisible by 3 (no-offset / +offset / -offs...". 

Haven't had the chance to debug

semaphoreP (commented):
the e2e test is now failing for new + 4 very minor comments.

corgidrp/l3_to_l4.py
aneeshb97: Just an extra check for all spec modes that are not coron. Can remove if confusing. Or set a warning

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
aneeshb97: if len(group_files) = 4, then we have already created a fake frame and don't need more. 
If len(group_files) = 2, it would imply the original satspot dataset had just 2 frames and throw an error

corgidrp/l3_to_l4.py
aneeshb97: okay!

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
aneeshb97: Since we change them only if psfref VISITID = target VISITID, just storing the psfref VISITID as a variable

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
aneeshb97: If you have already created new frames earlier, you'll have to delete them before running the e2e again

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
semaphoreP: I still don't quite understand... but could we make a new l1_datadir in the outputdir and process from the l1 folder in the outputdir, so we don't have to worry about the state of the l1_datadir folder?

semaphoreP: we should try avoiding saving data to the original e2e test data dir, only to the outputdir.

corgidrp/l3_to_l4.py
semaphoreP: set a warning. I don't like hidden behavior happening.

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
semaphoreP: that works, thanks!

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
aneeshb97: With latest commit, all fake files are created and saved to outputdir, which is purged and remade on each run. Now fake file/s are created on each run

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
semaphoreP: can we move all of these files to e2eoutput_path rather than e2edata_path, so it gets saved in the output folder?

semaphoreP (commented):
sorry, just one more path thing I noticed when looking at this a bit closely. Can we save these into the outpudir folder instead?

tests/e2e_tests/spec_l3_to_l4_psfsub_e2e.py
aneeshb97: Done!

semaphoreP (approved):

Issue #895, opened 2026-05-18
l1_to_l2b_e2e.py fails with AssertionError


Rumpelstil:
**Describe the bug**
l1_to_l2b_e2e.py results in "FAILED tests/e2e_tests/l1_to_l2b_e2e.py::test_l1_to_l2b - assert False"

Error in tests/e2e_tests/l1_to_l2b_e2e.py:249: AssertionError

>           assert np.array_equal(e2e_nans,tvac_nans)
E           assert False
E            +  where False = <function array_equal at 0x102c759f0>((array([746, 746, 746, 746, 746, 746, 746, 746, 746, 747, 747, 747, 747,\n       747, 747, 747, 747, 747, 748, 748, 748...37, 728,\n       729, 730, 731, 732, 733, 734, 735, 736, 737, 728, 729, 730, 731,\n       732, 733, 734, 735, 736, 737])), (array([746, 746, 746, 746, 746, 746, 746, 746, 746, 746, 747, 747, 747,\n       747, 747, 747, 747, 747, 747, 747, 748...28, 729, 730, 731,\n       732, 733, 734, 735, 736, 737, 726, 727, 728, 729, 730, 731, 732,\n       733, 734, 735, 736])))
E            +    where <function array_equal at 0x102c759f0> = np.array_equal

**To Reproduce**
pytest --which e2e --e2edata_path /Users/brandner/data/Roman/E2E_Test_Data --e2eoutput_path tests/e2e_tests/ tests/e2e_tests/l1_to_l2b_e2e.py

**Expected behavior**
"passed"

**corgidrp version:**
 4.1

**Additional context**
running it with python 3.14.4 on macOS 26.5, M4. All tests, and all e2e tests up to this test (when called in alphabet order) passsed.

semaphoreP:
all the e2e tests pass on a clean Github Actions run, so I think this is an issue with your specific setup, and not a general DRP issue. Did you clean out your .corigdrp folder in your home directory (just delete it completely), before running the tests. Furthermore, we haven't officially tested the DRP on python 3.14, as we've been baselining python 3.12.

maxwellmb:
Another thing to check: when is the last time you downloaded the e2e test data? TBH, I'd be surprised if it caused this issue, but it's always a good thing to cross off the list as a potential cause.

Rumpelstil:
.corigdrp was deleted just ahead of starting the tests, and e2e is the zenodo version as of May 18. 

[...]

PR #905, opened 2026-05-21
Keep CR frames through L2a and flag them  IS_BAD

### LINKED ISSUE #843: Cosmic Ray Identification code throws out frames and FDD forbids that
**Describe the bug**
In order to avoid unreasonably long computation times the Cosmic Ray Identification step function (detect_cosmic_rays; l1_to_l2a.py) silently discards frames and FDD forbids that. 

**To Reproduce**
Make some super saturated frames and run them through the 

**Expected behavior**
We need to keep the frames, but not run the cosmic ray code on them, and then reject them in the appropriate step function in l2a_to_l2b.py. 

**Additional context**
Do so, or face Eric's wrath.

clarissardoo:
## Describe your changes

Updates cosmic ray detection so oversaturated frames are no longer silently removed during L1 -> L2a processing. Instead the oversaturated frames are kept, marked with IS_BAD=True and assigned saturation DQ across the frame. Frames are still skipped by the expensive cosmic ray detection step.

Updates frame_select() so frames already marked IS_BAD=True by an upstream step are rejected during L2a -> L2b processing.

Adds unit tests for both behaviors under test_cr_detection and test_frame_selection. 

Also makes the l1_to_l2b_e2e input selection deterministic by sorting the L1 files and avoiding known problematic inputs to address #895 

## Type of change

Fixes #843 , #895

- [X ] I have verified that all unit tests pass in a clean environment and added new unit tests, as appropriate
- [ X] I have checked that I am merging into the right branch
- [ X] I have checked if my code modifies an existing step function or modifies the data format docs. If it does, I've added my PR to the [list maintained here](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/212028061/Log+of+PRs+that+could+affect+existing+VAP+Tests).

### DIFF
--- a/corgidrp/l1_to_l2a.py
+++ b/corgidrp/l1_to_l2a.py
@@ -159,7 +159,8 @@ def prescan_biassub(input_dataset, noise_maps=None, return_full_frame=False,
 
 def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh=0.95,
                        plat_thresh=0.85, cosm_filter=1, cosm_box=3, cosm_tail=10,
-                       mode='image', detector_regions=None, pct_oversat_lim=20, dataset_copy=True):
+                       mode='image', detector_regions=None, pct_oversat_lim=20,
+                       dataset_copy=True, discard_oversat=False):
     """
     Detects cosmic rays in a given dataset. Updates the DQ to reflect the pixels that are affected.
     TODO: (Eventually) Decide if we want to invest time in improving CR rejection (modeling and subtracting the hit
@@ -205,9 +206,11 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
             found in detector_areas in detector.py. Defaults to detector_areas in detector.py.
         pct_oversat_lim: (float):
             Percent of total frame over sat_fwc over which we determine the frame is oversaturated
-            and will be discarded. Frame saturations equal to this argument are not discarded.
+            and will be marked bad or discarded. Frame saturations equal to this argument are not flagged.
         dataset_copy (bool): flag indicating whether the input dataset will be preserved after this function is executed or not.  If False, the output dataset will be the input dataset modified, and 
             the input and output datasets will be identical.  This is useful when handling a large dataset and when the input dataset is not needed afterwards. Defaults to True.
+        discard_oversat (bool): if True, discard frames that exceed pct_oversat_lim, preserving the previous behavior.
+            If False, keep them, mark IS_BAD, and skip cosmic ray identification for those frames. Defaults to False.
 
     Returns:
         corgidrp.data.Dataset:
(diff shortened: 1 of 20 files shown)


corgidrp/l1_to_l2a.py
```diff
@@ -255,15 +255,26 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
     crmasked_cube = crmasked_dataset.all_data
+    sat_fwcs = initial_sat_fwcs
+
+    # keep oversaturated frames, but mark them bad and skip the expensive CR search.
+    oversat_frames = set()
+    for i, frame in enumerate(crmasked_dataset.frames):
+        pct_frame_sat = ((frame.data > sat_fwcs[i]).sum() / frame.data.size) * 100
+        if pct_frame_sat > pct_oversat_lim:
+            frame.ext_hdr['IS_BAD'] = True
+            oversat_frames.add(i)
```
maxwellmb: I think it would be good to add some kind of HISTORY header comment as to why the frame was marked bad.

corgidrp/l1_to_l2a.py
```diff
@@ -255,15 +255,26 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
         frame.ext_hdr['FWC_EM_E'] = fwcem_e_arr[i]
         frame.ext_hdr['SAT_DN'] = initial_sat_fwcs[i]
 
-    # Remove images that are too saturated to remove cosmics in a timely manner
-    crmasked_dataset, sat_fwcs = remove_sat_images(initial_dataset, initial_sat_fwcs, pct_oversat_lim, dataset_copy=dataset_copy)
```
maxwellmb, line 259: Rather than completely remove the old functionality, could we just make it an option? It's nice to stay somewhat backwards compatible when possible. (Default to new behavior though)

corgidrp/l1_to_l2a.py
maxwellmb: Also perhaps add a note that cosmic ray identification was skipped.

corgidrp/l1_to_l2a.py
```diff
@@ -255,15 +255,26 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
+            frame.ext_hdr['IS_BAD'] = True
+            oversat_frames.add(i)
 
     sat_fwcs_array = np.array([np.full_like(crmasked_cube[0],sat_fwcs[i]) for i in range(len(sat_fwcs))])
 
     # threshold the frame to catch any values above sat_fwc --> this is
     # mask 1
     m1 = (crmasked_cube >= sat_fwcs_array) * sat_dqval
+    for i in oversat_frames:
+        m1[i, :, :] = sat_dqval
```
maxwellmb: I don't think this is necessary. We already flag the pixels that are saturated. We don't need to flag ones that aren't.

maxwellmb (changes requested):
Some quick thoughts

maxwellmb (approved):

Issue #899, opened 2026-05-19
Pol calibration chained recipes not available for auto-select


mygouf:
Pol calibration chained recipes not available for auto-select in guess-template:

Missing at least:
- [l1_to_l2a_basic.json+l2a_to_l2b.json+l2b_to_fluxcal_factor_pol.json]
- [l1_to_l2a_basic.json+l2a_to_l2b.json+l2b_to_polcal.json] 

+ need to check if others may be missing

maxwellmb:
Example of how to do this for spec/imaging: https://github.com/roman-corgi/corgidrp/blob/c23cc8359cb225ed1a54fd1deb2838bba17de641/corgidrp/walker.py#L379

adrienmaillard:
In the recipes you mention, you use `l2a_to_l2b.json` but I see that `l2a_to_l2b_pol.json` is used in the L2 block for POL data. Is that a typo ? 
Also I see that photon counting (ISPC if I am not mistaken) is not taken into account in all L1 chains, I guess that's by design ?

maxwellmb:
1) Yeah, we want it to use the _pol recipes
2) Photon counting will only really happen in some of the science data, so for these calibrations we don't need to worry as much about it

semaphoreP:
No photon counting for calibration data I believe.

maxwellmb:
Closed with PR #908

PR #908, opened 2026-05-26
Add 2 recipes with polarization

adrienmaillard:
## Describe your changes
I have added the 2 missing recipes mentioned in #899. 

One is a case of the ABSFLUX handler that was already existing so I added it after the spectroscopy case. I guessed they are incompatible because it's 2 different DPAMNAME.  There is no recipe for multiple FSM + POL (I guess it would be like nd_filter_pol) so I put it in the general case after.  

The other case is the pol calibration, I followed the recipes used in l2a and l2b stages.

Compared to the ticket, I used `l2a_to_l2b_pol.json` instead `l2a_to_l2b.json` because it's the recipe used for pol data.

I added some tests exercising these paths. 

## Type of change

### DIFF
--- a/corgidrp/walker.py
+++ b/corgidrp/walker.py
@@ -425,7 +425,11 @@ def guess_template(dataset):
                     recipe_filename = ["l1_to_l2a_basic.json", "l2a_to_l2b.json", "l2b_to_nd_filter.json"]
                     chained = True
                 else:
-                    recipe_filename = ["l1_to_l2a_basic.json", "l2a_to_l2b.json", "l2b_to_fluxcal_factor.json"]
+                    # Check for polarimetry mode to use appropriate flux calibration recipe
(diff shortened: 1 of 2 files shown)


maxwellmb (approved):
Overall this looks good to me. I'd like to hear from @mygouf to see if it solves her DIP problem. Marie can you checkout this branch and see if it does what you need? 

@adrienmaillard for future PRs, it would be great if you could turn off your auto-linter that's introducing various extra spaces around the different files. For bigger PRs it can make it much harder to review when there are many minor inconsequential changes.

maxwellmb (changes requested):
I'm actually having second thoughts about this implementation. I think the better way forward is just to merge this into one recipe. l1_to_polcal for example. The steps were original separate because of how the DIP was originally going to operate, but that has changed, so merging things into one recipe makes sense.

maxwellmb (approved):
After discussions with @semaphoreP we decided this is fine as-is.

PR #913, opened 2026-05-27
add combine function in l3 to l4 step

### LINKED ISSUE #897: Want to add a combination step to l3_to_l4 to reduce number of frames in PSF subtraction
Current datasets might have thousands of frames and this is causing execution time issues with the current psf-subtraction setup. 

We'd like to add a combine frames step to l3_to_l4 processing  (e.g., combine all frames in a single visit, combine every 100 frames, or combine frames so that there are max 100 frames). It should work for both pol and standard imaging (maybe spec too TBD). 

**Additional context**
The issue is an off-shoot from Issue #726.

zjwhale:
## Describe your changes

This PR adds `l3_to_l4.combine_frames()` in `l3_to_l4.py` to combine frames after `l3_to_l4.find_star()` and before `l3_to_l4.do_psf_subtraction()`. `l3_to_l4.combine_frames()` calls the `combine_subexposures` function from combine.py

The input dataset is first split into subgroups with matching `VISITID` and `DPAMNAME`, and frames are then combined within each subgroup. The function accepts a `num_frames_per_group` keyword to control how many frames are combined at a time for each type, but this value is automatically overridden when needed to keep the resulting number of output frames below 100.

A unit test file "test_combine_l2_to_l4.py" is also added. 

## Type of change

- Issue #897


corgidrp/l3_to_l4.py
```diff
@@ -2092,3 +2092,98 @@ def update_to_l4_pol(input_dataset, corethroughput_cal, flux_cal_pol0, flux_cal_
         frame.ext_hdr['FLXCLF45'] = flux_cal_pol45.filename.split("/")[-1]
 
     return updated_dataset
+
+def combine_frames(input_dataset, collapse="mean", num_frames_per_group=None, num_frames_scaling=False, combine_other_hdus=False):
```
semaphoreP: can we rename this combine_frames_per_visit or something, to make it clear it's specific purpose?

Can we also move this function to combine.py? It may not necessarily always be called in the l3_to_l4 step given its usefulness.

corgidrp/l3_to_l4.py
```diff
@@ -2092,3 +2092,98 @@ def update_to_l4_pol(input_dataset, corethroughput_cal, flux_cal_pol0, flux_cal_
...
+
+    Args:
+        input_dataset (corgidrp.data.Dataset): a dataset of Images (L3-level)
+        collapse (str): "mean" or "median". (default: mean) 
+        num_frames_per_group (int or None): number of consecutive frames to combine
+                                            within each VISITID/DPAMNAME split subset.
+                                            If None, combine all frames in each subset
+                                            into one output frame. If the requested
+                                            grouping would create more than 100 output
+                                            frames in a split subset, it is overridden
```
semaphoreP: we should have an optional keyword that's like max_combined=100 to specify this, instead of hardcoding 100.

corgidrp/recipe_templates/l3_to_l4.json
semaphoreP, line 1: should we add the step in the pol and spec l3 to l4 psfsub recipes too?

corgidrp/l3_to_l4.py
```diff
@@ -2092,3 +2092,98 @@ def update_to_l4_pol(input_dataset, corethroughput_cal, flux_cal_pol0, flux_cal_
...
+        present for all frames. If ``num_frames_per_group`` is set, each split
+        subset is further combined in sequential groups of
+        ``num_frames_per_group``. If a requested grouping would create more than
+        100 output frames in a split subset, the grouping is increased so that
+        the output count stays at or below 100 while still dividing the subset
+        length exactly. Each output frame keeps the same internal data cube
+        shape as an input frame; only the dataset/frame axis is collapsed.
+    """
+    # check that we are running this on L3 data
+    for orig_frame in input_dataset:
```
semaphoreP: I don't think we should have this check, as I'd like this step function to be more flexible than l3_to_l4 only.

corgidrp/l3_to_l4.py
```diff
@@ -2092,3 +2092,98 @@ def update_to_l4_pol(input_dataset, corethroughput_cal, flux_cal_pol0, flux_cal_
...
+        if num_frames_per_group is not None:
+            requested_output_frames = len(split_dataset) / num_frames_per_group
+            if requested_output_frames > 100:
+                # combine_subexposures requires an exact divisor, so choose the
+                # largest allowed output-frame count that divides the subset length.
+                max_output_frames = 1
+                for candidate_output_frames in range(100, 0, -1):
+                    if len(split_dataset) % candidate_output_frames == 0:
+                        max_output_frames = candidate_output_frames
+                        break
```
semaphoreP: I think we should just do effective_num_frames_per_group = len(split_dataset)//100. Combine_subexposures can handle non-exact divisors and will just leave out the extra remainder frames.

corgidrp/l3_to_l4.py
```diff
@@ -2092,3 +2092,98 @@ def update_to_l4_pol(input_dataset, corethroughput_cal, flux_cal_pol0, flux_cal_
...
+            num_frames_scaling=num_frames_scaling,
+            combine_other_hdus=combine_other_hdus,
+        )
+        combined_frames.extend(combined_split.frames.tolist())
+
+    combined_dataset = data.Dataset(combined_frames)
+
+    if num_frames_per_group is None:
+        history_msg = (
+            f"Combined frames by applying {collapse} within each VISITID/DPAMNAME subset, "
```
semaphoreP: maybe make it more explicit we binned all frames from each subset into a single frame through {collapse} collapse.

semaphoreP (commented):
It seems like the l2b_to_l4_e2e.py test fails because the reference star and science data all have the same VISITID (I think). I think we need to modify that e2e test to change the VISITID of the science and ref data, so they don't get combined together.

We should also and test this combine step for spec and pol mode. 

Otherwise, this generally looks good. I have some stylistic comments mainly.

corgidrp/l3_to_l4.py
```diff
@@ -2092,3 +2092,98 @@ def update_to_l4_pol(input_dataset, corethroughput_cal, flux_cal_pol0, flux_cal_
...
+            raise ValueError(err_msg)
+    if num_frames_per_group is not None and num_frames_per_group <= 0:
+        raise ValueError("num_frames_per_group must be a positive integer or None")
+
+    split_kwargs = {"prihdr_keywords": ["VISITID"]}
+    if all("DPAMNAME" in frame.ext_hdr for frame in input_dataset):
+        split_kwargs["exthdr_keywords"] = ["DPAMNAME"]
+
+    # Combine each visit/polarization subset independently.
+    split_datasets, _ = input_dataset.split_dataset(**split_kwargs)
```
maxwellmb: I think it'd be nice to allow the user to pass in their own pri_split_kwargs and ext_split_kwargs. 
This is also a place we can make sure to split by target keyword (although I think @semaphoreP's right in real data different targets will always have different VISITIDs)

maxwellmb (changes requested):

corgidrp/l3_to_l4.py
maxwellmb: see the implementation in the `find_star` function

corgidrp/l3_to_l4.py
zjwhale: Sure.

zjwhale:
> It seems like the l2b_to_l4_e2e.py test fails because the reference star and science data all have the same VISITID (I think). I think we need to modify that e2e test to change the VISITID of the science and ref data, so they don't get combined together.
> 
> We should also and test this combine step for spec and pol mode.
> 
> Otherwise, this generally looks good. I have some stylistic comments mainly.

Hi Jason, I have addressed all comments except for this one. I found in l2b_to_l4_e2e.py, the input e2edata_dir is a local path. Is that right? woule it cause issue in  test run in repo? 
`    e2edata_dir = '/Users/kevinludwick/Documents/DRP_E2E_Test_Files_v2/E2E_Test_Data' `

semaphoreP:
No that little piece of code of if name==main is a chunk of debugging code. We normally call these e2e tests using pytest (See README for instructions), and in calling pytest, we specify the e2edata_dir on our machine

corgidrp/l3_to_l4.py
zjwhale: Hi  Jason @semaphoreP , I changed the VISID for both the reference frames and the science frames, and now l2b_to_l4_e2e.py can run through the L2-to-L4 steps. However, the test still fails at the final step: “Read in the psf_subtracted images and test for source detection.” The companion is not detected at the expected locations. I suspect that something may have gone wrong during the PSF subtraction, possibly due to the frame combination step.

I have attached the saved PSF-subtracted image from my test run with this branch, compared with the one from dev branch. 

Left is the psf subtracted image using this branch (with combing frames), right is the psf subtracted image using dev branch (without combining frames). The recovered injected companions look fainter. 
Screenshot 2026-06-03 at 4 42 29 PM

Is the performance difference expected?

semaphoreP: That seems like a bug. Here's the combined image on the main branch: 
image

My guess is: did you forget to assign different VISITIDs to the different rolls for the science image? each roll should have a different VISITID. If they don't, they get combined together in your PR, and that will smear out the planet.

zjwhale: how could I tell which frame is which roll? I check the headers for the mock dataset, there is no headers for roll angle
Screenshot 2026-06-04 at 10 01 10 AM

should I use Batch ID?
I checked hlc_os11_batch_info.txt and it shows diffrenet batch id corresponding to different roll

semaphoreP: sure, we also create the files with different PA_APER keywords, to record the different rolls.

zjwhale: Hi  Jason @semaphoreP , when science frames with different roll angles and reference frames have different visit IDs, should separate satellite-spot images be added for each visit?

I’m currently seeing an error in the `find_star `step because it splits the dataset by VISITID. In l2_to_l4.py, satellite-spot frames are only added for the science frames. This worked before because all the relevant frames shared the same VISITID, but now the science and reference frames have different VISITIDs, so one split has no satellite-spot frames.

If I change the find_star setting so it does not split the dataset by VISITID, the test passes. But I wanted to check with you what the best practice should be here.

maxwellmb: Yes, ref and science both need sat spot frames. Rolls (which will have different VISITIDs) will also need different sat spots. It could be that the original test data didn't generate that.

semaphoreP: Agreed with Max. You'll probably need to update the e2e test to make more sat spot images.

corgidrp/l3_to_l4.py
zjwhale: I added the satellite spot frames to ref stars as well. the `l2b_to_l4_e2e.py` passed

tests/e2e_tests/l1_to_l4_pol_e2e.py
```diff
@@ -27,6 +27,162 @@
...
+    if source_filelist is None:
+        source_filelist = filelist
+
+    target_entries = {}
+    for filepath in source_filelist:
+        with fits.open(filepath) as hdul:
+            target = hdul[0].header.get('TARGET', '')
+            dpamname = hdul[1].header.get('DPAMNAME', '')
+            visitid = str(hdul[0].header.get('VISITID', '')).strip()
+        target_entries.setdefault(target, {})[dpamname] = visitid
```
semaphoreP: what is this line of code doing? can we split it up into multiple lines. it's quite hard to understand the intention.

tests/e2e_tests/l1_to_l4_pol_e2e.py
```diff
@@ -27,6 +27,162 @@
...
+
+    visitids_by_target = {}
+    for target, visitids_by_dpam in target_entries.items():
+        if 'POL0' in visitids_by_dpam:
```
semaphoreP, line 96: why would pol0 be in VISITID? I thought VISITIDs are just numbers?

tests/e2e_tests/l1_to_l4_pol_e2e.py
```diff
@@ -27,6 +27,162 @@
...
+            visitid = visitids_by_target[target][dpamname]
+            hdul[0].header['VISITID'] = visitid
+            hdul.flush()
+    return filelist
+
+
+def make_satspot_frame_for_visit(visitid, output_dir, filepath=None,
+                                 template_img=None, science_img=None,
+                                 fake_time=None, exptime_ratio=None):
+    """Create or copy one L1 satellite-spot frame for a requested visit split.
```
semaphoreP, line 125: this function is a bit misleading because you aren't actually creating satellite spots in an image. you're just scaling and modifying file headers. Can you rename this function to be more clear

tests/e2e_tests/l1_to_l4_pol_e2e.py
```diff
@@ -354,11 +510,12 @@ def run_l1_to_l4_e2e_test(l1_datadir, l4_outputdir, processed_cal_path, logger):
     # Update headers
     # TO DO: pol sims should have the correct VISTYPE, currently undefined
```
semaphoreP: is this comment now fixed?

tests/e2e_tests/l1_to_l4_pol_e2e.py
```diff
@@ -457,22 +623,33 @@ def run_l1_to_l4_e2e_test(l1_datadir, l4_outputdir, processed_cal_path, logger):
...
+            accepted_real_files.extend(list(group_files))
             exptime_ratio = template_img.ext_hdr['EXPTIME'] / science_exptime
 
+            # Duplicate real satspot frames so unocculted visits have SATSPOTS=1
+            # frames when find_star uses its default VISITID/DPAMNAME split.
+            for real_file in group_files:
+                accepted_real_files.append(make_satspot_frame_for_visit(
+                    visitid=unocculted_visitid,
+                    output_dir=sat_spot_dir,
+                    filepath=real_file,
```
semaphoreP: unocculted visits should not have SATSPOT frames.

tests/e2e_tests/l1_to_l4_pol_e2e.py
```diff
@@ -457,22 +623,33 @@ def run_l1_to_l4_e2e_test(l1_datadir, l4_outputdir, processed_cal_path, logger):
...
-                fake_frame.data[:] = science_img.data[:] * exptime_ratio
-                if fake_frame.err is not None:
-                    fake_frame.err[:] = science_img.err[:] * exptime_ratio
-                fake_frame.ext_hdr['SCTSRT'] = fake_base_time.isoformat()
-                time_str = fake_base_time.strftime('%Y%m%dt%H%M%S%f')[:-5]
-                fake_frame.save(filedir=sat_spot_dir, filename=f"cgi_{visitid}_{time_str}_l1_.fits")
-                fake_l1_paths.append(fake_frame.filepath)
-                fake_base_time += timedelta(seconds=1)
+                for fake_visitid in [visitid, unocculted_visitid]:
+                    fake_l1_paths.append(make_satspot_frame_for_visit(
```
semaphoreP: I'm confused by this double for loop, how many satellite spot frames do we end up with in total? it should be a number divisible by 3.

tests/e2e_tests/l2b_to_l4_e2e.py
```diff
@@ -23,6 +23,99 @@
...
+    unique_pa_apers = sorted(set(pa_aper_values))
+    if len(unique_pa_apers) != len(visitids):
+        raise ValueError(
+            f"Expected {len(visitids)} PA_APER values, but found "
+            f"{len(unique_pa_apers)}: {unique_pa_apers}"
+        )
+    return [visitids[unique_pa_apers.index(pa_aper)] for pa_aper in pa_aper_values]
+
+
+def make_satspot_frame_for_visit(visitid, pa_aper, psfref, sctsrt,
```
semaphoreP: shoudl rename this function since it doesn't actually generate satspot data, it just modifies an existing image in some way for sat spots.

tests/e2e_tests/l2b_to_l4_e2e.py
```diff
@@ -119,8 +212,16 @@ def test_l2b_to_l3(e2edata_path, e2eoutput_path):
     istart = 0
 
     big_array_size = [1024,1024]
+    science_visitids = ("1111111111111111101", "1111111111111111201")
+    reference_visitid = "2222222222222222222"
+
+    science_pa_aper_values = -rotation[star != 2]
```
semaphoreP, line 219: what does this != 2 mean? to ignore the reference star?

tests/e2e_tests/l3_to_l4_pol_e2e.py
```diff
@@ -386,13 +458,34 @@ def test_l3_to_l4_pol_e2e(e2edata_path, e2eoutput_path):
...
+                    visit_offset=UNOCCULTED_VISITID_OFFSET,
+                )
                 stellar_nd_wp_img.ext_hdr['FPAMNAME'] = "ND225"
                 stellar_nd_wp_img.pri_hdr['PA_APER'] = rotation_angle
                 stellar_nd_wp_img.ext_hdr['NORTHANG'] = astrom_cal.northangle-(rotation_angle-astrom_cal.pri_hdr['PA_APER'])
                 stellar_nd_wp_img.ext_hdr['FSMPRFL'] = 'NFOV'
 
                 input_image_list.append(stellar_nd_wp_img)
 
+        for i in range(number_of_sat_spot_images):
```
semaphoreP: why are the unocculted images getting sat spot images?

semaphoreP (commented):
The new tests are passing, but I don't fully understand some of the additions, so I have some questions and requests for clarity below.

semaphoreP:
@agreenbaum for awareness: we're adding a keyword NUM_FR to the l4 data products for imaging and pol mode. Let us know if there's any additional actions we should do for this.

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: The function `apply_pol_visitids_to_l1_files()` assigns different VISITID values to L1 files based on the target name and polarization mode. It first read the target, DPAMNAME and VISITID from the orignal L1 fits file used in the test and builds a lookup table `target_entries`. The code can later access the VISITID associated with that target/product combination value by `target_entries[target][dpamname]`

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: The code assigns the VISITID based on the value of DPAMNAME in the FITS header. If DPAMNAME is POL0, it keeps the original VISITID from the header. Otherwise, such as for POL45, it applies an offset to the original VISITID. Since VISITID values are numeric strings, the code converts them to numbers before adding the offset, then writes the updated value back as a string.

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: if I do not add SATSPOT frames to the unocculted visits, the find_star() step fails. By default, find_star() splits frames by VISITID. Since the unocculted and occulted frames are assigned different VISITIDs, find_star() cannot process the unocculted visits unless they also have corresponding SATSPOT frames. Is that correct?

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: It will also apply an additional offset for unocculted frames.

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: Sure.

tests/e2e_tests/l1_to_l4_pol_e2e.py
semaphoreP: well unocculted images will never be taken with satellite spots for the real data. Can we have find_star skip the unocculted data?

tests/e2e_tests/l1_to_l4_pol_e2e.py
semaphoreP: I understand that, but this specific line looks like it's trying to initialize a dictionary entry and set it at the same time. I think it would be more clear if you split it up to a multi-line oepration.

tests/e2e_tests/l1_to_l4_pol_e2e.py
semaphoreP: ok can you make a comment to explain this. it's not very obvious that's what happens from reading the code

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: group_files contains only the real offset satellite-spot frames: half are +offset frames and half are -offset frames. The loop over len(group_files) // 2 adds one fake no-offset frame for each +offset / -offset pair. This makes the final number of satellite-spot frames divisible by 3, matching the three groups expected by find_star(): no-offset, +offset, and -offset.
The loop over [visitid, unocculted_visitid] adds these fake satellite-spot frames for both the occulted and unocculted visits. However, since you suggest not to add satellite-spot frames to unocculted_visitid, I will change that

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: Sure! I will split it into two lines

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: Sure!

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: I will delete this comment

tests/e2e_tests/l1_to_l4_pol_e2e.py
zjwhale: I think the way we call find_star() is through l3_to_l4_pol.json recipe template. We may need to run different l3_to_l4_pol.json recipe templates for occulted visits and uncounted visits if we want to skip find_star for uncounted visits. Any thoughts on how to do that?

tests/e2e_tests/l2b_to_l4_e2e.py
zjwhale: yes, this variable only save pa_aper_values for science frames. the reason for this line is because the rotataion value is opposite to pa_aper value in headers

tests/e2e_tests/l2b_to_l4_e2e.py
zjwhale: Sure

tests/e2e_tests/l2b_to_l4_e2e.py
semaphoreP: can we add a comment that we exclude the reference star rotations?

tests/e2e_tests/l1_to_l4_pol_e2e.py
semaphoreP: ok thanks. just add some comments to explain this behavior. hard to understand from just looking at the code.

semaphoreP (approved):

PR #918, opened 2026-05-28
Center PSFs on the central pixel in the core throughput calibration cube

### LINKED ISSUE #368: Ensure core throughput PSFs are centered on a pixel
I believe the current CT PSFs are not centroided and interpolated so that they are centered on a pixel. For usability, it would be nice if they are. Interpolation errors should be minimal for well-sampled data.

Update the CT PSF generation to centroid on each PSF and interpolate it so it's centered on a pixel (the center is in the center of the stamp).

**Additional context**
For beyond v2.0

juliamilton:
## Describe your changes
Addresses #368. PSF stamps in the core throughput calibration cube are now centered on the center pixel. After cutting out the stamp around the centroid pixel, each stamp is shifted by the residual offset from the center (via cubic spline interpolation) so the PSF centroid lands on the central pixel. A function subpixel_center_stamp in corethroughput.py does the shift on both the data and DQ arrays, and generate_psf_cube calls it for each PSF. 

On the test side, test_cal_file's bit-for-bit cube comparison is relaxed so that it just checks whether the flux is conserved to within a tolerance (since stamps are no longer exact cutouts of the original image), and a new test test_psf_cube_subpixel_centered checks that the centroid of every PSF in the cube lands within 0.05 px of the central pixel.

## Type of change

#368 

### DIFF
--- a/corgidrp/corethroughput.py
+++ b/corgidrp/corethroughput.py
@@ -356,12 +356,17 @@ def generate_psf_cube(
                 continue
         except:
             pass
-
+        
+        # Sub-pixel centroid of this PSF on EXCAM, and the integer pixel it falls on
         cx = float(psf_loc[i_psf][0])
         cy = float(psf_loc[i_psf][1])
         cx_round = int(np.round(cx))
         cy_round = int(np.round(cy))
 
+        # Bounding box of size (2*n_pix_psf + 1)^2 around the rounded centroid
+        # pixel. Axis 0 = y (rows), axis 1 = x (cols). max/min clip to the frame
+        # edges so PSFs near the boundary produce smaller cutouts; these are
+        # padded back to uniform size below.
         idx_0_0 = max(cy_round - n_pix_psf, 0)
         idx_0_1 = min(frame.data.shape[0], cy_round + n_pix_psf + 1)
         idx_1_0 = max(cx_round - n_pix_psf, 0)
--- a/corgidrp/corethroughput.py
+++ b/corgidrp/corethroughput.py
@@ -46,7 +46,7 @@ def di_over_pil_transmission(
  
       Multiplying the counts of the pupil image by this factor translates them
       into equivalent counts of the direct imaging lens.
-
+ 
     Args:
       cfam_name (string): Filter in CFAM. For instance, '1F', '4A', '3B' or '2C'.
       cfam_version (int): version number of the filters (CFAM, pupil, imaging
@@ -63,7 +63,7 @@ def di_over_pil_transmission(
         lambda_pupil_nm = lambda_pupil_A / 10
     except:
         raise Exception('* File with the transmission of the pupil lens not found')
-    
+
     try:
         lambda_imaging_A, trans_imaging = np.loadtxt(os.path.join(here, 'data',
             'filter_curves', f'imaging_lens_v{cfam_version}.txt'),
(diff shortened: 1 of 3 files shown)


semaphoreP:
@juliamilton looks like the corethroughput e2e test also needs its tolerance changed to pass.

juliamilton:
> @juliamilton looks like the corethroughput e2e test also needs its tolerance changed to pass.

Updated. e2e test passes now for me locally.

corgidrp/corethroughput.py
```diff
@@ -235,15 +236,84 @@ def estimate_psf_pix_and_ct(
...
+            nan_mask.astype(np.float32),
+            shift=(dy, dx),
+            order=spline_order,
+            mode='constant',
+            cval=0.0,
+        ) > 0.5
+        shifted_data = np.where(shifted_nan_mask, np.nan, shifted_data)
+
+    # Shift the DQ array with nearest-neighbour interpolation
+    shifted_dq = ndi_shift(
```
semaphoreP, line 295: derotate_arr in combine.py uses an optional keyword `dq_round_threshold=0.05` and regular interpolation. Perhaps we should align with it to stay consistent with DQ interpolation handling?

corgidrp/corethroughput.py
```diff
@@ -401,16 +492,24 @@ def generate_ct_cal(
...
+    #       sub-pixel centered on their central pixel).
+    ct_excam = np.array([centered_psf_loc[:,0], centered_psf_loc[:,1], ct_est])
     ct_hdr = fits.Header()
-    # Core throughput values on EXCAM wrt pixel (0,0) (not a "CT map", which is
-    # wrt FPM's center 
-    ct_hdr['COMMENT'] = ('PSF location with respect to EXCAM (0,0) pixel. '
-        'Core throughput value for each PSF. (x,y,ct)=(data[0], data[1], data[2])')
+    ct_hdr['COMMENT'] = ('PSF stamp center on EXCAM (0,0). Each stamp is '
+        'sub-pixel-centered so the PSF centroid lies on the central pixel; '
+        'the recorded (x,y) is the integer EXCAM position of that central '
```
semaphoreP: does it make more sense to store the centroided location to sub-pixel accuracy instead? or does this break things

semaphoreP (commented):
overall looks good! I had one (possibly tedious) optional request, and one question about what coordinate we should record for the PSF position

corgidrp/corethroughput.py
juliamilton: Happy to do either, but wondering if the regular interpolation approach makes as much sense here as it does in de-rotation. Here we are shifting the entire array by (ideally) less than a pixel and we know where the bad pixels are to begin with. So I think we could preserve that location without needing to spread those to other nearby pixels by doing some averaging.

corgidrp/corethroughput.py
juliamilton: I realized that storing the new, shifted sub-pixel location actually breaks things. Shifting the PSFs just makes nice stamps, but doesn't actually change where the PSF was on EXCAM when the throughput was measured. For core throughput interpolation, I think we should save the PSF location as (original x, original y, ct). I've reverted back to this.

corgidrp/corethroughput.py
semaphoreP: ok, it's not a huge deal. let's not worry about it for now

corgidrp/corethroughput.py
semaphoreP: Well technically the exact location of the psf is the sub-pixel location, but this somehow messes with some downstream dependency, so we're using the approximate pixel-rounded location, and assuming CT changes very slowly (which we are assuming anyways), this difference between the sub-pixel and rounded location should not have a huge effect. Is that correct?

juliamilton: I think that's no longer correct after I made the recent commit. It is now storing the exact centroided sub-pixel position from the original PSF. Nothing downstream breaks as far as I can tell from testing.

semaphoreP (approved):

PR #919, opened 2026-05-29
l1_to_nd_spec_cal_e2e uses sweet spot dataset

### LINKED ISSUE #878: Implement "sweet spot" dithering aspect of spectroscopy ND filter calibration product and update VAP tests
According to CGI-REQT-5477 and CGI-REQT-5478, the spectroscopy ND filter calibration should be made using a "sweet spot" dataset that uses data collected at different FSM positions. The full text of the requirements is found below. 

The existing spectroscopy ND filter calibration code should be updated to accommodate this, analogous to how the imaging mode ND filter calibration code is done. 

The existing l1_to_nd_filter_spec_e2e.py test should be updated to be in VAP format and should verify that these requirements are met.

aneeshb97:
## Describe your changes
Sweet spot dataset (brightstandstar dataset) from DIP includes images with ND filter at different (FSMX,FSMY) combinations. We modify the nd spec cal e2e to use this new dataset. Additionally, minor changes to dataformat e2e test.

## Type of change

#878 


aneeshb97:
This is incorrect, just wanted to have a PR up

juliamilton:
Thanks! This is generally looking good to me. Could you also add an additional test that tests the interpolation? This could be done by adding something to the existing l1_to_nd_spec_cal_e2e test that: 

- loads the sweet spot dataset, 
- creates a new observation at a different FPAM_H/ FPAM_V position (near to/ within the existing dithers)
- calls calculate_od_spec_at_new_location()
- prints OD(λ) and checks that OD values are reasonable

corgidrp/data.py
```diff
@@ -4162,21 +4159,102 @@ def __init__(self, data_or_filepath, err=None, dq=None,
...
+        interpolating.  If those keywords are absent the offsets default to 0,
+        so the call is also correct for full-frame (uncropped) images.
+
+        Args:
+            x (float): x pixel location in the science frame's own coordinate system.
+            y (float): y pixel location in the science frame's own coordinate system.
+            method (str): only "nearest" supported currently.
+            wave_grid (list of float or np.array, optional): Wavelength grid specfied
+            by user. Default to None (wavelength grid taken from image hdu 'SPEC_WAVE').
+            image (corgidrp.data.Image, optional): the science Image whose pixel
```
semaphoreP: should update the docstring that passing in this parameter will override the wave_grid keyword.

corgidrp/data.py
```diff
@@ -4162,21 +4159,102 @@ def __init__(self, data_or_filepath, err=None, dq=None,
...
+            od_stack = []
+            for i in range(self.od_spectra.shape[0]):
+                od = self.od_spectra[i,:]
+                wave = self.wavelengths[i,:]
+                if not np.allclose(wave, common_wave, atol=0.01):
+                    remap_od  = interp1d(wave, od,    kind='linear',
+                                        bounds_error=False, fill_value=np.nan)
+                    od    = remap_od(common_wave)
+                    
+                od_stack.append(od)
```
semaphoreP: This code seems quite duplicated. Can we condense the if statement, and reduce the amount of duplicated code?

corgidrp/l3_to_l4.py
```diff
@@ -1017,18 +1019,23 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
...
 
     all_science_frames = []
 
-    for visid in visitid:
-        if subtract_no_offset_frames:
-            
-            satspot_subset = satspot_dataset[int(np.nonzero(visitid == visid)[0].item())]
+    for keyword in keywords:
+
+        matched_index = [i for i, key in enumerate(keywords) if key == keyword]
```
semaphoreP: is this logically the same as `for matched_index, keyword in enumerate(keywords):`

corgidrp/l3_to_l4.py
```diff
@@ -1083,7 +1090,7 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
             science_subset = offset_dataset
         
         if with_science:
-            science_subset = science_dataset[int(np.nonzero(visitid == visid)[0].item())]
+            science_subset = science_dataset[int(matched_index[0])]
```
semaphoreP: there's no guarantee that the science datasets are ordered in the same way as the satspot datasets. We should really be matching science_dataset visitids to the given satspot visitid. (This probably also means not overriding the `keyword` variable when splitting the science_dataset as well.

corgidrp/nd_filter_calibration.py
```diff
@@ -532,26 +532,120 @@ def create_nd_filter_cal(stars_dataset,
...
+        
+        od_stack = []
+        for i in range(od_val.shape[0]):
+            od = od_val[i,:]
+            wave = wav_val[i,:]
+            if not np.allclose(wave, common_wave, atol=0.01):
+                remap_od  = interp1d(wave, od,    kind='linear',
+                                    bounds_error=False, fill_value=np.nan)
+                od    = remap_od(common_wave)
+
```
semaphoreP: can we reuse the code in `SpecNDFilter.interpolate_od()` in data.py and/or merge the two?

corgidrp/nd_filter_calibration.py
```diff
@@ -532,26 +532,120 @@ def create_nd_filter_cal(stars_dataset,
...
+        for i in range(len(common_wave)):
+            monochromatic_sweet_spot_data = np.column_stack((
+                                            od_array[:,i],
+                                            x_val[:,i],
+                                            y_val[:,i]
+                                            ))
+            # Interpolate OD at that new location
+            interp_od[i] = interpolate_od(monochromatic_sweet_spot_data, x_adj, y_adj)
+
+    # TO DO: add in interpolated od into the header of the file and re-save? determine how the OD
```
semaphoreP, line 575: should we have a wrapper function that divides out the ND filter OD from the 1-D spectrum? basically return a spectrum corrected for ND filter strength

corgidrp/nd_filter_calibration.py
```diff
@@ -532,26 +532,120 @@ def create_nd_filter_cal(stars_dataset,
...
+    # Metadata checks
+    if (hdr.get('FPAMNAME') != ref_fpam_name or 
+        abs(hdr.get('FPAM_H') - ref_fpam_h) > 1.2 or    # within non-repeatability tolerance of 1.2 um
+        abs(hdr.get('FPAM_V') - ref_fpam_v) > 1.2 or 
+        hdr.get('CFAMNAME') != ref_cfam_name or
+        hdr.get('DPAMNAME') != ref_dpam_name):
+        raise ValueError(
+            f"Inconsistent FPAM/CFAM/DPAM header values in target {target} for file {frame}!"
+        )
+
```
semaphoreP, line 620: why is this check in this function, and not in the parent function that calls it (`create_nd_filter_cal_spec`)? It seems unnecessary to pass in 5 extra arguments, just to do this check.

corgidrp/nd_filter_calibration.py
```diff
@@ -592,8 +686,187 @@ def compute_od_spectrum_for_frame(frame, sf_cal, calspec_filepath):
...
+    """
+    Process bright star files for one target to compute optical density (OD)
+    and (x, y) centroids for each dithered observation.
+    Checks that FPAM keywords are consistent across all files.
+    
+    Additional photometry options are passed via phot_kwargs.
+    This allows users to override default settings for functions like aper_phot.
+    
+    Parameters:
+        target (str): The target star name.
```
semaphoreP, line 671: isn't this value just in the header of the Images of `files`?

corgidrp/nd_filter_calibration.py
```diff
@@ -616,12 +889,14 @@ def create_nd_filter_cal_spec(stars_dataset, spec_fluxcal=None, calspec_files=No
 
     Args:
         stars_dataset (corgidrp.data.Dataset): L3 frames with SPEC extensions.
+        od_raster_threshold (float): Threshold for flagging OD variations.
```
semaphoreP: for complicated reasons, spec_fluxcal needs to be the first keyword of this function.

corgidrp/nd_filter_calibration.py
```diff
@@ -592,8 +686,187 @@ def compute_od_spectrum_for_frame(frame, sf_cal, calspec_filepath):
...
+
+        od_all_spectra.append(np.mean(od_spectra, axis=0))
+        wave_grids.append(np.mean(wave_spectra, axis=0))
+        od_all_err.append(np.sqrt(np.nansum(oderr_spectra ** 2, axis=0)) / len(oderr_spectra))
+        x_values.append(x_center)
+        y_values.append(y_center)
+        xerr_values.append(xcen_err)
+        yerr_values.append(ycen_err)
+
+    # If multiple bright frames, remap to a common wavelength grid and average
```
semaphoreP: are we not doing what's in the comments here: remapping to a common wavelength grid?

corgidrp/nd_filter_calibration.py
```diff
@@ -592,8 +686,187 @@ def compute_od_spectrum_for_frame(frame, sf_cal, calspec_filepath):
...
+    #common_wave = wave_grids[0]
 
-def create_nd_filter_cal_spec(stars_dataset, spec_fluxcal=None, calspec_files=None, outputdir=None):
+    if len(od_all_spectra) == 1:
+        od_stack  = od_all_spectra[0]
+        wave_stack = wave_grids[0]
+        err_stack = od_all_err[0]
+    else:
```
semaphoreP: I'm not sure what's going on in this else statement. it seems the equivalent of doing nothing. you seem to be turning a list back to itself.

corgidrp/nd_filter_calibration.py
```diff
@@ -592,8 +686,187 @@ def compute_od_spectrum_for_frame(frame, sf_cal, calspec_filepath):
...
+        aggregated_sweet_spot_err (numpy.ndarray, optional): The aggregated N×N_wave×4 array containing 
+            OD_err as a function of wavelength in the format [wave_err, OD_err, x_err, y_err].
+            Defaults to None.
+
+    Returns:
+        tuple:
+            NDSpectroscopy: The generated ND spectroscopy sweet spot dataset.
+    """
+    final_sweet_spot_data = aggregated_sweet_spot_data.copy()
+    try:
```
semaphoreP: do you just want to do `if aggregated_sweet_spot_err is not None`?

tests/e2e_tests/l1_to_nd_filter_spec_e2e.py
```diff
@@ -322,12 +343,64 @@ def run_nd_filter_spec_e2e(l1_datadir, processed_cal_path, outputdir):
...
+    bright_star_filelist = sorted(f for f in l1_filelist if fits.getheader(f, ext=1)['FPAMNAME'] == 'ND225')
+    faint_star_filelist = sorted(f for f in l1_filelist if fits.getheader(f, ext=1)['FPAMNAME'] != 'ND225')
+    changed_visitid = False
+    if fits.getheader(bright_star_filelist[0])['VISITID'] ==  fits.getheader(faint_star_filelist[0])['VISITID']:
+        changed_visitid = True
+        brightdata_visitid = fits.getheader(faint_star_filelist[0])['VISITID']
+        for f in bright_star_filelist:
+            fits.setval(f, 'VISITID', value = '0200001001001002001', ext=0)
+    
+    # Changing TARGET if original TARGET not in calspec list
```
semaphoreP, line 371: question: this is needed because the default stars aren't in the calspec lookup right now?

semaphoreP (commented):
overall looks good. I have mostly style comments, and a couple suggestions.

tests/e2e_tests/l1_to_nd_filter_spec_e2e.py
aneeshb97: yes

corgidrp/nd_filter_calibration.py
aneeshb97: you mean first keyword after the stars_dataset?

corgidrp/nd_filter_calibration.py
aneeshb97: legacy comment, will remove this

corgidrp/nd_filter_calibration.py
aneeshb97: Again, remnants of legacy code. Retained if we wanted to do common wave interpolation here but will remove

corgidrp/nd_filter_calibration.py
aneeshb97: it mirrors the equivalent function for ND imaging. I assumed the check might be a requirement, since this is the function that handles individual frames

corgidrp/l3_to_l4.py
aneeshb97: okay

corgidrp/nd_filter_calibration.py
semaphoreP: yes, that's right. it should be stars_dataset, spec_fluxcal=None, ...

corgidrp/nd_filter_calibration.py
semaphoreP: Ok, that's fine then

corgidrp/nd_filter_calibration.py
aneeshb97: again, mirrors ND imaging. Can remove if don't want it here

corgidrp/data.py
aneeshb97: this will be somewhat ugly. 
Basically, if there is no user-provided input spec image or wave_grid, it remaps the ODs of each dither onto wave_grid of dither 1. So we do not need to interpolate OD of dither 1, but need to do for others.

If there a user-provided input spec image or wave_grid, ODs of each dither have to be remapped.

We could always make it such that if there's no user-provided input spec image or wave_grid, the interpolate_od errors out

aneeshb97: Can clean this up sure, but I think it's good to visualize the no provided wave_grid/image and user-specified wave_grid/image cases separately

corgidrp/data.py
aneeshb97: Passing in which parameter? Wave_grid takes priority, followed by image hdu 'SPEC_WAVE'

corgidrp/nd_filter_calibration.py
semaphoreP: no it's fine.

corgidrp/data.py
semaphoreP: Why can't we just always interpolate regardless. An interpolation of a wavelength grid onto itself will just return the original values (and this doesn't take a lot of time).

corgidrp/l3_to_l4.py
aneeshb97: I initially thought it provides an extra check, but now I think maybe not? Changing the code to the above gives the same outputs as before

corgidrp/data.py
aneeshb97: okay!

corgidrp/nd_filter_calibration.py
aneeshb97: done!

corgidrp/nd_filter_calibration.py
aneeshb97: you mean a method function attached to the NDSpec class that takes in a 1-D spectrum and gives a spectrum corrected for ND filter OD?

semaphoreP: it can also just be a function in this .py file that takes a L4 spectrum file and a ND filter acalibraiton and corrects for the OD

corgidrp/data.py
semaphoreP: oh you're right. never mind.

corgidrp/nd_filter_calibration.py
```diff
@@ -679,76 +906,161 @@ def create_nd_filter_cal_spec(stars_dataset, spec_fluxcal=None, calspec_files=No
...
+            fpamfsamcal = fpamfsamcal, ndspectroscopy_dataset = ndspectroscopy_dataset, wave_grid=wave_grid)
+    
+    corrected_image = clean_spec_image.copy()
+
+    spec = corrected_image.hdu_list['SPEC'].data
+    spec_wave = corrected_image.hdu_list['SPEC_WAVE'].data
+    spec_err = corrected_image.hdu_list['SPEC_ERR'].data
+    
+    #Interpolate SPEC values on new wavelength grid
+    if not np.allclose(spec_wave, common_wave, atol=0.01):
```
semaphoreP, line 1041: I have a preference that we interpolate the od_spec onto the science data wavelengths rather than the other way around (general principle of minimizing interpolation on the science data).

semaphoreP (commented):
one small comment about the new function

corgidrp/nd_filter_calibration.py
aneeshb97: By default the OD_spec is interpolated on the science data wavelengths (image.hdu_list['SPEC_WAVE']), so common_wave and spec_wave will be the same. 
The interpolation on science data happens ONLY if the user specifies a custom wavelength grid that they want for their output image. Then OD_spec and science data are both interpolated onto that custom wavelength grid (i.e., common_wave).
We can just not have a custom wavelength grid as keyword for this function. If users want to interpolate the science onto a different grid, they are free to do so. calculate_od_spec_at_new_location() allows custom wavelengths for OD_spec anyway, so OD correction won't be an issue

semaphoreP: oh gotcha. I missed that when I looked at it.

aneeshb97: Let me know if you would just prefer to remove the wave_grid keyword for the new function

semaphoreP:
Run e2e tests one final time before merge

corgidrp/nd_filter_calibration.py
semaphoreP: no it's fine. you can maybe add some comments about this logic only happens if wave_grid is passed in, because it's not obvious from reading the code (since it depends on the behavior of sub-functions).

aneeshb97: added some comments!

semaphoreP (approved):

Issue #923, opened 2026-06-01
need to be able to change step function arguments in production environment


nasavbailey:
In the production environment, there is no way to change the majority of step function arguments except to manually run a custom recipe. This is because the entire DRP repo is located within the production docker container, including the "recipe_templates" directory. Changing the contents of the production docker container requires release + test, which is an unworkably large overhead.

Additionally, there may be instances where the arguments to a given step function need to change depending on the recipe (eg: cosmic ray rejection parameters may be different for unilluminated vs illuminated images). 

We need a way to change the default arguments on a recipe-by-recipe basis.

nasavbailey:
Proposed solution: allow an external directory that can contain recipe templates that override the defaults. See details here: https://docs.google.com/presentation/d/1PXyab-oyWcDf2OyeBSwOlgzVjlwmNADr/edit?slide=id.p1#slide=id.p1

maxwellmb:
clarification: The preferred solution is a new user_templates directory. (recipes are basically dictionaries)

nasavbailey:
doh, typo fixed in above (recipe templates, *not* recipes themselves)

adrienmaillard:
I think I get most of what is needed. You want to have corgidrp look for recipe templates overrides in some user folder defined either by an env variable (and let's say defaulting to ~/.corgidrp/user_templates). They should match the name of the recipe they override. They completely replace the recipe, we don't merge them. You want the users to change only the arguments so the steps should match between the original and the override. Right ? 

What I don't get is this git based review process for overrides. You would have a separate repo and people would pull from this repo in their ~/.corgidrp/user_templates ?

adrienmaillard:
Also, I don't get what the production env looks like besides the docker running on a server. There's like one big deployment that processes all the images and then some user-deployed instances too ?

maxwellmb:
The users can completely replace the recipes, and there's an optional check to make sure the steps are the same. Defualts to false, but can be set in th corgidrp.cfg file. 

Rather than environment variable I think we adjust the corgidrp.cfg file too for the directory, if not default. Don't want to have too many settings in too many places. 

Don't worry about that second git thing, that's outside of corgidrp, to be set up on the DAE.

maxwellmb:
Overall seems like you've got it!

maxwellmb:
> Also, I don't get what the production env looks like besides the docker running on a server. There's like one big deployment that processes all the images and then some user-deployed instances too ? 

[...]

PR #925, opened 2026-06-03
Allow default recipes to be overriden by user recipes

adrienmaillard:
## Describe your changes

If it does not exist already, corgidrp will now create `~/.corgidrp/user_templates` where users can put recipe overrides. A recipe overrides must have the same name as an existing recipe. When a recipe is used, corgidrp will look for a recipe override and otherwise will fallback to the default one provided with corgidrp. 

There is a boolean in the configuration stating whether the override recipes must have the same steps or not. If they must have the same steps, it will be checked, and arguments can still differ (that's the original intent of this feature I understand). In ops, this boolean will be set to True (in ops.py). Otherwise, it's set to False. 

## Type of change

Solves #923 

### DIFF
--- a/corgidrp/walker.py
+++ b/corgidrp/walker.py
@@ -128,14 +128,14 @@ def _load_recipe_template(recipe_filename, user_templates_dir=None):
         f"({user_templates_dir}) or default templates ({recipe_dir})"
     )
 
-def _validate_template_structure(user_template, default_template, recipe_filename):
+def _validate_template_structure(user_template, default_template_path, recipe_filename):
     """
     Validate that user template has same step names/order as default.
     Raises ValueError if mismatch found when enforce_template_structure=True.
 
     Args:
         user_template (dict): User-provided template
-        default_template (dict): Default template from repo
+        default_template_path (str): Path of the default template from repo
         recipe_filename (str): Template name for error messages
 
     Raises:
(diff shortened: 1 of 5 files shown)


maxwellmb (changes requested):
This is in pretty good shape, thanks! 

It would be great if ops.py could change (via a keyword argument) the user_templates directory in the config. It would also be great if we could have a test of manually setting the user_templates directory to something that isn't the default value. I think this will make things easier for the DAE/DIP folks. 

Jason will likely also want to take a quick look.

adrienmaillard:
I see ops.py is called only by tests right now. You want me to add an optional argument to `step_1_initialize()` to change the `user_templates` dir ?

maxwellmb:
Yes please!

adrienmaillard:
I added the argument and a test in ops.py. I factored some code to reuse the existing test in ops.py.

corgidrp/walker.py
```diff
@@ -112,18 +180,44 @@ def walk_corgidrp(filelist, CPGS_XML_filepath, outputdir, template=None):
...
+                if not os.path.exists(default_template_path):
+                    raise FileNotFoundError(
+                        f"Cannot validate user template '{os.path.basename(recipe_filepath)}': "
+                        f"no corresponding default template found at {default_template_path}. "
+                        f"User templates must have same filename as a default template when "
+                        f"enforce_template_structure=True."
+                    )
+                with open(default_template_path, 'r') as f:
+                    default_template = json.load(f)
+                _validate_template_structure(template, default_template, os.path.basename(recipe_filepath))
```
semaphoreP: to reduce repetitive code, should we just pass in the default_template_path into `_validate_template_structure` and do all these checks in there?

semaphoreP (commented):
This looks good to me. I just have one comment to reduce the comment of repetitive code in walker.py

corgidrp/walker.py
adrienmaillard: I have done the change you proposed, can you take a look ?

semaphoreP: looks good

semaphoreP (approved):
Looks good to me. running the e2e tests just to check we didn't break anything

semaphoreP:
e2e tests pass for me. @maxwellmb let us know what you think.

maxwellmb (approved):

Issue #176, opened 2024-08-30
ability for desmear step function to determine if desmearing is worthwhile based on inputs


kjl0025:
**Is your feature request related to a problem? Please describe.**
For calibrating a synthetic master dark, some frames in the input dataset may need to be desmeared while others may not.  An edit to the structure of the walker could be done to handle this, or we could add optional inputs to the desmear step function to handle this.  This would also be useful for a regular (non-calibration) dataset as well; instead of having the user separately figure out if desmearing is appropriate before specifying a recipe for the dataset, the step function could handle it automatically.

**Describe the solution you'd like**
There could be inputs for expected values, in e- units, for mean FPN (or perhaps max value for FPN since it may have a mean of 0), read noise, mean CIC, and mean dark current.  If these are not specified, they could be taken from a specified synthesized master dark if appropriate.  (In the context of a recipe for making a synthesized master dark, the user would specify the numbers in the aforementioned inputs.)  If none of these are specified, the step function makes no recommendation.  Based on these inputs and the EM gain and exposure time used for a given frame, the step function would determine how far above the values to be subtracted in the desmearing are compared to the non-smeared noise.  If significantly above (determined by an optional input, number of standard deviations which could default to 2 maybe), the desmearing is recommended.  Then there could be an override flag (True or False) to override the step function's recommendation to smear or not.  

**Additional context**
If there are cases where a user wants to override the desmear step function's recommendations for only a subset of frames in a dataset, then I think a change to the walker's structure would be needed.  This would also be true for any other step function for which a user wants different keyword arguments per frame in an input dataset.

semaphoreP:
Is this something you want to try to get in this release? Or is this a longer-term project?

kjl0025:
It's not necessary for this release.  Just a longer-term project.

kjl0025:
Additional notes:
Any component linearly proportional to exposure time will have a smear to it (i.e., photon flux and dark current), while other components will not (CIC, FPN, read noise).

**synthetic master dark:**  
1.  It is made from a mix of frames at various exposure times, so in theory, one should apply desmearing (if appropriate, which is unlikely since the dark current is typically <= 1e-3 e- per pixel, unless the exposure time is large enough) to the stacks before constructing the synthetic master dark.  
2. For a given illuminated set with exposure time t, smear should be APPLIED to the dark current map according to t before synthetic master dark is constructed and subtracted since the illuminated frame also contains this smeared dark current in it.

[...]

PR #931, opened 2026-06-08
Median full cosmic

### LINKED ISSUE #821: Update HDU dimensions for trap-pump, astrometry and core throughput calibration in documentation
HDU dimensions are not fixed for the trap-pump calibration. They depend on the number of traps found. And other calibrations may be in the same situation. For example, CoreThroughput's dimensionality depends on the number of PSFs, though there may be a planned default based on the observation plan for that cal. Also, AstrometricCalibration depends on the order of the Legendre polynomial used for field distortion, though there is a default value for that. 

Would be good to add a note in documentation to reflect this similarly to what is done in confluence: collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/161613668/Trap+Pumping+Calibration+DRP+Version+R3.0.2#TrapPumpingCalibrationDRPVersionR3.0.2-FITSFileExtensionsandFormat where array size is (n, 10) for n traps and make sure this doesn't break the data format automatic checks that are made when running tests.

### LINKED ISSUE #836: disallow PC frame binning for the case of PC master dark
Currently, Dark products vary in the number of dimensions for an HDU:

Image HDU array: If analog master dark, (r,c), where this is (1024,1024) for image area or (1200,2200) for full frame.  
If photon-counted master dark, (N,1024,1024), where N is 1, or larger if frame binning is used.
ERR HDU array: (1,r,c) or (1,N,1024,1024)
DQ HDU array: (r,c) or (N,1024,1024)

This is problematic for SSC's check that each calibration product have a fixed number of HDU dimensions.

When get_pc_mean() is run in dark mode, frame binning should be disallowed so that the number of HDU dimensions will match that of analog master darks.  In fact, forcing the PC master dark to be made from all the input darks is all the better for the fidelity of the PC master dark.

**Additional context**
N/A

### LINKED ISSUE #903: desmear function desmear cosmic rays which are not smeared
**Describe the bug**
The desmear function currently desmears cosmic rays which are not physically smeared during read out.

**To Reproduce**
Running frame through desmear function with cosmic ray hits.

**Expected behavior**
Desmearing should ignore cosmic rays and replace those affected pixels' values with values from nearest neighbors as a proxy for the flux it would have received in the absence of cosmic rays.  

**corgidrp version:**
 DRP v5

**Additional context**
Replacing with values from nearest neighbors is the best we can do since we can't know what flux would have fallen on the affected pixels, but desmearing assuming nearest-neighbor values is more accurate than assuming saturation-level values.

kjl0025:
## Describe your changes
Auto-decide whether to smear or not based on values in frame, desmearing only happens in image area, desmear function does not desmear cosmic rays now, added desmear to k gain/nonlin cal recipes, improved nonlin cal based on results from CAR121 tests, improved darks.py based on CAR119 tests, cosmic ray flagging no longer restricted to image area for full-frame mode, cosm_filter increased from 1 to 2 to avoid false positive flags, RAM-heavy simulation and testing for trap pump cal, filename collisions avoided for trap pump frames in rename_files_to_cgi_format, photon-counted master darks no longer binned so that all Dark products agree in HDU dimensionality, note added to index.rst clarifying that HDU dimension sizes are subject to change after L2b but that the number of dimensions is not.

## Type of change

#903 , #867  , #836 , #821 , #176 


corgidrp/data.py
```diff
@@ -4633,6 +4634,39 @@ def get_bit_to_flag_map():
...
+    while nan_inds.size < temp_dq.size:
+        bad_inds = np.where(temp_dq == 0)
+        temp_dq[bad_inds] = np.nan
+        highest_power = np.floor(np.log(temp_dq)/np.log(2))
+        selected_inds = np.where(2**(highest_power) == val)
+        selected_dq[selected_inds] = val
+        temp_dq = temp_dq - 2**(highest_power)
+        nans = np.where(np.isnan(temp_dq))
+        nan_inds = nans[0]
+    return selected_dq
```
semaphoreP, line 4661: I found this a bit confusing to understand. Can we just do:

```
selected_dq = (dq & val)
```

I think this would return an array with values either = 0 or val.

corgidrp/l2a_to_l2b.py
```diff
@@ -542,34 +542,160 @@ def desmear(input_dataset, detector_params):
...
+                columnsum = 0
+                for s in range(r+1):
+                    columnsum = columnsum + rowreadtime_sec/exptime_sec*((1
+                    + rowreadtime_sec/exptime_sec)**((s+1)-(r+1)-1))*image_data[s,:]
+                smear[r,:] = columnsum
+            if image_frame_flag:
+                orig_data = input_dataset[i].data.copy()
+            else:
+                orig_data = slice_section(input_dataset[i].data.copy(), arrtype, 'image', detector_regions)
+            image_data -= smear
```
semaphoreP: is orig_data used here?

corgidrp/l2a_to_l2b.py
```diff
@@ -542,34 +542,160 @@ def desmear(input_dataset, detector_params):
...
+            orig_data = input_dataset[i].copy().data
+        else:
+            orig_data = slice_section(input_dataset[i].copy().data, arrtype, 'image', detector_regions)
+        if desmear_flag:
+            image_data[rows, cols] = orig_data[rows, cols] - smear[rows, cols]
+        else:
+            image_data[rows, cols] = orig_data[rows, cols]
 
-    return data
+    history_msg = "Desmear function applied to data"
```
semaphoreP: should update the history message to whether desmear was actually applied or not, possibly based on autoselection.

corgidrp/mocks.py
```diff
@@ -4555,35 +4555,48 @@ def rename_files_to_cgi_format(list_of_fits=None, output_dir=None, level_suffix=
...
+            # 2 to 3 frames per config
+            filename = os.path.split(file)[-1]
+            if '_config' in filename:
+                config_offset = int(filename[filename.find('_config')+7])+13
+            else:
+                config_offset = 0
+            if '_fr' in filename:
+                fr_offset = int(filename[filename.find('_fr')+3])+7
+            else:
+                fr_offset = 0
```
semaphoreP, line 4602: can you add a bit more comment to explain what's going on here. it's not obvious what's going on

corgidrp/photon_counting.py
```diff
@@ -173,7 +174,11 @@ def get_pc_mean(input_dataset, pc_master_dark=None, T_factor=None, pc_ecount_max
     if bin_size > len(input_dataset):
         raise ValueError('bin_size must be less than the number of frames in input_dataset.')
         
-    num_bins = len(input_dataset)//bin_size 
+    if inputmode == 'darks':
+        num_bins = 1
+        bin_size = len(input_dataset)
```
semaphoreP, line 172: maybe if the user set bin_sizes as something that's not None, we raise a warning that we're overriding it?

corgidrp/pump_trap_calibration.py
```diff
@@ -24,7 +24,8 @@
 
 import corgidrp.check as check
 from corgidrp.data import TrapCalibration, typical_bool_keywords, typical_cal_invalid_keywords
-
+from datetime import date
+#from memory_profiler import profile
```
semaphoreP: can we remove debug code?

docs/source/data_formats/dark.rst
```diff
@@ -7,10 +7,9 @@ Dark Data Product
 FITS HDU Structure
 ------------------
 
-Image HDU array: If analog master dark, (r,c), where this is (1024,1024) for image area or (1200,2200) for full frame.  
-If photon-counted master dark, (N,1024,1024), where N is 1 or larger if frame binning is used.
-ERR HDU array: (1,r,c) or (1,N,1024,1024)
-DQ HDU array: (r,c) or (N,1024,1024)
+Image HDU array: (r,c), where this is (1024,1024) for image area or (1200,2200) for full frame.
```
semaphoreP: can you put new lines between paragraphs. otherwise, this will display as a single line, which is hard to read.

tests/test_desmear.py
```diff
@@ -65,7 +67,69 @@ def test_desmear():
...
+    image1 = Image(smeared_frame, pri_hdr = prhd, ext_hdr = exthd, err = err, dq = dq)
+    dataset_smeared = Dataset([image1])
+    dataset_desmear = desmear(dataset_smeared, detector_params)
+    # slightly bigger discrepancy than previous tolerance b/c cosmic-flagged pixels take on the 
+    # value of nearest neighbors, which may be in a different row, which would include a different 
+    # amount of smear than the original row
+    assert(np.max(np.abs(dataset_desmear.all_data[0] - unsmeared_frame)) < 1e-7)
+
+    # test same thing with auto-decide on
+    dataset_desmear = desmear(dataset_smeared, detector_params)
```
semaphoreP: is auto-detect actually turned on?

semaphoreP (commented):
ok some small comments below

corgidrp/data.py
kjl0025: I wanted to extract DQ values that contained val.  E.g., a saturated cosmic ray would be 32+128 = 160.

kjl0025: So this breaks down any sum of powers of 2 to see if val is contained.

corgidrp/l2a_to_l2b.py
kjl0025: Oops, I took that out in my commit (I'm commenting retroactively).

corgidrp/l2a_to_l2b.py
kjl0025: Since the update function is for the Dataset class, I changed the history to reflect whether auto_decide was on or off (in my last commit), but I let the DESMEAR header say whether desmearing occurred or not.

corgidrp/mocks.py
kjl0025: I added clarification

corgidrp/photon_counting.py
kjl0025: I didn't commit that change b/c I was going to say that the doc string clarifies that, but actually, I think that would be good. I'll add it.

corgidrp/pump_trap_calibration.py
kjl0025: removed

docs/source/data_formats/dark.rst
kjl0025: will do in the next commit

tests/test_desmear.py
kjl0025: fixed

corgidrp/data.py
semaphoreP: I'm pretty sure `160 & 128 = 128`, doesn't that do what you are looking for?

kjl0025: Oh, I take that back, you are correct!  I didn't know that worked for numbers as well.  I just pushed a change and confirmed the desmear test still passes.

kjl0025: I also had to change around the KGAINPAR and RN header checks in desmear.  I had forgotten that KGAINPAR is present even at the L1 level, whereas RN is the one that doesn't get added until after convert_to_electrons().

corgidrp/photon_counting.py
semaphoreP: Can you update the unit test to not call bin_size when making darks to avoid your new warning. 

```
tests/test_photon_counting.py::test_pc
tests/test_photon_counting.py::test_pc
tests/test_photon_counting.py::test_pc
tests/test_photon_counting.py::test_pc_subsets
  /home/runner/work/corgidrp/corgidrp/corgidrp/photon_counting.py:176: UserWarning: bin_size is not meaningful if input_mode='darks'. All frames will be used in making the PC master dark.
    warnings.warn('bin_size is not meaningful if input_mode=\'darks\'. All frames will be used in making the PC master dark.')
```

kjl0025: I actually test that the number of frames of the PC master dark is independent of whether the user specifies bin_size, so I just pushed a change to catch the warning.  Hopefully that suffices?

semaphoreP: ok, but looks like there's still some more warnings.

semaphoreP (approved):

Issue #875, opened 2026-05-07
Implement a more robust fit of  wave_zeropoint


JuergenSchreiber:
**Is your feature request related to a problem? Please describe.**
There are cases that the fit of the wave_zeropoint does not converge or is found somewhere in the noise.

**Describe the solution you'd like**
To be investigated, but suppose it has to do with the initial guesses and/or template spectra

**Additional context**

JuergenSchreiber:
There is something rotten with the centroid fitting of spectra for the dispersion model, but up to now I could not find a difference to the old spectroscopy implementation. The whole fitting and guess parameter finding is very complex, so I need some time to find something out.

semaphoreP:
Added in PR #938

PR #938, opened 2026-06-11
Introduce crop for spec calfile generation in l2b_to_xxx.json recipe templates

JuergenSchreiber:
## Describe your changes
To avoid fits of spectra going wild we need to crop the 1024x1024 spec frames before fitting e.g. the wavelength zeropoint. This is done in the recipes for the e2e tests.
But crops should be avoided in the old e2e tests using small 80 by 80 template spectra simulations.

## Type of change

#875


corgidrp/l2b_to_l3.py
```diff
@@ -392,7 +392,8 @@ def crop(input_dataset, sizexy=None, centerxy=None):
             cropped_frame_err[:,below_pad:sizexy[1]-above_pad,
                                left_pad:sizexy[0]-right_pad] = frame.err[:,y1+below_pad:y2-above_pad,
                                                                          x1+left_pad:x2-right_pad]
-            cropped_frame_dq = np.full(sizexy[::-1],np.nan)
+            cropped_frame_dq = np.full(sizexy[::-1],np.nan).astype(int)
+            print(cropped_frame_dq)
```
semaphoreP: debug code?

tests/e2e_tests/spec_linespread_e2e.py
semaphoreP, line 1: is it possible to keep this e2e test run with default parameters instead of custom parameters? do we need to modify the input data it uses?

tests/e2e_tests/spec_prism_disp_e2e.py
semaphoreP, line 1: is it possible to keep this e2e test run with default parameters instead of custom parameters? do we need to modify the input data it uses?

semaphoreP (commented):
Hi Juergen, i'm not going to include this in this release as I'd like to discuss whether we can keep our e2e tests not using custom parameters.

corgidrp/l2b_to_l3.py
JuergenSchreiber: yes, I think the crop code might be improved. Is it a good idea to initialize the dq with NaN floats?

tests/e2e_tests/spec_linespread_e2e.py
JuergenSchreiber: As far as I understand Neil, he wants to update the simulated template files which are also used here as l2b input, which crop tries to crop and fills with NaNs although they are very small (so maybe another thing that crop can be improved on.). Anyway, aren't these e2e tests obsolete anyhow, since we have the e2e starting from simulated L1 files?

corgidrp/l2b_to_l3.py
semaphoreP: we can initialize with 0's if preferred.

tests/e2e_tests/spec_linespread_e2e.py
semaphoreP: does @juliamilton know which spec e2e tests are used for VAP testing? if they are being used for VAP testing, we would need to figure out if we could transition to the other one. Same for the spec_prsim_disp_e2e I guess.

juliamilton: Yes, the following e2e tests (and one unit test file) are used for VAP testing:

spec_prism_disp_e2e.py
test_spec.py
spec_linespread_e2e.py
l1_to_l3_spec_e2e.py
spec_l3_to_l4_psfsub_e2e.py
spec_l3_to_l4_noncoron_e2e.py
l4_to_tda_spec_vap_e2e.py
l1_to_nd_filter_spec_e2e.py

tests/e2e_tests/spec_prism_disp_e2e.py
JuergenSchreiber: ok, I found a solution to keep  crop for the small frames of the test spectroscopy. I add the correct LSAMNAME "SPEC" and the center values EACQ_COL/ROW to the header and I convert the padded NaNs to 0 since ndi.median_filter does not like NaNs. The thing I don't understand in crop: We input frames (81x81) smaller than the wished cropped output size (125x125), crop pads this with NaNs. Is this the desired behaviour? shouldn't it just return the small frames unchanged?

semaphoreP: I'm not opposed to crop having the behavior of fixing the output size (both subtracting and adding to the image to reach the desired image size). Could you just pad the input images to be bigger (add 0's to the 81x81 image edges) if you don't want to deal with the undocumented behavior of feeding a smaller image into crop?

tests/e2e_tests/spec_linespread_e2e.py
semaphoreP: thanks, it looks like we can't drop them, so we should put a little bit of work to keep them working as they were intended. I suggest doing something like padding the input data to make it bigger.

tests/e2e_tests/spec_prism_disp_e2e.py
JuergenSchreiber: Essentially it works now like you propose. Crop centers the input 81x81 frame in the 125x125 frame. The solution was to replace the padded NaNs with 0 in fit_psf_centroid before ndi.median_filtering since this function cannot deal with NaNs.

corgidrp/spec.py
```diff
@@ -228,6 +228,8 @@ def fit_psf_centroid(psf_data, psf_template,
         xcom_template, ycom_template = (np.rint(xcent_template), np.rint(ycent_template))
 
     if xcent_guess is None or ycent_guess is None:
+        #filter NaNs
+        psf_data[np.isnan(psf_data)] = 0
```
semaphoreP: we should make a copy of the data before setting NaNs to 0's

semaphoreP (commented):
One small comment, and I'm having an issue that the L1 to ND Filter Spec e2e test is now failing. See error message below:

```
ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:763 Error: Improper input: func input vector length N=5 must not exceed func output vector length M=0
ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:764 Error type: TypeError
ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:765 Full traceback:
ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:766 Traceback (most recent call last):
  File "/home/jwang/pylibs/corgidrp/tests/e2e_tests/l1_to_nd_filter_spec_e2e.py", line 757, in test_nd_filter_spec_e2e
    run_nd_filter_spec_e2e(l1_datadir, processed_cal_path, outputdir, logger)
  File "/home/jwang/pylibs/corgidrp/tests/e2e_tests/l1_to_nd_filter_spec_e2e.py", line 501, in run_nd_filter_spec_e2e
    walker.walk_corgidrp(l2b_filelist, "", outputdir)
  File "/home/jwang/pylibs/corgidrp/corgidrp/walker.py", line 259, in walk_corgidrp
    output_filelist = run_recipe(recipe)
                      ^^^^^^^^^^^^^^^^^^
  File "/home/jwang/pylibs/corgidrp/corgidrp/walker.py", line 898, in run_recipe
    curr_dataset = step_func(curr_dataset, *other_args, **kwargs)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jwang/pylibs/corgidrp/corgidrp/l3_to_l4.py", line 1158, in determine_wave_zeropoint
    spot_centroids = compute_psf_centroid(dataset = offset_dataset, template_dataset = template_dataset, initial_cent = initial_cent)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jwang/pylibs/corgidrp/corgidrp/spec.py", line 431, in compute_psf_centroid
    xfit, yfit, gauss2d_xfit, gauss2d_yfit, psf_peakpix_snr, x_precis, y_precis = fit_psf_centroid(
                                                                                  ^^^^^^^^^^^^^^^^^
  File "/home/jwang/pylibs/corgidrp/corgidrp/spec.py", line 265, in fit_psf_centroid
    fitted_data_stamp, model, residual) = gaussfit2d_pix(psf_data,
                                          ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jwang/pylibs/corgidrp/corgidrp/spec.py", line 94, in gaussfit2d_pix
    p, success = optimize.leastsq(errorfunction, guess)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jwang/miniconda3/envs/corgidrp/lib/python3.12/site-packages/scipy/optimize/_minpack_py.py", line 430, in leastsq
    raise TypeError(f"Improper input: func input vector length N={n} must"
TypeError: Improper input: func input vector length N=5 must not exceed func output vector length M=0
```

JuergenSchreiber:
> One small comment, and I'm having an issue that the L1 to ND Filter Spec e2e test is now failing. See error message below:
> 
> ```
> ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:763 Error: Improper input: func input vector length N=5 must not exceed func output vector length M=0
> ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:764 Error type: TypeError
> ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:765 Full traceback:
> ERROR    l1_to_nd_filter_spec_e2e:l1_to_nd_filter_spec_e2e.py:766 Traceback (most recent call last):
>   File "/home/jwang/pylibs/corgidrp/tests/e2e_tests/l1_to_nd_filter_spec_e2e.py", line 757, in test_nd_filter_spec_e2e
>     run_nd_filter_spec_e2e(l1_datadir, processed_cal_path, outputdir, logger)
>   File "/home/jwang/pylibs/corgidrp/tests/e2e_tests/l1_to_nd_filter_spec_e2e.py", line 501, in run_nd_filter_spec_e2e
>     walker.walk_corgidrp(l2b_filelist, "", outputdir)
>   File "/home/jwang/pylibs/corgidrp/corgidrp/walker.py", line 259, in walk_corgidrp
>     output_filelist = run_recipe(recipe)
>                       ^^^^^^^^^^^^^^^^^^
>   File "/home/jwang/pylibs/corgidrp/corgidrp/walker.py", line 898, in run_recipe
>     curr_dataset = step_func(curr_dataset, *other_args, **kwargs)
>                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/jwang/pylibs/corgidrp/corgidrp/l3_to_l4.py", line 1158, in determine_wave_zeropoint
>     spot_centroids = compute_psf_centroid(dataset = offset_dataset, template_dataset = template_dataset, initial_cent = initial_cent)
>                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/jwang/pylibs/corgidrp/corgidrp/spec.py", line 431, in compute_psf_centroid
>     xfit, yfit, gauss2d_xfit, gauss2d_yfit, psf_peakpix_snr, x_precis, y_precis = fit_psf_centroid(
>                                                                                   ^^^^^^^^^^^^^^^^^
>   File "/home/jwang/pylibs/corgidrp/corgidrp/spec.py", line 265, in fit_psf_centroid
>     fitted_data_stamp, model, residual) = gaussfit2d_pix(psf_data,
>                                           ^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/jwang/pylibs/corgidrp/corgidrp/spec.py", line 94, in gaussfit2d_pix
>     p, success = optimize.leastsq(errorfunction, guess)
>                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/jwang/miniconda3/envs/corgidrp/lib/python3.12/site-packages/scipy/optimize/_minpack_py.py", line 430, in leastsq
>     raise TypeError(f"Improper input: func input vector length N={n} must"
> TypeError: Improper input: func input vector length N=5 must not exceed func output vector length M=0
> ```

That seems to be a hard nut, as we say in Germany. There is one frame where the guess position is outside the array

semaphoreP:
Is this because the frame is cropped too small? Could also ask Aneesh since he wrote the test if you are looking for help

JuergenSchreiber:
> Is this because the frame is cropped too small? Could also ask Aneesh since he wrote the test if you are looking for help

No, the frame is big enough, it happens even in the smallest 3D spectrum. It is perfectly centered in the 125x125 frame but the registration_result of the optimize.minimize in psf_registration_costfunc yields a value outside the frame in y direction. This I could avoid by using the 'Nelder-Mead' instead of 'Powell' method in the minimize function. Don't ask me why. But unit tests fail with Nelder-Mead....

semaphoreP:
Is it an initial with the issue guess not being good?

JuergenSchreiber:
Hi Jason, I cannot find something wrong with the initial guess, the reason for failure is at the 3D frame the y registration_result.x[1] yields a very high value of 170 which moves the y value outside the 125 frame. I skipped now the crop value from l2b_to_nd_filter_spec.json since the e2e seems to work here with the big frame. If we want the crop here maybe Neil has to look into this error, it might be necessary to review the complex fitting procedure of the spectra, since somehow there are "shaky" cases.

semaphoreP:
tagging @neilzim and @aneeshb97 to see if they have any thoughts about cropping for the spec nd filter calibration or not. Some of these spec mode observations are taken without the coronagraph and without the slit, so the source could be in many places along the detector. Do we still want to crop?

neilzim:
The calibration PSF images such as the ND filter calibration would be expected to be positioned on the same EXCAM region as the science image. So I would think cropping down from 1024x1024 would be fine in general.
However, there's an underlying concern of behavior we don't understand from the least-squares registration subroutine. I would like to reproduce the error @JuergenSchreiber described to investigate.

semaphoreP:
How do we know what is the center of the region to crop down to for unocculted observations again? In occulted mode, we can use the target location from the wavefront sensing tip/tilt loop in the header. Do we have something similar in occulted mode too?

neilzim:
In the TVAC data, the "EACQ" row and column header values were set for the calibration data we collected with the tip/tilt loop open.

JuergenSchreiber:
> In the TVAC data, the "EACQ" row and column header values were set for the calibration data we collected with the tip/tilt loop open.

Crop takes EACQ as center if centerxy input parameter is not set.

JuergenSchreiber:
> The calibration PSF images such as the ND filter calibration would be expected to be positioned on the same EXCAM region as the science image. So I would think cropping down from 1024x1024 would be fine in general. However, there's an underlying concern of behavior we don't understand from the least-squares registration subroutine. I would like to reproduce the error @JuergenSchreiber described to investigate.

Just run l1_to_nd_filter_spec_e2e by adding crop to l2b_to_nd_filter_spec.json after divide_by_exptime

semaphoreP:
Hi @neilzim just want to confirm you're planning on investigating the least squares fitting routine yourself, based on your earlier message.

semaphoreP:
hey @JuergenSchreiber, in PR #953, we upgraded the ND filter calibration (which also needs to find a star on the detector) to be more robust by using the `astrom.centroid_with_roi()` function. In there, Eric added a `gaussian_kernel_size` keyword that smooths the image by a gaussian to get a good first guess on the centering. I'm wonder if we can either use that function or adopt the same philosophy here, to help the fitting.

JuergenSchreiber:
> hey @JuergenSchreiber, in PR #953, we upgraded the ND filter calibration (which also needs to find a star on the detector) to be more robust by using the `astrom.centroid_with_roi()` function. In there, Eric added a `gaussian_kernel_size` keyword that smooths the image by a gaussian to get a good first guess on the centering. I'm wonder if we can either use that function or adopt the same philosophy here, to help the fitting.

I am not sure whether smoothing of the spectra would help, since as far as I understand we do a kind of cross correlation between the template spectra and the data spectra. There is a ndi.shift function used in psf_registration_costfunc for this, which uses a default constant mode parameter, when the shift goes outside the frame. There are a lot of other modes, but I am not an expert in this function. Would be scipy.signal.correlate2d maybe a better option? 
There are also other functions in spec.py like get_shift_correlation, star_spec_registration that seem to deal with similar things?

neilzim:
In https://github.com/roman-corgi/corgidrp/tree/template_xcorr I have modified the fit_psf_centroid() function to correct the root issue behind the unreliable PSF centroids discussed above. I started a draft PR #984.

JuergenSchreiber:
the l1_to_nd_filter_spec_e2e.py e2e tests works now with cropped l2b frames by adjusting the expected excam position to center of 125x125 frame

semaphoreP (approved):
Looks good. thanks for diving into the details here.

Issue #917, opened 2026-05-28
Chaining recipes doesn't write all the recipes to the file headers.


maxwellmb:
**Describe the bug**
When a dataset is passed through that requires chained recipes only the last recipes is written into the header, but all of them should be. 

**To Reproduce**
Take a look at the output of e2e tests where chained recipes were used. 

**Expected behavior**
All the recipes should be written. To keep backwards compatibility, I suggest we keep the first recipes as using the. RECIPE header, but, then go to RECIPE2, RECIPE3 for subsequent headers. We probably need some logic to asses how many RECIPE header keywords already exist, and/or we should keep an NRECIPES keyword. 

**corgidrp version:**
current version

maxwellmb:
@zjwhale, @kjl0025 pointed out that some of the current RAM-heavy implementation for the chaining of recipes only writes the full input list one time. We could try to generalize that I think to only write all of the very first input files to the header.

maxwellmb:
closed with merge of PR #939

PR #932, opened 2026-06-09
write all recipes in to headers

zjwhale:
## Describe your changes

This PR updates recipe history bookkeeping so chained recipes are all preserved in FITS headers.

- Keeps the first recipe in the existing `RECIPE` keyword for backward compatibility.
- Writes subsequent chained recipes as `RECIPE2`, `RECIPE3`, etc.
- Adds `NRECIPES` to record how many recipes are present.
- Updates walker recipe propagation so intermediate pipeline outputs do not lose recipe history.
- Preserves full recipe metadata when building synthesized darks.
- Updates header merge/data-format checks to treat numbered recipe keywords as generated metadata.
- Adds tests for chained recipe headers and synthesized-dark recipe propagation.

## Type of change

- New feature (non-breaking change which adds functionality)

## Checklist before requesting a review
- [x] I have verified that all unit tests pass in a clean environment and added new unit tests, as appropriate
- [x] I have checked that I am merging into the right branch
- [ ] I have checked the output of the latest Github Actions run associated with this PR and confirmed running `pytest` did not produce any warnings
- [ ] I have checked if my code modifies an existing step function or modifies the data format docs. If it does, I've added my PR to the [list maintained here](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/212028061/Log+of+PRs+that+could+affect+existing+VAP+Tests).

PR #939, opened 2026-06-11
Store chained recipes

semaphoreP:
## Describe your changes
New attempt to address #917, focused on just writing chained recipe to header. deals with edge case of RAM heavy. 

## Type of change

#917 


corgidrp/walker.py
```diff
@@ -173,6 +173,45 @@ def _validate_template_structure(user_template, default_template_path, recipe_fi
...
+        prev_recipes (list): recipes that ran before this one in the chain.
+            Empty list or None means standalone. Defaults to None.
+        is_last_frame (bool): True stores full input lists; False uses compact
+            pointers for RAM-heavy recipes. Defaults to True.
+    """
+    # Clear stale RECIPE2+/NRECIPES headers left by a previous chain
+    if 'NRECIPES' in frame.ext_hdr:
+        for idx in range(2, frame.ext_hdr['NRECIPES'] + 1):
+            del frame.ext_hdr['RECIPE{0}'.format(idx)]
+        del frame.ext_hdr['NRECIPES']
```
maxwellmb, line 242: Why do we want to clear out these old recipes? Don't we want to always keep track of the history? 

I think a potential failure mode here is if the drp is being called on files that have already been processed by a chained recipes, and then gets called again for a subsequent recipe. For example the drp is called for a chained [l1 to l2a, l2a to l2b]. Later, it is then called to combine several l2b files into a calibration.

maxwellmb (commented):
A bit of a fundamental question about what we're trying to accomplish here. See my comment in the code.

corgidrp/walker.py
semaphoreP: Saving old recipes from previous processing is not what we currently do. (PR #932 is addressing that). Currently we override the RECIPE keyword, essentially clearing out old recipes before writing a new one. 

I'm not sure I understand why this is a failure mode?

maxwellmb: Indeed it's not what we currently do, but I think it's ultimately what we want to do, would you not agree? I don't think we want to lose track of what was previously done. 

Wouldn't it just be simpler at this stage to append a new recipe if there already is one? Two birds with one stone?

semaphoreP: I didn't want to go test everything to make sure it worked. I could, but it would take a bit (and revisions to data formats).

maxwellmb: which data formats? can't we just append a new recipe anytime we get one? Similar to what you have now. We could just check to see if there are N recipes and then append a new one, no? Perhaps I'm not thinking of something..

semaphoreP: we'd need to update the data format doc pages

semaphoreP: ok, I've added this.

semaphoreP:
@ecady-jpl @ajeldorado @agreenbaum: I wanted to check in, we're adding in additional header keywords about all the recipes used to process the data from L1, not just the most recent one used (and prev overwritten). New **optional** keywords that exist for some outputs are NRECIPES (when more than 1 recipe in the processing history) and RECIPEX (where X is a value for X to NRECIPES). Let me know if this is a problem for anything.

ecady-jpl:
As long as you're only adding new keywords and not modifying existing ones, none of the behaviors of existing tools should change.  But what happens if you have a file made with an earlier version of DRP that doesn't have these keywords present?  Does anything break?

Also consider using RECIPExx for the second one, and zero-pad the number.  Going over 9 recipes would not necessarily surprise me; going over 99 would, and making this change now would mildly simplify the sorting/parsing later.

semaphoreP:
It does not break anything to feed earlier files into the DRP with the current implementation. We currently don't have any processing that uses more than 4 recipes, so I'm not really worried about going over 9.

ecady-jpl:
No further objections from me, then.

maxwellmb (changes requested):
As discussed over slack, it would be good to have a unit test for the funky logic required when we have a ram_heavy situation.

maxwellmb (approved):

Issue #945, opened 2026-06-24
calibrate_fluxcal_aper and combine_subexposures fail silently if you pass in a dataset with multiple targets


juliamilton:
`calibrate_fluxcal_aper` uses `combine_subexposures`, which takes the median combine of the input dataset to identify which calspec file should be used to do flux calibration calculations.

If a dataset consisting of multiple different targets is passed in to either `calibrate_fluxcal_aper` or `combine_subexposures`, it will combine PSFs of different sources and produce an answer without erroring but which is not scientifically meaningful. 

Guarding against this would be nice. Fix could be adding in an error in `calibrate_fluxcal_aper` if a dataset is passed in with different targets, and/ or a warning in `combine_subexposures ` if that is done.

juliamilton:
This was identified when testing CAR 123, which takes data from 3 separate dim star sources.

JuergenSchreiber:
Hi Julia, in which e2e pipeline did this happen? shouldn't the dataset be split before going to calibrate_fluxcal_aper. Do the different targets have different VISITIDs?

semaphoreP:
Fixed in PR #950

PR #950, opened 2026-06-29
check on multiple targets before combining datasets

JuergenSchreiber:
## Describe your changes
added function check_uniq_keyword() to check whether a header keyword is unique in a dataset. Raise error if dataset in fluxcal.calibrate_flux_xxx has no unique targets.

## Type of change

#945 

### DIFF
--- a/corgidrp/fluxcal.py
+++ b/corgidrp/fluxcal.py
@@ -481,9 +481,12 @@ def calibrate_fluxcal_aper(dataset_or_image, calspec_file = None, flux_or_irr = 
     """
     d_or_i = dataset_or_image.copy()
     if isinstance(d_or_i, corgidrp.data.Dataset):
-        #take the median of images in the dataset
-        image = combine_subexposures(d_or_i, collapse = "median", num_frames_scaling=False)[0]
-        dataset = d_or_i
+        if corgidrp.check.check_uniq_keyword(d_or_i, "TARGET"):
+            #take the mean of images in the dataset
+            image = combine_subexposures(d_or_i, collapse = "median", num_frames_scaling=False)[0]
+            dataset = d_or_i
+        else:
+            raise AttributeError("dataset of different targets cannot be medianed")
     else:
         image = d_or_i
         dataset = corgidrp.data.Dataset([image])
@@ -613,9 +616,12 @@ def calibrate_pol_fluxcal_aper(dataset_or_image,
     """
     d_or_i = dataset_or_image.copy()
     if isinstance(d_or_i, corgidrp.data.Dataset):
-        #take the mean of images in the dataset
-        image = combine_subexposures(d_or_i, collapse = "median", num_frames_scaling=False)[0]
-        dataset = d_or_i
+        if corgidrp.check.check_uniq_keyword(d_or_i, "TARGET"):
+            #take the mean of images in the dataset
+            image = combine_subexposures(d_or_i, collapse = "median", num_frames_scaling=False)[0]
+            dataset = d_or_i
+        else:
+            raise AttributeError("dataset of different targets cannot be medianed")
     else:
         image = d_or_i
         dataset = corgidrp.data.Dataset([image])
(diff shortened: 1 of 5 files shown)


JuergenSchreiber:
I check on prihdr keyword "TARGET", but I guess "VISITID" would be better because more general, correct?

semaphoreP:
@JuergenSchreiber if we take calibration data of the same star multiple times, does it make sense to combine the datasets? My guess is probably not?

JuergenSchreiber:
@juliamilton: test_nd_filter_calibration is failing now since there are 2 different dim stars in the dataset for calibrate_fluxcal_xxx. This function was written to determine the fluxcal factor for one calibration star in the input data. If we have a dataset at all as input it should be several subsequent observations frames of one calibration star with the same pointing to improve SNR.  
I reduced the DIM_STARS input now to 1 star, then it works. Would this be ok, or would this mess up the logic of the nd filter calibration?

corgidrp/fluxcal.py
```diff
@@ -481,9 +481,12 @@ def calibrate_fluxcal_aper(dataset_or_image, calspec_file = None, flux_or_irr =
...
     if isinstance(d_or_i, corgidrp.data.Dataset):
-        #take the median of images in the dataset
-        image = combine_subexposures(d_or_i, collapse = "median", num_frames_scaling=False)[0]
-        dataset = d_or_i
+        if corgidrp.check.check_uniq_keyword(d_or_i, "TARGET"):
+            #take the mean of images in the dataset
+            image = combine_subexposures(d_or_i, collapse = "median", num_frames_scaling=False)[0]
+            dataset = d_or_i
+        else:
+            raise AttributeError("dataset of different targets cannot be medianed")
```
semaphoreP: maybe useful to print out what are the TARGET values you got? useful for error message debugging.

semaphoreP (commented):
other than the question to Julia and the visitid/target decision, which I will respond separately, the code looks fine to me.

semaphoreP:
> @juliamilton: test_nd_filter_calibration is failing now since there are 2 different dim stars in the dataset for calibrate_fluxcal_xxx. This function was written to determine the fluxcal factor for one calibration star in the input data. If we have a dataset at all as input it should be several subsequent observations frames of one calibration star with the same pointing to improve SNR. I reduced the DIM_STARS input now to 1 star, then it works. Would this be ok, or would this mess up the logic of the nd filter calibration?

I think most of the tests can handle multiple DIM_STARS except the `test_nd_filter_calibration_with_fluxcal` test. Would it be better to upgrade that test to only pick one dim star to test on? It seems like the ND filter code itself (`create_nd_filter_cal`) actually can handle multiple dim stars. But I'd like @juliamilton to confirm on this.

juliamilton:
Yes, the ND filter code is set up to be able to handle multiple dim stars, and the `test_nd_filter_calibration_with_fluxcal` test is the only unit test that should error now if the dim star dataset contains multiple different dim targets. (It actually doesn't error for me when I test it on this branch, because the mock files are produced so quickly that they overwrite each other, leaving only one dim star image in the dataset. Which is a separate problem.)

I would suggest we do what Jason said and upgrade that unit test to only use a dataset consisting of a single TARGET. I think we should also fix the issue where the mock files are colliding and overwriting each other.

juliamilton:
Also to answer your question about using TARGET or VISITID - I agree with what Jason said. Either is probably fine, but using VISITID would be a stronger check.

JuergenSchreiber:
> Also to answer your question about using TARGET or VISITID - I agree with what Jason said. Either is probably fine, but using VISITID would be a stronger check. 

I use VISITID now and only one image in the dataset. @juliamilton I do not quite understand your comment with overwriting mock files, maybe you can raise an issue or solve this in this PR if you think this is a problem.

JuergenSchreiber:
up to now the VISITID in mocks.create_flux_image was always the same, now I randomly change the VISITID in the test when generating the dim/bright images

semaphoreP (approved):

PR #953, opened 2026-06-30
Created L1 to NDFilterSweetSpot e2e test

everlastingEric:
## Describe your changes
- Created a new E2E test to test running L1 corgisim data through the DRP to obtain an NDFilterSweetSpot calibration
- Check that the output calibration file is formatted correctly and the calculated OD is correct
- The dataset consists of a set of unocculted simulations of dim and bright targets. Bright targets were simulated with the ND475 filter
- Uploaded simulated dataset to the E2E testing dropbox folder

#707 

### DIFF
--- a/tests/e2e_tests/l1_to_nd_filter_e2e.py
+++ b/tests/e2e_tests/l1_to_nd_filter_e2e.py
@@ -39,7 +39,7 @@ def test_l1_to_nd_filter_e2e(e2edata_path, e2eoutput_path):
     
     # suppress warnings in the pipeline
     with warnings.catch_warnings():
-        warnings.simplefilter("ignore", category=UserWarning) # warning about OD variation
+        warnings.simplefilter("ignore", category=UserWarning) # warning about detector noisemap calibration
         warnings.simplefilter("ignore", category=RuntimeWarning) # warning about different EM gain between bright and dark frames
         walker.walk_corgidrp(l1_input_data_list, "", l2b_outputdir)
 
--- a/tests/e2e_tests/l1_to_nd_filter_e2e.py
+++ b/tests/e2e_tests/l1_to_nd_filter_e2e.py
@@ -14,7 +14,7 @@ thisfile_dir = os.path.dirname(__file__)
 @pytest.mark.e2e
 def test_l1_to_nd_filter_e2e(e2edata_path, e2eoutput_path):
     # grab input L1 data, consisting of dim (no ND) and bright (ND 475) stars
-    l1_input_data_dir = os.path.join(e2edata_path, "ND_sims")
+    l1_input_data_dir = os.path.join(e2edata_path, "ND_sims", "L1")
     l1_input_data_list = glob.glob(os.path.join(l1_input_data_dir, "*_l1_*.fits"))
 
     # Initialize a connection to the calibration database
@@ -23,8 +23,11 @@ def test_l1_to_nd_filter_e2e(e2edata_path, e2eoutput_path):
     # remove any existing caldb file so that CalDB() creates a new one
     if os.path.exists(corgidrp.caldb_filepath):
         os.remove(tmp_caldb_csv)
+
+    # grab calibration files for processing L1 data
+    cal_dir =os.path.join(e2edata_path, "ND_sims", "Cals") 
     db = caldb.CalDB()
-    db.scan_dir_for_new_entries(corgidrp.default_cal_dir)
+    db.scan_dir_for_new_entries(cal_dir)
 
     # create empty output directory
     test_outputdir = os.path.join(e2eoutput_path, "l1_to_ND_filter_e2e")
@@ -32,17 +35,12 @@ def test_l1_to_nd_filter_e2e(e2edata_path, e2eoutput_path):
         shutil.rmtree(test_outputdir)
     os.makedirs(test_outputdir)
     l2b_outputdir = os.path.join(test_outputdir, "l2b_results")
-    if not os.path.exists(l2b_outputdir):
-        os.mkdir(l2b_outputdir)
-
-    # clean up by removing old files
-    for file in os.listdir(l2b_outputdir):
-        os.remove(os.path.join(l2b_outputdir, file))
+    os.makedirs(l2b_outputdir)
     
     # suppress warnings in the pipeline
     with warnings.catch_warnings():
-        warnings.simplefilter("ignore", category=UserWarning)
-        warnings.simplefilter("ignore", category=RuntimeWarning)
+        warnings.simplefilter("ignore", category=UserWarning) # warning about OD variation
+        warnings.simplefilter("ignore", category=RuntimeWarning) # warning about different EM gain between bright and dark frames
         walker.walk_corgidrp(l1_input_data_list, "", l2b_outputdir)
 
 
@@ -52,8 +50,9 @@ def test_l1_to_nd_filter_e2e(e2edata_path, e2eoutput_path):
 
     recovered_od = float(nd_cal.od_values[0])  # use the first entry for the check
     assert recovered_od == pytest.approx(4.75, abs=1e-1)
-
     check.compare_to_mocks_hdrs(nd_file[0])
+    assert nd_cal.ext_hdr["DATATYPE"] == "NDFilterSweetSpotDataset"
+    assert nd_cal.ext_hdr["DATALVL"] == "CAL"
     
     # remove temporary caldb file
     os.remove(tmp_caldb_csv)
--- a/tests/e2e_tests/l1_to_nd_filter_e2e.py
+++ b/tests/e2e_tests/l1_to_nd_filter_e2e.py
@@ -6,9 +6,7 @@ import corgidrp
 import corgidrp.mocks as mocks
 import corgidrp.data as data
 import corgidrp.walker as walker
-import corgidrp.nd_filter_calibration as nd_filter_calibration
 from corgidrp import caldb, check
-from astropy.time import Time
 
 # this file's folder
 thisfile_dir = os.path.dirname(__file__)
@@ -16,7 +14,7 @@ thisfile_dir = os.path.dirname(__file__)
 @pytest.mark.e2e
 def test_l1_to_nd_filter_e2e(e2edata_path, e2eoutput_path):
     # grab input L1 data, consisting of dim (no ND) and bright (ND 475) stars
-    l1_input_data_dir = os.path.join(e2edata_path, "ND_sims", "L1")
+    l1_input_data_dir = os.path.join(e2edata_path, "ND_sims")
     l1_input_data_list = glob.glob(os.path.join(l1_input_data_dir, "*_l1_*.fits"))
 
     # Initialize a connection to the calibration database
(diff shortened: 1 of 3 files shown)


tests/e2e_tests/l1_to_nd_filter_e2e.py
```diff
@@ -0,0 +1,75 @@
...
+    if os.path.exists(test_outputdir):
+        shutil.rmtree(test_outputdir)
+    os.makedirs(test_outputdir)
+    l2b_outputdir = os.path.join(test_outputdir, "l2b_results")
+    if not os.path.exists(l2b_outputdir):
+        os.mkdir(l2b_outputdir)
+
+    # clean up by removing old files
+    for file in os.listdir(l2b_outputdir):
+        os.remove(os.path.join(l2b_outputdir, file))
```
semaphoreP: after you delete and remake the l1_to_ND_filter_e2e folder, can you assume for simplicity that there is no l2b_outputdir subfolder, and that the there are no files in there?

tests/e2e_tests/l1_to_nd_filter_e2e.py
```diff
@@ -0,0 +1,75 @@
...
+    
+    # suppress warnings in the pipeline
```
semaphoreP, line 40: could you document which specific errors you are aiming to suppress?

tests/e2e_tests/l1_to_nd_filter_e2e.py
```diff
@@ -0,0 +1,75 @@
...
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        walker.walk_corgidrp(l1_input_data_list, "", l2b_outputdir)
+
+
+    # 5. Load product & assert if calculated OD matches what is expected
+    nd_file = glob.glob(os.path.join(l2b_outputdir, "*_ndf_cal*.fits"))
+    nd_cal  = data.NDFilterSweetSpotDataset(nd_file[0])
+
+    recovered_od = float(nd_cal.od_values[0])  # use the first entry for the check
+    assert recovered_od == pytest.approx(4.75, abs=1e-1)
```
semaphoreP, line 52: I just ran the e2e test myself and this assert fails with `assert 5.373562812805176 == 4.75 ± 0.1`. It works for you?

semaphoreP (commented):
Hmm the e2e test is not passing for me.

tests/e2e_tests/l1_to_nd_filter_e2e.py
everlastingEric: hmm yeah the test works for me, let me double check that the data I uploaded is correct. The only other reason I could think of is that the default TVAC calibrations that caldb generates is somehow different and that messes up the numbers but I'm not sure if that's possible. I'll also double check that the caldb is always initialized from a fresh state.

semaphoreP: oh I might have missed downloading something.. I'll double check on my end.

tests/e2e_tests/l1_to_nd_filter_e2e.py
everlastingEric: yeah good point this is probably redundant

tests/e2e_tests/l1_to_nd_filter_e2e.py
```diff
@@ -0,0 +1,75 @@
...
+
+    check.compare_to_mocks_hdrs(nd_file[0])
```
semaphoreP, line 53: Just realized another comment: in the Core Throughput e2e test you check against the data format docs. Is there a reason you don't here?

everlastingEric: oh I think that slipped my mind, I could add it back in for completeness

tests/e2e_tests/l1_to_nd_filter_e2e.py
everlastingEric: One of them says that OD variation of 0.185 exceeds the threshold of 0.1, and the other says that EMGAIN_A is not identical across frames, which just comes from the bright and dim frames having different gain. Are these fine to suppress?

tests/e2e_tests/l1_to_nd_filter_e2e.py
semaphoreP: Same problem still for me after redownloading the test data.

everlastingEric: I redownloaded the data too and it still works fine so it's not that, I'll try to force it to load in a specific set of calibration files and see if that changes anything

everlastingEric: ok I updated the code so the calibration files are getting pulled out of a custom directory now and also addressed the other comments, can you try again to see if the test passes now? I did also reorganize the test files so you'll have to redownload them.

tests/e2e_tests/l1_to_nd_filter_e2e.py
semaphoreP: The OD variation one might be pointing to some issue. that means the OD measured between files varies by more than 10%. Maybe it's worth taking a look at the individual file OD measurements (inside of process_bright_target()) to understand what's going on.

tests/e2e_tests/l1_to_nd_filter_e2e.py
semaphoreP: Just reran and got the same error `assert 5.377058982849121 == 4.75 ± 0.1`. Very small numerical differences between calibration files. 

I don't think it really matters, but have you tried running the test using the pytest command rather than the python file directly?

semaphoreP: I could also send you my l2b products produced by the pipeline to see if they look the same as yours.

everlastingEric: hmm interesting, I have been using pytest but I think I know what the issue is actually and fixing the high OD variation probably is the right move

everlastingEric: Or nevermind maybe not I need to think about this a bit more, but the l2b files as a comparison would be helpful

everlastingEric: if you can send me your final ND calibration file too that would also be helpful

semaphoreP: I put them on sharepoint. At the top level, if you go into the "Scratch" folder, go into /Scratch/jwang/nd_filter_debug

everlastingEric: Ok I figured out what is causing the issue. Our L2b files are exactly the same but the internal ordering of the files in the pipeline for the final NDFilter step differs, which means the ordering of our final list of ND values also differ. The ND values are the exact same when sorted but since by default it's not, the line of code `nd_file[0]` gives different results and it just happens that your value is outside the range set by the test. The average of all ND values is still 4.76 which is very close to the 4.75 ideal. Easiest solution is to modify the test so it looks at the average instead of the first value. However if we want to reduce the std between the OD values then the input dataset might need some work.

everlastingEric: For the input data right now I am using a trimmed version of the CAR 123 data, where for the bright target instead of 21 frames per FSM dither I am only using 3. This reduced number of datapoints could be what is causing the wide OD variation but let me know what you think.

semaphoreP: Ok I think the difference between 5.3 and 4.76 is more than a difference of a factor of 3x in linear space. This seems like a lot and may be a symptom of us not doing something right.

Are you saying that we are producing more than one ND filer calibration file? If so, can you remind me why?

everlastingEric: The test produces one file only, I'm not super familiar with the file formatting of the ND filter calibration but I think in that one file it contains an OD value for each of the bright frames and that is where the variation and different ordering comes in

everlastingEric: my apologies I wrote the wrong line, I was referring to this: `recovered_od = float(nd_cal.od_values[0])`

semaphoreP: the ND filter sweetspot dataset measures the change in the ND filter transmission at different x/y positions across the detector. Does the simulations simulate the change in ND filter transmission across the filed of view or should they really all be the same OD. If the later, maybe something we should debug (probably somehow the flux is not being measured properly in some frames).

everlastingEric: I'm trying to find how corgisim or cgisim treats ND filters, but regardless I think the flux not being measured properly is a likely culprit given that the difference is a factor of 3 on a linear scale, so I'll try to dig into that

everlastingEric: So I looked at the distribution of OD values and identified two outliers which produced >5 OD values. What I found is that in the corresponding L2b frames there is like a bright outlier pixel at the edge of the image (bad pixel or cosmic ray that didn't get masked with NaN possibly) which the centroid algorithm grabs on to instead of the actual PSF and because of that the entire flux calculation for those frames is just wrong.

semaphoreP: Is it in the original L1 data? and some bad pixel created by the sims?

everlastingEric: Yeah it's in the L1 data. It looks to be a bad pixel of some sorts although this one really stands out because it's almost as bright as the cosmic rays but it's just a column of 3 pixels with no tail. When the image got cropped this pixel happened to be right at the image edge.

semaphoreP: do we need to, say, run a 3x3 median filter on the image, guess the peak pixel from the peak of the median filtered image, and start the centroiding with that peak pixel as the initial guess?

everlastingEric: That's a good idea, I'll put that in and see if it helps

everlastingEric: I put in the median filter, a size of 3x3 causes issues but 5x5 works. I think the issue with 3x3 is that the residues left behind by imperfect cosmic ray masking sometimes doesn't get averaged down properly because it's surrounded by a bunch of NaN pixels and that biases the median in a weird way.  In many cases the stellar PSF is averaged down comparatively more and that causes the cosmic ray residue to become the brightest pixel in the frame. But it seems a 5x5 filter is able to capture just enough of the background to average down the cosmic ray residues.

semaphoreP: Sounds good. maybe it's an adjustable size that defaults to 5 then?

everlastingEric: ok quick question, I am currently putting the median filter in the `centroid_with_roi` function in `astrom.py`. I can put the default directly in there, but that would affect everything else that calls this function. Or I can put the global default as none, but make the ND filter calibration steps override that to a default of 5 for `compute_od_per_file` specifically. Which one would work better?

semaphoreP: Yeah, default to None (which will skip it) by default. And then have ND filter default to 5, with an optional keyword in ND filter calibration to change the med filter size.

everlastingEric: ok upon closer inspection the median filter overall is still pretty buggy and giving wrong locations. What seemed to work better is to use a Gaussian kernel instead, and now every frame is locked onto the actual star. Now the OD have a mean of 4.71 and standard deviation of 0.014. If we want to use that instead I'll keep the same logic with having a default kernel size specific to the ND filter calibration.

semaphoreP: interesting finding. ok sounds good to me.

everlastingEric: I committed the change, just going to wait and make sure the unit tests are still fine

semaphoreP:
This is looking good to me. Is there anything left to do?

semaphoreP (approved):

everlastingEric:
I think we are good, there were a couple of unit tests failing earlier but I fixed those already

PR #959, opened 2026-07-09
added L1 to polcal e2e test and fixed bugs during testing

everlastingEric:
## Describe your changes
- Added a L1 to mueller matrix calibration e2e test starting from simulated L1 pol data (unocculted polarized and unpolarized stars with ND225 filter)
- uploaded relevant test files to the sharepoint
- Fixed bug in `generate_mueller_matrix_cal` where the number of rows in the measurement matrix is determined by the length of the input dataset but the number of rows that actually get filled with measurement data is determined by the number of unique targets in the observation sequence. This creates an issue where the measurement matrix isn't filled up correctly if there are more input data frames than there are unique targets (for example same target observed with different roll angles). The fix is to simply keep everything referenced to the length of the input dataset rather than the number of unique targets. 
- Made a change to the `aper_phot` function to account for undercounted flux if a large chunk of the aperture consists of bad pixels (most common example being cosmic rays). Currently, the aperture photometry sum does not include pixel values from pixels flagged by the DQ extension, but the flux is always computed with reference to the full aperture size even if a sizable chunk of the aperture didn't contribute to the count. This creates an issue for polarimetric calibration, where if a cosmic ray lands on one of the two Wollaston spots and gets masked out with NaNs, the flux from that spot gets undercounted. The pipeline interprets this as a severe polarization aberration and spits out unphysical values for the Mueller matrix. To account for this to the first order, I modified the pipeline to count the number of bad pixels in the aperture so that the flux is normalized by the ratio of the full aperture size to the effective aperture size with the area taken up by bad pixels removed. (might be better ways of doing this and might want to make this into an optional step since not every single operation requiring aperture photometry might benefit from this, but for now I have it hard coded) Also should I add this change to the list of PRs that could affect existing VAP tests?

## Type of change

#708 

### DIFF
--- a/corgidrp/fluxcal.py
+++ b/corgidrp/fluxcal.py
@@ -364,9 +364,24 @@ def aper_phot(image, encircled_radius, frac_enc_energy=1., method='subpixel', su
         method=method,
         subpixels=subpixels
     )
+
+    # compute renormalization factor to account for lost flux if significant amount of NaN shows up in the aperture (for example cosmic rays)
+    # sum all good pixels (non NaNs) over the aperture
+    good_pixels = aper.do_photometry(
+        (~image.dq.astype(bool)).astype(float),
+        method=method,
+        subpixels=subpixels
+    )[0][0]
+
+    # find fraction of the aperture without NaNs
+    if good_pixels > 0:
+        correction_factor = aper.area / good_pixels
+    else:
+        # edge case where the aperture is all NaN, propagate NaN instead of a count of 0
+        correction_factor = np.nan
     
-    flux = aperture_sums[0] / frac_enc_energy
-    flux_err = aperture_sums_errs[0] / frac_enc_energy
+    flux = (correction_factor * aperture_sums[0]) / frac_enc_energy
+    flux_err = (correction_factor * aperture_sums_errs[0]) / frac_enc_energy
     
     if background_sub:
         return flux, flux_err, back
--- a/corgidrp/fluxcal.py
+++ b/corgidrp/fluxcal.py
@@ -364,24 +364,9 @@ def aper_phot(image, encircled_radius, frac_enc_energy=1., method='subpixel', su
         method=method,
         subpixels=subpixels
     )
-
-    # compute renormalization factor to account for lost flux if significant amount of NaN shows up in the aperture (for example cosmic rays)
-    # sum all good pixels (non NaNs) over the aperture
-    good_pixels = aper.do_photometry(
-        (~image.dq.astype(bool)).astype(float),
-        method=method,
-        subpixels=subpixels
-    )[0][0]
-
-    # find fraction of the aperture without NaNs
-    if good_pixels > 0:
-        correction_factor = aper.area / good_pixels
-    else:
-        # edge case where the aperture is all NaN, propagate NaN instead of a count of 0
-        correction_factor = np.nan
     
-    flux = (correction_factor * aperture_sums[0]) / frac_enc_energy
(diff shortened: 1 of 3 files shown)


semaphoreP:
Yes, I think we should note that change in aper phot behavior just in case it affects VAP tests (given it doesn't affect our unit tests, we're probably safe, but better to be thorough). 

Just to check, do we get multiple exposures taken with the pol cal, so that the data could be combined and NaN's disappear from the combination of frames?

corgidrp/fluxcal.py
```diff
@@ -364,9 +364,24 @@ def aper_phot(image, encircled_radius, frac_enc_energy=1., method='subpixel', su
...
+        method=method,
+        subpixels=subpixels
+    )[0][0]
+
+    # find fraction of the aperture without NaNs
+    if good_pixels > 0:
+        correction_factor = aper.area / good_pixels
+    else:
+        # edge case where the aperture is all NaN, propagate NaN instead of a count of 0
+        correction_factor = np.nan
```
semaphoreP: why nan instead of 1?

corgidrp/fluxcal.py
```diff
@@ -364,9 +364,24 @@ def aper_phot(image, encircled_radius, frac_enc_energy=1., method='subpixel', su
...
+    # sum all good pixels (non NaNs) over the aperture
+    good_pixels = aper.do_photometry(
+        (~image.dq.astype(bool)).astype(float),
+        method=method,
+        subpixels=subpixels
+    )[0][0]
+
+    # find fraction of the aperture without NaNs
+    if good_pixels > 0:
+        correction_factor = aper.area / good_pixels
```
semaphoreP: we should note that this is a very rough approximation, since pixels don't contribute equally to the total flux.

semaphoreP (commented):
just a couple questions/comments on the modifications to aperture photometry, and my comment above about whether we could average out the nans with multipel frames. otherwise looks good.

corgidrp/fluxcal.py
everlastingEric: no particular reason other than to make the flux nan in this case, I can change it to 1 if that's the usual convention

everlastingEric:
regarding combining frames, I do have multiple frames for each observation to reduce the effects of cosmic rays but the frames don't get combined until after the Stokes vectors are extracted from it. I think it's the Stokes vectors that gets combined into one per observation so having cosmic rays will still bias it somewhat if not very many frames are used. Is there any way to combine the frames earlier?

everlastingEric:
Nevermind I misspoke the fluxes are averaged to produce one Stokes vector, so the averaging is already done before not after

everlastingEric:
ok another update, I ran the test again without the modifications to aperture photometry and it still passed without any issue. Given that I am already passing in multiple frames per observation for this test maybe I could just leave the function as is then and not modify anything.

semaphoreP:
My internet is really bad right now, so going to try to reply over email. I think given that the proper way to do corrections for missing pixels in aperture photometry requires assuming a PSF shape, I think if this test works without modifications to aperture photometry, let's keep that.
________________________________

semaphoreP:
Oops, I misunderstood the code. Nan is the right decision.
________________________________

everlastingEric:
ok I reverted the edits I made to aperture photometry code, it should be good to go

semaphoreP:
Had to merge develop to get the `check_uniq_keyword` function. Running unit tests again to make sure they pass (Except the ones that are broken) before merge.

semaphoreP (approved):

PR #962, opened 2026-07-14
accommodate new version of emccd_detect

kjl0025:
## Describe your changes
I changed the specification of the calls to emccd_detect to be consistent with the calls to the previous version of emccd-detect so that the tests pass.  The only calls to emccd-detect in the DRP are made in 2 functions in mocks.py: generate_mock_pump_trap_data and create_photon_countable_frames.  I changed requirements.txt to specify emccd-detect >= 2.6.1.  The Linux issue with FPN_path vs fpn_path is also resolved in version 2.6.1 of emccd-detect (its latest version).

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Reference any relevant issues (don't forget the #)
N/A

## Checklist before requesting a review
- [x] I have verified that all unit tests pass in a clean environment and added new unit tests, as appropriate
- [x] I have checked that I am merging into the right branch
- [x] I have checked the output of the latest Github Actions run associated with this PR and confirmed running `pytest` did not produce any warnings
- [x] I have checked if my code modifies an existing step function or modifies the data format docs. If it does, I've added my PR to the [list maintained here](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/212028061/Log+of+PRs+that+could+affect+existing+VAP+Tests).

PR #963, opened 2026-07-15
Remove chained recipe for 2 vistypes

### LINKED ISSUE #960: Remove chained L1->L4 for CGIVST_CAL_SPEC_TGTREF and CGIVST_CAL_TGTREF_PHOT VISTYPEs
**Describe the bug**
When running CAR124 with @mygouf, we noticed that the L1 data with VISTYPE=CGIVST_CAL_SPEC_TGTREF was processed up to L4. 

When looking at walker.py, we also see that the same is done for L1 data with VISTYPE=CGIVST_CAL_TGTREF_PHOT. 

**To Reproduce**
Take L1 data with VISTYPE=CGIVST_CAL_SPEC_TGTREF and process it via corgidrp. 

**Expected behavior**
CGIVST_CAL_SPEC_TGTREF with ISPC=0 corresponds to science data (spectroscopy-non coronographic). It should not be considered calibration and thus not processed to L4 as it is the case right now (walker.py:576). 

Same for CGIVST_CAL_TGTREF_PHOT that corresponds to science data (Imaging WFOV - No Coronagraphic).  

Note that the same VISTYPE can also corresponds to a (TBD) Dispersion Model Calibration but this does not seem to be implemented yet.

**corgidrp version:**
 - 5.0

**Additional context**

adrienmaillard:
## Describe your changes

I have removed 2 chained recipes. These 2 vistypes are for science data (`Spectroscopy - No Coronagraphic` and `Imaging NFOV (TTR5) - No Coronagraphic`), not calibrations.

## Type of change

Fixes #960 

### DIFF
--- a/corgidrp/walker.py
+++ b/corgidrp/walker.py
@@ -658,7 +658,7 @@ def guess_template(dataset):
             recipe_filename = "l2b_to_polcal.json"
         elif image.ext_hdr['DPAMNAME'] == 'POL0' or image.ext_hdr['DPAMNAME'] == 'POL45':
             recipe_filename = "l2b_to_l3_pol.json"
-        elif 'TDD' not in image.pri_hdr['VISTYPE']:
+        elif 'TDD' not in image.pri_hdr['VISTYPE'] and image.pri_hdr['VISTYPE'] != "CGIVST_CAL_SPEC_TGTREF":
             warnings.warn("Only VISTYPE TDD and certain cal frames should be processed beyond L2b. Double-check which frames are being processed from L2b -> L3.")
             recipe_filename = "l2b_to_l3.json"
         else:
--- a/corgidrp/walker.py
+++ b/corgidrp/walker.py
@@ -573,12 +573,6 @@ def guess_template(dataset):
         elif image.pri_hdr['VISTYPE'] == 'CGIVST_CAL_SPEC_LINESPREAD':
             recipe_filename = ["l1_to_l2a_basic.json", "l2a_to_l2b_spec.json", 'l2b_to_spec_linespread.json']
             chained = True
-        elif image.pri_hdr['VISTYPE'] == 'CGIVST_CAL_SPEC_TGTREF':
-            recipe_filename = ["l1_to_l2a_basic.json","l2a_to_l2b_spec.json","l2b_to_l3.json","l3_to_l4_noncoron_spec.json"]
-            chained = True
-        elif image.pri_hdr['VISTYPE'] == 'CGIVST_CAL_TGTREF_PHOT' and image.ext_hdr['DPAMNAME'] not in ['POL0','POL45']:
-            recipe_filename = ["l1_to_l2a_basic.json","l2a_to_l2b.json","l2b_to_l3.json","l3_to_l4_nopsfsub.json"]
-            chained = True
         elif image.pri_hdr['VISTYPE'] == 'CGIVST_CAL_TPUMP':
             recipe_filename = ['trap_pump_cal_1.json', 'trap_pump_cal_2.json']
             chained = True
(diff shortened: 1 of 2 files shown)


semaphoreP:
Don't worry about the issues with the photon counting and trap tump unit tests causing them to fail. Those will be addressed in PR #962

semaphoreP (commented):
I think the only other thing that needs to be done is that we need to modify the l1_to_l4_spec_noncoron_e2e.py e2e test so that it calls the walker 4 times (l1 -> l2a, l2a -> l2b, l2b -> l3, l3 -> l4) rather than one time. You'll need to keep track of the inputs/outputs of each walker stage, but hopefully very straightforward file pattern matching.

adrienmaillard:
> I think the only other thing that needs to be done is that we need to modify the l1_to_l4_spec_noncoron_e2e.py e2e test so that it calls the walker 4 times (l1 -> l2a, l2a -> l2b, l2b -> l3, l3 -> l4) rather than one time. You'll need to keep track of the inputs/outputs of each walker stage, but hopefully very straightforward file pattern matching.

OK, I'll look into that. I am wondering why the test passed though.

adrienmaillard:
The test already had a manual path in which it executed l1 -> l4 manually. I just removed the chained reciped branch. And fixed the output path of the l4 fits that was wrong. It passes on my machine.

corgidrp/walker.py
semaphoreP: we should check and not throw a warning if VISTYPE is CGIVST_CAL_SPEC_TGTREF

tests/e2e_tests/l1_to_l4_spec_noncoron_e2e.py
```diff
@@ -289,124 +289,116 @@ def run_l1_to_l4_e2e_test(l1_datadir, l4_outputdir, processed_cal_path, logger):
...
+    l2a_files = [f for f in os.listdir(l4_outputdir) if f.endswith('_l2a.fits')]
+    l2a_filelist = [os.path.join(l4_outputdir, f) for f in l2a_files]
+    logger.info(f'L1 to L2a complete. Generated {len(l2a_filelist)} L2a files.')
     
-        # Find the L2a output files
-        l2a_files = [f for f in os.listdir(l4_outputdir) if f.endswith('_l2a.fits')]
-        l2a_filelist = [os.path.join(l4_outputdir, f) for f in l2a_files]
-        logger.info(f'L1 to L2a complete. Generated {len(l2a_filelist)} L2a files.')
+    '''
+    if l2a_filelist:
```
semaphoreP, line 304: Is the reason why you are commenting out blocks of this code and not deleting it is because it could be useful in the future? I.e., why comment out instead of deleting?

semaphoreP: Actually, I think noncoron data will never be photon counting, so we probably can delete it.

semaphoreP (commented):
Thanks for doing that. I noticed one other place in the walker logic we should modify to avoid a warning, as well as question about whether we should just delete some unused code to keep things clean .

tests/e2e_tests/l1_to_l4_spec_noncoron_e2e.py
adrienmaillard: It was already commented by the original author. I just removed an if/else block.

corgidrp/walker.py
adrienmaillard: so you mean adding a `VISTYPE!=CGIVST_CAL_SPEC_TGTREF` to this condition on line 661 ?

semaphoreP: Yes, we don't want this warning being through trying to process CGIVST_CAL_SPEC_TGTREF through L3.

tests/e2e_tests/l1_to_l4_spec_noncoron_e2e.py
adrienmaillard: I'll delete it. We can always find it later if needed.

semaphoreP: ok no worries then.

adrienmaillard: I did not delete the commented code then.

corgidrp/walker.py
adrienmaillard: Done.

semaphoreP (approved):

Issue #954, opened 2026-07-01
Astrometric E2E test with corgisim data


semaphoreP:
**Is your feature request related to a problem? Please describe.**
Update the astrometry E2E test to use simulated corgisim data

**Describe the solution you'd like**
Use L1 data simulated for the CAR tests and ensure we still meet the performance goals.

semaphoreP:
Added in PR #973

PR #973, opened 2026-07-20
Add l1 to astrom e2e test

everlastingEric:
## Describe your changes
- Added e2e test that takes in simulated L1 observations of stellar fields and performs astrometric calibration
- Uploaded relevant test files to the e2e test data sharepoint
- Updated `zz_dataformats_e2e.py` to use the astrometric calibration generated from L1 data 

#954 


docs/source/data_formats/astrom.rst
```diff
@@ -471,6 +471,10 @@ Image Header (HDU 1)
...
 | F0POS      | float      | 80.55342878952176              | -69.51409683261818                                                                                                       |
 +------------+------------+--------------------------------+--------------------------------------------------------------------------------------------------------------------------+
+| F1POS      | float      | 80.540037490674                | -69.51279841553037                                                                                                       |
++------------+------------+--------------------------------+--------------------------------------------------------------------------------------------------------------------------+
+| F2POS      | float      | 80.54532306822296              | -69.5163111965594                                                                                                        |
++------------+------------+--------------------------------+--------------------------------------------------------------------------------------------------------------------------+
```
semaphoreP, line 480: question: for your test data, did you have data from 3 different dithers? 

I wonder if we should explain that some of the keyword patterns could have multiple depending on number of dithers, number of stars used, etc. It could be done at the very top of this .rst file (above FITS data structure)

tests/e2e_tests/l1_to_astrom_e2e.py
```diff
@@ -0,0 +1,94 @@
...
+    # check that the recovered platescale, north angle, and offsets match up
+    astrom_cal_file = glob.glob(os.path.join(l2b_outputdir, '*_ast_cal.fits'))[0]
+    astrom_cal = data.AstrometricCalibration(astrom_cal_file)
+    actual_platescale = astrom_cal.platescale
+    actual_north_angle = astrom_cal.northangle
+    actual_ra_offset  = astrom_cal.avg_offset[0] * 3.6e6 # convert from deg to mas
+    actual_dec_offset = astrom_cal.avg_offset[1] * 3.6e6
+    assert expected_platescale == pytest.approx(astrom_cal.platescale, rel=0.05)
+    assert expected_north_angle == pytest.approx(actual_north_angle, abs=0.05)
+    assert expected_ra_offset == pytest.approx(actual_ra_offset, abs=30)
```
semaphoreP: Hmm, this is erroring for me. I get:

```
E       assert 436.0 == 466.176877755219 ± 30
E         
E         comparison failed
E         Obtained: 436.0
E         Expected: 466.176877755219 ± 30

tests/e2e_tests/l1_to_astrom_e2e.py:75: AssertionError
```

I wonder if this is yet another file order thing, since I noticed there's a keyword that combines frames `frames_to_combine` that is hard-coded to 24 in the default recipe. Ideally, we combine all frames that coorespond to a single dither. Are there header keywords that handle that? Does @manduhmia have thoughts on how to make `frames_to_combine` more flexible?

semaphoreP (commented):
Hmm this e2e test isn't passing for me. I can upload my reduced data if it's useful to compare again.

docs/source/data_formats/astrom.rst
everlastingEric: yeah, the data was partially based on the existing CAR test data so I have three different fields, except that I took out the dithers within each field because I didn't want to mess around with the default `files_to_combine` keyword since unless I change the number frames with different dithers would get combined together.

tests/e2e_tests/l1_to_astrom_e2e.py
everlastingEric: I did notice that also and I don't think right now there are any checks on which files are getting combined together. I have 24 frames per dither to match the number the pipeline combines and I put them in name order inside the test directory, but if glob mixes them up beforehand then different dithers could be getting combined together. I explicitly sorted the input data but not sure if there's other places in the pipeline that might still put the files out of order/

semaphoreP: I think it'd be more robust if we could split the datasets by some header keywords to separate the various dithers, and then combine all exposures in a subset (rather than hard coding the number of consecutive frames to combine). Maybe worth asking in slack to understand if such a keyword exists?

everlastingEric: Just throwing an idea out here, not sure if there's any reason I'm unaware of that makes this unfeasible. Instead of having a fixed number of frames to combine, would it be better to first group the frames by target, dither, roll, and other instrument configurations and combine based off those groupings?

everlastingEric: yeah just saw this while I was typing that out, I think a lot of other steps in the pipeline already implement this so I could take a look and try to find some good examples

semaphoreP: I think the main question for me is whether there is a header keyword we can use or not.

everlastingEric: I'm working on the splitting mechanism right now. So far I'm using TARGET, PA_APER, and FSMX/FSMY to distinguish between different observations, rolls, and dithers. Do you think there might be anything missing that I should ask about or are you thinking of doing it with a singular keyword specific to the data for astrometric calibration?

semaphoreP: I think as long as you check that different dithers are equivalent to changes in FSMX/Y and we wouldn't expect FSMX/Y to change during an astrometric calibration sequence otherwise

everlastingEric: I see. To double-check how the dithering is done in the context of astrometric calibration, is it purely through adjusting the spacecraft pointing rather than moving the FSM around? I was a bit confused since the corresponding CAR 125 issue mentioned a 3x3 FSM dither for each boresight visit.

semaphoreP: No, I think you're probably right that it is done with FSM mirror. I think if it's written in the CAR, then that's probably indeed true. I just wanted to double check.

everlastingEric: Ok I implemented the dataset splitting, I also widened the tolerance of the RA/Dec offset check a little bit since I noticed the calibration isn't entirely deterministic and occasionally it will split out values that goes over the 30 mas tolerance by 1-2 mas or so

semaphoreP: Thanks! this is a bit surprising we have to go over the 30 mas threshold, as I feel like we should be doing better than 30 mas by far with the data we have.

everlastingEric: The 72 frames I have do get reduced down to only 3 because each set of 24 essentially has the same configuration and only differs by detector noise. Should I maybe add some dithers back in to give the calibration more data points? That would also be a better check for the splitting code I wrote. I would just have run the sims overnight since each unique scene takes around an hour to generate on my end

semaphoreP: What happens if you just run the test with just one dither? does it do better? I think we should get better than 30 mas performance even with one dither.

everlastingEric: hmm yeah I just printed the individual frames. 2/3 gets the platescale and northangle basically on point but the offset differs 20-30 mas from what I computed to be the expected. One of the frames is just completely off for some reason. Specifically it gave this: platescale=19.9859  northangle=-82.0473  offset=(3500.3, 8903.6) mas. This frame might be another issue to figure out, but if there's nothing wrong with how I'm computing the expected offset, could the calibration results be a systematic that shifts the center by a pixel or so?

semaphoreP: the last one probably found the wrong constellation of stars. Maybe need to ask @manduhmia to help debug at some point, or you can dive into the star finding code. 

I'm going to be out of town for the rest of the week, but maybe you can slack Amanda and figure out the offsets together. There shouldn't be an offset, but how did you figure this out `dx_pix, dy_pix = 532 - 512, 505 - 512 `. Seems a bit hard coded to me?

everlastingEric: Yeah that sounds good, the detector offset is what the sim script for the CAR tests coded in so I just assumed they purposely added it for testing purposes?

semaphoreP: but surely that can't be the same for every dither?

everlastingEric: Just checked again, I think this line gets hard coded for everything pretty much `sim_scene = detector.generate_detector_image(sim_scene, exptime, full_frame=True, loc_x=532, loc_y=505)`, regardless of dither the scene is placed on the detector at that location.

semaphoreP: I see. but the dither will also move the FOV, and the astrometric calibration just reports the offset between the commanded RA/Dec location and 512/512, regardless of things like offsets and dithers.

everlastingEric: I think the reason why I was checking for one constant offset across all frames is because I did take out the actual FSM dithers. In the corgi-car-tests directory there are three CSVs containing the list of stars for each of the three simulated fields, and the simulation script gives a RA/Dec center for each field. Because the change in RA/Dec between each field is already tracked and I'm not adding in any additional dither, I just assumed all three would give the same offset corresponding to the injected detector placement

everlastingEric: I figured out the offset issue. Corgisim places the image directly at location 532, 505 in the array, but since the array is 0-indexed, the actual location corresponds to 533, 506 in a 1-indexed system. That one pixel shift basically takes care of the offset.

everlastingEric: Gonna try to figure out the issue with that one frame next

everlastingEric:
Ok so an update, the reason why that particular dither was giving wrong results is that one of the stars used for matching isn't found in the list of 50 brightest catalog stars. I decreased the default search radius so fewer stars outside the field of view are considered for matching and now all three dithers are giving the right results. With the offset I just fixed the calculation on the testing end and narrowed the tolerance down to 10 mas.

corgidrp/astrom.py
```diff
@@ -1035,28 +1036,60 @@ def boresight_calibration(input_dataset, field_path='JWST_CALFIELD2020.csv', fie
...
+                        "frames": [frame],
+                    })
+                else:
+                    cluster = clusters[best_idx]
+                    cluster["frames"].append(frame)
+                    cluster["sum_sin"] += pa_sin
+                    cluster["sum_cos"] += pa_cos
+                    cluster["pa_center"] = np.degrees(
+                        np.arctan2(cluster["sum_sin"], cluster["sum_cos"])
+                    ) % 360.0
```
semaphoreP, line 1072: is this a meaningful edge case to solve? I'm wondering if for simplicity it's just easier to compare against the first frame's PA_APER.

corgidrp/astrom.py
```diff
@@ -1035,28 +1036,60 @@ def boresight_calibration(input_dataset, field_path='JWST_CALFIELD2020.csv', fie
...
-            im = corgidrp.data.Image(comb, pri_hdr=input_dataset[count - frames_to_combine].pri_hdr, ext_hdr=input_dataset[0].ext_hdr)
-            im.filename = file_name
+        for grouped_dataset in grouped_datasets:
+            # grab all frames in the dataset
+            data_array = [frame.data for frame in grouped_dataset]
+            # median combine
+            comb = np.nanmedian(data_array, axis=0)
+            # create image object
+            im = corgidrp.data.Image(comb, pri_hdr=grouped_dataset[0].pri_hdr, ext_hdr=grouped_dataset[0].ext_hdr)
+            im.filename = grouped_dataset[-1].filename
```
semaphoreP, line 1089: why do you take the filename of the last frame in the sequence but the header of the first? Should we be consistent on this?

tests/e2e_tests/l1_to_astrom_e2e.py
```diff
@@ -72,8 +75,8 @@ def test_l1_to_astrom_e2e(e2edata_path, e2eoutput_path):
     actual_dec_offset = astrom_cal.avg_offset[1] * 3.6e6
     assert expected_platescale == pytest.approx(astrom_cal.platescale, rel=0.05)
     assert expected_north_angle == pytest.approx(actual_north_angle, abs=0.05)
-    assert expected_ra_offset == pytest.approx(actual_ra_offset, abs=30)
-    assert expected_dec_offset == pytest.approx(actual_dec_offset, abs=30)
+    assert expected_ra_offset == pytest.approx(actual_ra_offset, abs=35)
+    assert expected_dec_offset == pytest.approx(actual_dec_offset, abs=35)
```
semaphoreP: can we reduce these tolerances now?

semaphoreP (commented):
thanks for digging through this and finding the root cause. I had a few minor comments/questions, nothing show-stopping but just wanted to understand implementation decisions. 

Edit: oops it looks like the astrom_e2e.py test is breaking due to the filename issue.

corgidrp/astrom.py
everlastingEric: I copied this chunk of code from another step function that splits by PA. I could do the direct frame comparison but I think that's going to require a nested for-loop if that's not a big deal.

corgidrp/astrom.py
everlastingEric: That was how the header/filename was when the frames were combined even before I updated this, should I change this to make is consistent?

tests/e2e_tests/l1_to_astrom_e2e.py
everlastingEric: ah yes I forgot to do that

corgidrp/astrom.py
semaphoreP: whoops I spoke too soon: I see it is trying to imitate the previous logic. However, the new logic causes the astrom_e2e.py test to now fail: 

```
>       assert astrom_cal.filename.split('l2b')[-1] == expected_last_filename + '_ast_cal.fits'
E       AssertionError: assert 'cgi_00890010..._ast_cal.fits' == 'cgi_00890010..._ast_cal.fits'
E         
E         - cgi_0089001001001001027_20240308t0648092_ast_cal.fits
E         ?                      ^^        ^    ^^^^
E         + cgi_0089001001001001031_20240309t0612244_ast_cal.fits
E         ?                      ^^        ^   +++ ^

tests/e2e_tests/astrom_e2e.py:300: AssertionError
```

semaphoreP: I think we do want to preserve the original file name behavior.

semaphoreP:
@ecady-jpl for your awareness, this PR updates the default astrometric calibration in two ways. I think you probably don't care about these changes, but wanted to keep you in the loop in case it matters
 * Changes the behavior of the `combine_frames` keyword in `boresight_calibration` step to be True/False (whether all frames of a given target + PA_APER + FSMX/Y should be median combined). Previously it was an int (number of frames to combine), which was rather inflexible and required hard coding. 
 * Cuts the defaults search radius by a factor of 2, which I think makes it go from a 43 arcsec to a 21 arcsec search radius. (also tagging @manduhmia for awareness)

Let us know if you have any questions/concerns about this.

corgidrp/astrom.py
semaphoreP: ok no worries then. It's fine as is.

ecady-jpl:
Those seem like reasonable updates to me.

corgidrp/astrom.py
everlastingEric: hmm I'll look at that, but it seems that since the frame combining was relatively straightforward before, the filename of the calibration matched up with the last file of the input data. But now since we are splitting the dataset with a bunch of keywords, I'm not sure if the order is preserved anymore? I did run the test and it passes if we explicitly sort the input data filelist beforehand

semaphoreP: the expectation is that the output file inherits the last filename's base by UTC time. Maybe we should explicitly sort by UTC time to find the filename? I wonder if this code exists somewhere else that we can borrow.

everlastingEric: that's a good idea. Quick clarification, by the last filename, do we care if the files might be in random order or should the filename inherit the specific file with the latest UTC time?

semaphoreP: Hmm I think lines 2151-2166 in data.py already do this for AstrometricCalibration so I'm not sure what's happening.

semaphoreP: maybe we need to copy those lines and also do it here when you are doing your intermediate combination.

semaphoreP: or pass the original non-combined frames as the input dataset to the creation of the AstrometricCalibraiton object, so this is handled with the existing code.

everlastingEric: I did notice that, I think what I'll add is that during the frame combining step, I'll explicitly sort the frames by filename to make sure the name of the combined frame is always set to the filename with the latest UTC time within that grouping

everlastingEric: or actually I should just copy the existing code in data.py that seems a lot more robust than just sorting

semaphoreP: to maximize code reuse, maybe slightly easier to do `in_dataset = grouped_dataset[i]` and just pass that into the creation of the astrometric calibraiton.

semaphoreP: that way you can just use the sorting code that's already there

everlastingEric: it seems that the astrometric calibration class checks the headers to make sure an actual calibration is passed in. What do you think about pulling that chunk of code out of the object initialization and putting it in a separate helper function instead? This way I can just directly call the helper

everlastingEric: wait I see what you mean I was looking at the wrong place, yeah that should work I'll try it

everlastingEric: ok I made some updates to the code, a part of the issue was that the old e2e test wasn't checking for the correct filename either. The test assumes the last file in the filelist is the one with the latest date but that was not the case, so in addition to changing the calibration code to select the correct name I also explicitly sorted the files by the date/time inside the test.

tests/e2e_tests/l1_to_astrom_e2e.py
everlastingEric: tolerance should be set to 10 now

semaphoreP (approved):
Looks good, thanks!

PR #912, opened 2026-05-27
Speedups for tests

adrienmaillard:
## Describe your changes
Some changes to speed up the tests. The runtime of tests in this PR is 31mins, compared to the 48mins of other PRs. On my machine it went from 28 mins to 12 minutes.

- @semaphoreP, you replaced the full-array generic_filter(np.nanmedian) calls for bad pixel correction — which invoked
pure-Python nanmedian once per pixel (~8M calls on a 1024x1024 pol dataset) with a hybrid strategy
- @semaphoreP also sped up test_ct_interp 21x by avoiding redundant CT cal rebuilds
- Replaced pytest.approx (does element by element comparison) with np.allclose (vectorized) for array comparisons
- Disabled multiprocessing in find_source for find_source tests and fluxcal tests
- For the wcs, Claude suggested to make the test data smaller and indeed it did not affect the test results while providing a good speedup 

### Parallelization efforts 
I tried to make changes to be able run tests in parallel via github actions. Public repos have 4 cores and we only use one when running pytest. I spent a lot of time trying to make them work but in the end these changes are not on this branch but on branch [enable-parallel-testing](https://github.com/roman-corgi/corgidrp/tree/enable-parallel-testing) so I can move on to more urgent tasks. Most issues had to do with concurrent read/write on the same folders. I had to make tmp folders to avoid collisions. I made it work locally but I think the github.com resources are too limited to run some of the big tests in parallel without OOM. 

## Type of change

- Bug fix (non-breaking change which fixes an issue)

## Reference any relevant issues (don't forget the #)
#714 

## Checklist before requesting a review
- [x] I have verified that all unit tests pass in a clean environment and added new unit tests, as appropriate
- [x] I have checked that I am merging into the right branch
- [ ] I have checked the output of the latest Github Actions run associated with this PR and confirmed running `pytest` did not produce any warnings
- [x] I have checked if my code modifies an existing step function or modifies the data format docs. If it does, I've added my PR to the [list maintained here](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/212028061/Log+of+PRs+that+could+affect+existing+VAP+Tests).

PR #981, opened 2026-07-27
Enable parallel unit and e2e testing

adrienmaillard:
## Describe your changes 

I tried to make tests run in parallel so they take less time to execute for developers. The main problem was linked to multiple tests writing .corgidrp and the caldb so they had to be isolated in their own tmp folder with their own caldb. A lot of tests can be ran in parallel. 

On my mbp m5 pro max (18 cores), all the unit tests now pass in 8mins30.  

### Changes by type

Worker isolation infrastructure
- corgidrp/__init__.py: config/caldb/default-cal paths now derive from a per-worker temp dir (CORGIDRP_WORKER_TMP) when set, falling back to ~/.corgidrp otherwise. Config read uses config_folder
instead of a hardcoded home path.
- tests/conftest.py (new): pytest_configure gives each xdist worker a private .corgidrp temp dir and re-derives corgidrp path globals (handles the fork start method used in CI, where workers inherit
 already-bound paths).
- conftest.py: session-scoped autouse fixture initializes caldb once per worker.

Concurrency-safe directory creation
- Replaced os.mkdir with os.makedirs(..., exist_ok=True) in calibrate_kgain.py, calibrate_nonlin.py, fluxcal.py, and mocks.py (multiple sites) so parallel workers don't race on directory creation.

Serial test handling
- Added a serial marker (pytest.ini) and tagged order-dependent / shared-state tests across 11 unit files plus the E2E data-format consumer.
- tests/conftest.py: pytest_collection_modifyitems sorts items by file/line so serial tests keep deterministic order under --dist loadscope.

test refactors for parallel safety
- Rewrote tests that mutated module-level globals to use local variables / session-scoped fixtures, eliminating cross-test race conditions and redundant data generation.

pytest configuration
- pytest.ini: --dist loadscope (keeps a module's tests on one worker; needed for serial ordering) and the serial marker definition.

CI workflows
- python-app.yml / e2e-testing.yml: install pytest-xdist; run -n auto -m "not serial" then a -m "serial" pass; cap workers to 2 on runners (PYTEST_XDIST_AUTO_NUM_WORKERS); added memory monitoring +
OOM check to the unit workflow.

docs
- README.md: documented the two-command parallel + serial run pattern for both suites.

## Type of change

#912 

- [ ] I have checked that I am merging into the right branch

### DIFF
--- a/.github/workflows/python-app.yml
+++ b/.github/workflows/python-app.yml
@@ -30,12 +30,46 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install flake8 pytest flake8-docstrings-complete
+        pip install flake8 pytest pytest-xdist flake8-docstrings-complete
         if [ -f requirements.txt ]; then pip install -r requirements.txt .; fi
     - name: Lint with flake8
       run: |
         # stop the build if there are Python syntax errors or undefined names
         flake8 . --count --select=E9,F63,F7,F82,DCO020,DCO021,DCO022,DCO023,DCO024,DCO030,DCO031,DCO032,DCO060,DCO061,DCO062,DCO063,DCO064,DCO065 --show-source --statistics
-    - name: Test with pytest
+    - name: Start memory monitor
       run: |
-        pytest
+        # log memory every 5s in the background so we can see growth before a cancel/OOM
+        ( while true; do
+            echo "[$(date +%H:%M:%S)] $(free -m | awk '/Mem:/ {print "used="$3"MB free="$4"MB avail="$7"MB"}')";
+            sleep 5;
+          done ) > mem.log 2>&1 &
+        echo "MEM_MON_PID=$!" >> "$GITHUB_ENV"
+    - name: Test with pytest (parallel)
+      env:
+        CORGIDRP_DO_NOT_AUTO_INIT_CALDB: "1"
+        PYTEST_XDIST_WORKER_START_METHOD: fork
+        # cap `-n auto` on the CI runner; local dev leaves this unset to use all cores
+        PYTEST_XDIST_AUTO_NUM_WORKERS: "2"
+      run: |
+        pytest -n auto -vv -ra --durations=20 --max-worker-restart=0 -m "not serial"
+    - name: Test with pytest (serial)
+      env:
+        CORGIDRP_DO_NOT_AUTO_INIT_CALDB: "1"
+      run: |
+        pytest -vv -ra --durations=20 -m "serial"
+    - name: Memory log
+      if: always()
+      run: |
+        echo "===== memory timeline ====="
+        cat mem.log || echo "no mem log"
+        echo "===== peak ====="
+        sort -t= -k2 -n mem.log 2>/dev/null | tail -1 || true
+    - name: Check for OOM kills
+      if: always()
+      run: |
+        echo "===== dmesg OOM check ====="
+        sudo dmesg | grep -iE "killed process|out of memory|oom" || echo "No OOM events found in dmesg"
+    - name: Stop memory monitor
+      if: always()
+      run: |
+        kill "$MEM_MON_PID" 2>/dev/null || true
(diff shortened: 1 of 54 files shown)


semaphoreP:
Is this ready to review (review requested) or still in progress (in draft stage). If it's ready for review, can you give a more detailed summary of the changes? There looks like there's a lot of changes across many files, and it would make review easier to understand why all these changes are necessary.

maxwellmb:
It also looks like it's enabled automatic e2e tests, which we've specifically avoided since it's expensive. Is this a change we wanted to make?

adrienmaillard:
@semaphoreP I guess it'll be ready when I write a better description ;) 
@maxwellmb I am not sure what you mean, can you tell me what you refer to ? The e2e tests are still not included in the standard workflow. And in the e2e test workflow, there are just 2 branches instead of one for tests who can be ran in parallel and those who can't.

maxwellmb:
Maybe I'm wrong! Ignore me for the moment!

pytest.ini
```diff
@@ -1,3 +1,9 @@
 [pytest]
 norecursedirs = .git .vscode build docs dist
 python_files = test_*.py *e2e.py
+# loadscope keeps all tests from one module on the same worker (required for
+# module-level state and correct serial ordering). It only takes effect when
+# running under xdist (-n ...); harmless for a plain sequential run.
+addopts = --dist loadscope
```
semaphoreP: we should add `pytest-xdist` to the requirements.txt since otherwise unit tests won't run for developers (even ones that don't need parallelization).

README.md
semaphoreP: developer instructions got moved to the docs folder as .rst files. can you move your new section over there too?

.github/workflows/python-app.yml
```diff
@@ -30,12 +30,46 @@ jobs:
...
+            echo "[$(date +%H:%M:%S)] $(free -m | awk '/Mem:/ {print "used="$3"MB free="$4"MB avail="$7"MB"}')";
+            sleep 5;
+          done ) > mem.log 2>&1 &
+        echo "MEM_MON_PID=$!" >> "$GITHUB_ENV"
+    - name: Test with pytest (parallel)
+      env:
+        CORGIDRP_DO_NOT_AUTO_INIT_CALDB: "1"
+        PYTEST_XDIST_WORKER_START_METHOD: fork
+        # cap `-n auto` on the CI runner; local dev leaves this unset to use all cores
+        PYTEST_XDIST_AUTO_NUM_WORKERS: "2"
```
semaphoreP: we seem dangerously close to the memory limit of the github actions unit tests (only ~300 MB free). Is there 1 or 2 particular unit tests that are gobbling memory that we should be putting into the serial running part.

semaphoreP:
Thanks for doing this and providing the summary. I started reviewing this PR, and I have a couple concerns about long-term sustainability of doing this. Basically, I'm wondering if  adding this is worth it in terms of future maintenance (e.g., years down the line). In my past experience, maintaining parallelism always adds some significant time. 

It seems like committing to parallelized pytest requires developers moving forward to think about marking whether their test supports parallelism or not. It also requires developers to write unit tests that support an optional tmp_path to place the output in. Combined with all the existing procedure for submitting a PR, I'm worried about adding too many hoops for us to jump through for a 2x speed up on tests. Is defining custom output directories necessary for tests to pass?

It also seems like we're barely squeezing by in terms of memory footprint right now. I'm also a bit worried that if we add some unit tests in the future, we're going to run out of memory, and possibly have to go back down to 1 thread, erasing the gains. How many tests take a lot of memory, and could some be punted to serial processing?

adrienmaillard:
I totally see your point. I was also wondering about that.

Temp folders are necessary because of the `.corgidrp` folders. Many tests write in these folders so if you run them in parallel, you have race conditions.  

There is a limited gain on the github workers because we are constrained in terms of memory indeed. Biggest gain is for local testing. It's a pain that I have to wait 10s of minutes to run unit tests when my computer has 18-1 idle cores. 

I agree this would require more training for developers and probably one more step in every PR. I'm okay if you just let it sit. Your call.

semaphoreP:
Interested in hearing what @maxwellmb thinks when he's back. We can maybe decide collectively at the in-person meeting.

PR #984, opened 2026-07-30
improve spec PSF registration

neilzim:
## Describe your changes

The least-squares fit used in the PSF registration in the fit_psf_centroid() function of spec.py was insufficiently constrained for noisy images. Adding an initial cross-correlation stage to provide initial values for position and flux before applying the Powell method least-squares fit.

## Type of change

Fixes the root cause of the failure reported in #938:

- [X] I have checked the output of the latest Github Actions run associated with this PR and confirmed running `pytest` did not produce any warnings
- [X] I have checked if my code modifies an existing step function or modifies the data format docs. If it does, I've added my PR to the [list maintained here](https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/212028061/Log+of+PRs+that+could+affect+existing+VAP+Tests).


JuergenSchreiber:
Thanks Neil, I merge this with the rotten fits branch and run the according e2e tests. Let me try to understand what you did: you determined the initial guess offsets (which were 0 before) much better by an cross correlation (fft_convolve). You additionally set boundaries to the initial guesses in the scipy.optimize.minimize functions, which is by the way a very powerful function with a lot of options...I am impressed... I kept the crop function in the e2e test recipes for spectroscopy, since I think the interesting portion of the spectrum is very small compared to the big frame, but I think it would also work with the big frame. What do you think?

neilzim:
Thanks for looking at this. Exactly as you observed, I ran a few tests with a "sandbox" script to show that the cross correlation is more robust for getting a decent starting point. Then the least-squares fit can be restricted to a small range of parameters.
I agree the image crop operation is still a sensible step, so let's keep that option in the recipe.

neilzim:
Closing this draft PR since @JuergenSchreiber has merged the PSF registration function changes into the branch https://github.com/roman-corgi/corgidrp/tree/rotten_fits, to be applied in PR#938.

Issue #966, opened 2026-07-17
Update l1 to corethroughput e2e test to also generate CTMap


semaphoreP:
**Is your feature request related to a problem? Please describe.**
CT Map is also a calibration product we are supposed to be able to produce. It's currently tested as part of ctmap_e2e.py but not l1_to_corethroughput_e2e.py. We should also add it to the latter to check everything is fine when starting from L1. 

**Describe the solution you'd like**
Update l1_to_corethroughput_e2e.py to proudce the CT Map after the regular corethroughput proudct is generated.

semaphoreP:
Implemented in PR #985

PR #985, opened 2026-07-30
Add l1 to corethroughputmap e2e test + more robust corethroughput calibration/test

everlastingEric:
## Describe your changes
- Extended the l1 to corethroughput e2e test to use another l1 coronagraphic dataset and the generated corethroughput calibration to create a corethroughput map calibration
- Modified the l2a to corethroughput map recipe template to be able to process l2a data correctly without throwing unit conversion errors
- In the process of adding this new test, I discovered the old corethroughput calibration being generated was grabbing onto cosmic ray/detector noise instead of the actual stellar PSFs. To make the pipeline more robust against this issue, I added in the option to median-combine frames of the same dither (based on FSMX/FSMY keywords) so the centroiding algorithm is more robust against noise
- Similarly, I found that any NaN pixels in the pupil images causes the NaN to propagate to the pupil-integrated flux and therefore also the actual core throughput fraction. To fix this, pupil frames can now also be median-combined before their fluxes are integrated and averaged. In addition, I changed the flux integration method from sum to nansum so that it is immune to any remaining NaNs in the median-combined pupil image. 
- Updated the corethroughput calibration to set its FPAM/FSAM headers to match that of the coronagraphic data rather than just the first frame of the input dataset, which could be a pupil image. Since these keywords are needed to compute the FPM center for the corethroughput map calibration, having them correspond to that of the pupil image would be incorrect, as the FPM is taken out in a pupil image
- To test that the updated corethroughput calibration can now find the stellar PSFs correctly, I updated the test to check that the PSF locations it returns matches what was injected into the input data
- Updated the input data used for the corethroughput calibration test (less dithers but more frames per dither), and added additional input data for testing the corethroughput map. The data are found in the `corethroughput_sims_HLC_B1` and `ctmap_sims` sims subfolders in the e2e test data sharepoint
- Since these new changes cause previous e2e tests and unit tests to break (due to code compatibility issues), I went through and made updates to those tests as well so they work again
- Updated `test_zz_dataformats_e2e.py` to check the header of the new corethroughput map calibration generated with the simulated data

## Type of change

#966 


tests/e2e_tests/l1_to_corethroughput_e2e.py
```diff
@@ -80,17 +80,91 @@ def test_expected_results_band1_nfov_e2e(e2edata_path, e2eoutput_path):
...
+            expected_loc = expected_psf_locs[i]
+            # distance between this particular recovered psf location to this particular expected psf location
+            dist = np.sqrt((loc[0] - expected_loc[0])**2 + (loc[1] - expected_loc[1])**2)
+            # check if this is smaller than the current minimum distance measurement, update if so
+            if dist < min_dist:
+                min_dist = dist
+                min_dist_idx = i
+        # assert the minimum distance difference is within a small tolerance
+        tol = 3
+        assert min_dist < tol
```
semaphoreP: Q: this is a 3-pixel tolerance? Does the test fail if go to a smaller tolerance?

tests/e2e_tests/l1_to_corethroughput_e2e.py
```diff
@@ -80,17 +80,91 @@ def test_expected_results_band1_nfov_e2e(e2edata_path, e2eoutput_path):
...
+
+    # grab ctmap file
+    ctmap_file = glob.glob(os.path.join(ctmap_outputdir,
+        '*ctm_cal.fits'))[0]
+    ct_map = data.CoreThroughputMap(ctmap_file)
+
+    # check headers
+    check.compare_to_mocks_hdrs(ctmap_file)
+    assert ct_map.ext_hdr["DATATYPE"] == "CoreThroughputMap"
+    assert ct_cal_drp.ext_hdr["DATALVL"] == "CAL"
```
semaphoreP, line 169: is it straightforward to add some checks on the correctness of the values of the CTMap?

semaphoreP (commented):
This generally looks good. Just a couple of questions on the testing goals.

semaphoreP:
For awareness tagging:

@agreenbaum about the changes to the core throughput map data format docs. Main changes are just more accurate set of header keywords created by the DRP (certain step functions are not run, and others are). I don't think this should be a significant impact. 

@juliamilton since they are the owner of the core throughput calibration files for the upgrades to the processing.

tests/e2e_tests/l1_to_corethroughput_e2e.py
everlastingEric: Think the lowest I can narrow the tolerance down to is 2.3, is that not accurate enough?

tests/e2e_tests/l1_to_corethroughput_e2e.py
everlastingEric: I could probably add some header checks but I'm not really sure how I would check the actual values unless there's some reference I could use.

tests/e2e_tests/l1_to_corethroughput_e2e.py
semaphoreP: I'm surprised we can't do better than 1 pixel.

tests/e2e_tests/l1_to_corethroughput_e2e.py
semaphoreP: can we at least check they are values between 0 and 1?

tests/e2e_tests/l1_to_corethroughput_e2e.py
everlastingEric: I see I'll look into that a bit more

tests/e2e_tests/l1_to_corethroughput_e2e.py
everlastingEric: yes that's a good idea

tests/e2e_tests/l1_to_corethroughput_e2e.py
everlastingEric: I misspoke earlier from memory about the 2.3 pixels mismatch, if I check individual x and y offsets, most of the PSFs are matched to less than 1 pixel but there are a few outliers. Does this seem fine or should I try to get it to be more accurate still? 
```dist: 1.044407809034142, dx: -0.9999703223552387, dy: 0.3014084039642171
dist: 1.6913864503819587, dx: -1.0592964942302387, dy: -1.318589800448649
dist: 0.9272479501397608, dx: -0.9096035738604087, dy: -0.18002805186625892
dist: 1.3045641950945754, dx: -0.9277920504229087, dy: -0.9171093993056729
dist: 1.0223736435804358, dx: -0.7194684229859831, dy: -0.7263697787036563
dist: 0.7702556576670416, dx: -0.6780865870484831, dy: -0.3653660611400937
dist: 0.7979920911136456, dx: -0.7814916208249656, dy: -0.16143798828125
dist: 0.4309248040826272, dx: -0.36987976634173947, dy: -0.221099853515625
dist: 0.4431424862148644, dx: 0.33075119158542066, dy: -0.294921875

semaphoreP: I'm not at my computer right now. Can you remind me where the PSF locations are placed on the detector? Are the first few further away from mask, or closer in?

everlastingEric: I printed it out along with the distance between the found and actual location. It would make sense that the PSFs closer to the mask would be suppressed more and their locations less accurate but I didn't really see much of a correlation here. 
```dist: 1.044407809034142, expected_psf_location: (522.7798165137615, 493.32881010189516)
dist: 1.6913864503819587, expected_psf_location: (518.8807339449542, 500.08221921397563)
dist: 0.9272479501397608, expected_psf_location: (518.8807339449542, 500.08221921397563)
dist: 1.3045641950945754, expected_psf_location: (515.6697247706422, 505.64385024745366)
dist: 1.0223736435804358, expected_psf_location: (515.6697247706422, 505.64385024745366)
dist: 0.7702556576670416, expected_psf_location: (515.6697247706422, 518.3561497525463)
dist: 0.7979920911136456, expected_psf_location: (490.4403669724771, 512.0)
dist: 0.4309248040826272, expected_psf_location: (490.4403669724771, 512.0)
dist: 0.4431424862148644, expected_psf_location: (490.4403669724771, 512.0)

semaphoreP: ok I reviewed these psf stamps by eye and I think only the 4th one is obviously off by 0.5-1 pixel. The rest might be off due to the PSF distortion actually distorting the true position of the source (a reason why we may actually use simulated PSFs for core throughput rather than real ones). I think it's fine for now.

semaphoreP (approved):

PR #990, opened 2026-08-05
Pc dq fix and string recipe length in output FITS files

### LINKED ISSUE #978: detect_cosmic_rays fails when EMGAIN_M is present
`detect_cosmic_rays()` fails when `EMGAIN_M` is present due to the indent at [line 244](https://github.com/roman-corgi/corgidrp/blob/develop/corgidrp/l1_to_l2a.py#L244): 

```
for frame in initial_dataset:
        try: # use measured gain if available
            emgain = frame.ext_hdr['EMGAIN_M']
        except:
            if frame.ext_hdr['EMGAIN_A'] > 0: # use applied EM gain if available
                emgain = frame.ext_hdr['EMGAIN_A']
            else: # otherwise use commanded EM gain
                emgain = frame.ext_hdr['EMGAIN_C']
            emgain_list.append(emgain)
```

The final line is indented one level too far, and if `EMGAIN_M` is present, it defines `emgain` each loop iteration but does not append it to the list.  It then fails when you get to [line 253](https://github.com/roman-corgi/corgidrp/blob/v5.0/corgidrp/l1_to_l2a.py#L253) and your arrays have a size mismatch.

### LINKED ISSUE #980: PC function should flag pixels too bright for PC rather than crash
Currently, a bright pixel in a frame otherwise suitable for photon counting will cause the PC function to crash.

Issue a warning instead, and flag the pixel with a new DQ flag specific to this scneario.

**Additional context**
This will allow smoother PC processing of frame stacks.  In theory, unwanted bright pixels in the region of interest (e.g., spot of Arago) should be flagged already by the detect_cosmic_rays() function and handled in the bad pixel application function later on in the pipeline after PC processing, but flagging these pixels in this PC context will provide even more safety.

kjl0025:
## Describe your changes
Negative values for photon-counted frames are masked with DQ value 256 now instead of resulting in a crash.  (These are pixels that have a mean electron count of more than ~0.25, making the iterative Newton's method which accounts for photometric corrections give a negative value, and such mean electron count values are inappropriate for photon counting.)

I also fixed the indent in the code so that the presence of the header EMGAIN_M does not cause a crash in the cosmic ray function.

Also, the walker currently records the full recipe (which includes a list of all input files) in the header of all FITS files that are not produced in RAM-heavy mode.  However, RAM-heavy products have recipes in their recipe chains which are run in chunk processing (i.e., not RAM-heavy mode where the data is only loaded in one frame at a time), and all those FITS files get a huge recipe string write to the header which makes the files larger and slows down the processing noticeably.  I changed the walker to only produce the full list of input files in the recipe string in one of the output FITS files, and that file is specified in the recipe string in every other FITS file.

## Type of change

#980 
#978 


semaphoreP:
@kjl0025 can you push a trivial commit to re-launch the tests?

corgidrp/walker.py
```diff
@@ -797,12 +797,11 @@ def run_recipe(recipe, save_recipe_file=True):
...
                     curr_dataset = data.Dataset(filelist, no_data=True, no_err=True, no_dq=True)
-                    recipe_temp = recipe.copy()
-                    # don't want to keep all ~26000 filepaths in all ~26000 ext headers b/c that's a lot of memory
-                    recipe_temp["inputs"] = "See RECIPE header value in {0}".format(curr_dataset[-1].filepath)
                 else:
                     curr_dataset = data.Dataset(filelist)
-                    recipe_temp = recipe
+                recipe_temp = recipe.copy()
+                # don't want to keep all ~26000 filepaths in all ~26000 ext headers b/c that's a lot of memory makes walker take longer
+                recipe_temp["inputs"] = "See RECIPE header value in {0}".format(curr_dataset[-1].filepath)
```
semaphoreP: this now affects all recipes to not have the input files written into most files, which is not what we ant.

tests/test_walker.py
```diff
@@ -93,12 +93,19 @@ def test_autoreducing():
...
         hdr_recipe = json.loads(frame.ext_hdr["RECIPE"])
-        assert json.dumps(hdr_recipe) == json.dumps(recipe)
+        for key in hdr_recipe.keys():
+            if key == 'inputs':
+                continue
+            assert hdr_recipe[key] == recipe[key]
+        assert hdr_recipe['inputs'] == 'See RECIPE header value in ' + filelist[-1]
+
+    hdr_recipe_last = json.loads(output_dataset[-1].ext_hdr["RECIPE"])
+    assert json.dumps(hdr_recipe_last) == json.dumps(recipe)
```
semaphoreP: we don't want this behavior that only the last file has this by default.

semaphoreP (commented):
two comments (also discussed offline) regarding not modifying the current logic to save input files to the header.

corgidrp/walker.py
kjl0025: I just changed it back to the original

tests/test_walker.py
kjl0025: I just changed it back to the original

semaphoreP: I think you missed the ` for frame in output_dataset[:-1]:` change.

kjl0025: Ah, sorry.  One sec

semaphoreP (approved):

Issue #976, opened 2026-07-21
process pol mode CGIVST_CAL_TGTREF_PHOT images to L3 by default


maxwellmb:
The walker needs to make sure to process pol mode CGIVST_CAL_TGTREF_PHOT images to L3 by default. 

We probably want this as a chained recipe: ["l1_to_l2a_basic.json", 'l2a_to_l2b_pol.json','l2b_to_l3_pol.json']

shalverson:
FWIW, for my own CAR testing I just added a custom template to go directly from L1 to L3 for now:

https://github.com/roman-corgi/corgidrp/blob/pol_phot_to_l3/corgidrp/recipe_templates/l1_to_l3_pol.json

nasavbailey:
In troubleshooting this issue, we realized that CGIVST_CAL_TGTREF_PHOT should not be used as the default visit file for collecting polarized photometry, because it has a hardcoded FSAM = OPEN.  Instead the CGIVST_CAL_POL_SETUP visit type with only a single dither location is the default recommendation (see CAR-156). That said, there may be cases where we don't care that FSAM=open so it probably makes sense to maintain support for TGTREF_PHOT w/ WP selection rule.

I'm going to suggest adding that scope to this ticket, since it's related to generating the correct recipe for polarized target/ref photometry. But if you'd rather spin off into a separate ticket, please do!

The way to differentiate target/ref photometry collected by POL_SETUP from Mueller Matrix data collected by POL_SETUP is that target/ref will NOT dither the FSM. So the guess_template selection rule could send POL_SETUP to target/ref photometry recipe if only a single FSM XY position (within some small tolerance <lam/D) is detected in the headers, and send to MM recipe if multiple FSM XY dithers are detected.

shalverson:
hopefully addressed in this [PR](https://github.com/roman-corgi/corgidrp/pull/991)

semaphoreP:
Added in #991

PR #991, opened 2026-08-05
Pol cal files with no dither to l3

shalverson:
## Describe your changes
Added walker logic to change the recipe chain for pol setup observations based on dithers. If no dithers in image stack, produce l3 files and skip MM estimation.

## Type of change

Issue [976](https://github.com/roman-corgi/corgidrp/issues/976)

- [-] I have checked the output of the latest Github Actions run associated with this PR and confirmed running `pytest` did not produce any warnings -- SEVERAL WARNINGS PRODUCED, though these don't appear associated with code associated with this PR


corgidrp/walker.py
```diff
@@ -579,10 +577,18 @@ def guess_template(dataset):
             recipe_filename = ['trap_pump_cal_1.json', 'trap_pump_cal_2.json']
             chained = True
         elif image.pri_hdr['VISTYPE'] == 'CGIVST_CAL_POL_SETUP':
-            recipe_filename = ["l1_to_l2a_basic.json", "l2a_to_l2b_pol.json", "l2b_to_polcal.json"]
-            chained = True
+            # check if there are multiple FSM positions. If so, we need to do a polcal recipe. 
+            # if not, go to l3.
+            _, fsm_unique = dataset.split_dataset(exthdr_keywords=['FSMX', 'FSMY'])
```
maxwellmb: Vanessa pointed out at some point that the actual FSM values might drift slightly, and so rather than looking for unique values, we may need to set some kind of threshold instead. Have you looked into this at all?

corgidrp/recipe_templates/l1_to_l3_pol.json
maxwellmb, line 1: it doesn't seem like this recipe is ever used since in walker.py the recipes are all chained (which is preferred anyway). Can we get rid of it?

corgidrp/walker.py
maxwellmb: If we expect this, we should update the test as well.

maxwellmb (changes requested):
A couple of quick comments.

corgidrp/recipe_templates/l1_to_l3_pol.json
shalverson: Yup

corgidrp/walker.py
shalverson: Sure. That's an easy fix. What drift tolerance should we set? (~mas-level?) I'll add a helper function to return a dithered/not dithered bool

maxwellmb: I'm not sure what level of drift we expect. I think we'd want to ask someone, but I'm not sure who.

shalverson: For now I'll leave this at 1 mas as a default parameter while I track down what a reasonable number is

maxwellmb: Were you able to track down what a reasonable number is? Since it won't be easy to change this on the fly I think we want to make sure that we get it right the first time.

shalverson: Fair. I made some inquiries but haven't heard back so far.

semaphoreP: what's the typical dither size for the pol cal? is it much much larger than picking some intermediate value (e.g., 10 mas) should be safe?

shalverson: @semaphoreP @maxwellmb  the smallest dither pattern across all other .aux files is ~40 max. I proposed fixing the tolerance for this checker at 4 mas. I've made this change in the branch

maxwellmb (commented):
@shalverson can you take a look at the merge conflicts?

semaphoreP (approved):
E2e tests are failing but I've made issue #1023 to address that

Issue #989, opened 2026-08-05
Image class: cast of data in additional HDUs to the configured image_dtype does not work in all cases


JuergenSchreiber:
**Describe the bug**
in image.save() the hdu.data extensions are tried to be casted to the image type set in the corgidrp.cfg file.
There are cases especially in spec mode, where the hdu can be a binary table or a spec_dq type.

**To Reproduce**
add a save in a spec recipe after add_wavelength_map or extract_spec

**Expected behavior**
the loop should skip the cast if the hdu.data type is not an np.array, and set dq_type if it is a dq.

**corgidrp version:**
 - Release 5.0

**Additional context**

semaphoreP:
Fixed in #992

PR #992, opened 2026-08-06
when saving cast only np.ndarray in additional HDUs, and consider SPE…

JuergenSchreiber:
…C_DQ

## Describe your changes
in Image.save(): cast only np.ndarray and consider SPEC_DQ as dq_dtype.

## Type of change
- 
#989


corgidrp/data.py
```diff
@@ -590,10 +590,13 @@ def save(self, filedir=None, filename=None):
         dqhdu = fits.ImageHDU(data=self.dq, header = self.dq_hdr)
         hdulist.append(dqhdu)
 
-        # Cast data in additional HDUs to the configured dtype before appending
+        # Cast data in additional HDUs to the configured dtype before appending excluding binary tables etc.
```
semaphoreP: do you mean ", excluding binary tables etc.". With the added comma, that binary tables do not get modified.

semaphoreP (changes requested):
You need to update the dataformat pages for test_l4_spec_coron_dataformat_e2e and test_l4_spec_noncoron_dataformat_e2e due to these changes. Those data formats have changed due to this change. Otherwise, one clarification question.

corgidrp/data.py
JuergenSchreiber: Sorry, forgot the comma, yes the way the casting is done, binary tables yield an error. So better leave them as is. Notice, that this problem is only occuring when you try to save an intermediate product like in e.g. l3_to_l4_spec after add_wavelength_map, the POSLOOKUP table yields an error, but this is deleted before saving to l4 product.

semaphoreP (approved):

Issue #951, opened 2026-06-29
Exclude central region (interior to IWA) from Cosmic ray flagging


jrhom1:
**Is your feature request related to a problem? Please describe.**
Currently, the entire EXCAM image is considered for flagging and masking cosmic rays and saturated pixels. However, after discussions in in both the Exozodi and reflected light telecons, there has been a desire to not flag any pixels in the central region (spot of arago, HOWFSC artifacts) interior to the IWA of a particular mode as this region sometimes gets flagged as a cosmic ray (particularly in reference star/bright star images). Because cosmic ray masks cannot be fixed in l3 to l4 with replace_bad_pixels(), PSF subtraction relying purely on RDI will fail because of the cosmic ray nan pixels in reference star images. Information from this region is most likely not useful for science anyways, so saturation can be tolerated. The only thing that has not been checked however is whether or not in lower contrast scenarios if the central region will saturate enough to streak across the detector.

**Proposed Solution**
For each observing mode (NFOV, WFOV, B1 and B4, integrated and pol mode), define the region interior to the IWA that does not get flagged for saturation/cosmic rays in L1 to L2a processing.

**Additional context**
Examples of the central region being flagged as a cosmic ray

<img width="1400" height="600" alt="Image" src="https://github.com/user-attachments/assets/3628068c-6731-4aa9-905d-745fd4c2c9d3" />

maxwellmb:
A potential issue is if there actually _is_ a cosmic ray there, then we'd Want it masked and the tail corrected. The tail could affect real regions.

jrhom1:
But unless you look at the frame yourself, it doesn't seem like there's any possible automated way to differentiate between the two cases, which has also been an issue in general (differentiating between cosmic rays and saturated pixels)

jrhom1:
Another possible solution would be to have replace_bad_pixels in L3 to L4 actually be able to replace the nans in cosmic ray masks, but this also seems like a hard ask/maybe not possible

maxwellmb:
Is this still a problem with the combine step in L3 to L4?

jrhom1:
I have not tested that personally, maybe @zjwhale has with her simulations? But I think it's worthwhile to be more robust regardless and not dependent on the combine step resolving that issue. @semaphoreP could you confirm whether or not there is a way to get replace_bad_pixels to fix the bad pixels under cosmic ray masks?

maxwellmb:
[...]

PR #993, opened 2026-08-06
Exclude iwa from cosmic ray flagging

everlastingEric:
## Describe your changes
- Added a check to exclude the coronagraph inner working angle from getting flagged for cosmic rays. The iwa location and radius is defined for NFOV and WFOV modes, as well as imaging and pol modes. 
- Added a unit test in `test_cr_detection.py` to verify functionality

## Type of change

#951 


semaphoreP:
@everlastingEric can you push a trivial commit to relaunch the tests?

corgidrp/l1_to_l2a.py
```diff
@@ -210,6 +212,19 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
             the input and output datasets will be identical.  This is useful when handling a large dataset and when the input dataset is not needed afterwards. Defaults to True.
         discard_oversat (bool): if True, discard frames that exceed pct_oversat_lim, preserving the previous behavior.
             If False, keep them, mark IS_BAD, and skip cosmic ray identification for those frames. Defaults to False.
+        centerxy (array): 
+            Length 2 array containing the [x, y] coordinates of the image enter, 
+            for use in bypassing the coronagraph IWA from cosmic ray masking.
```
semaphoreP: I think we should use EACQ_ROW and EACQ_COL header keywords to figure out the position of the coronagraph. This is the goal location of the low order wfs, and should be the center of the FPM.

corgidrp/l1_to_l2a.py
```diff
@@ -218,6 +233,9 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
     sat_dqval = 32 # DQ value corresponding to full well saturation
     cr_dqval = 128 # DQ value corresponding to CR hit
 
+    platescale = 0.0218 # detector platescale in arcseconds/pixel
```
semaphoreP: this should be an optional keyword since this is just a good guess of the platescale.

corgidrp/l1_to_l2a.py
```diff
@@ -300,13 +318,66 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
+        iwa_mask = np.ones_like(curr_frame.data)
+        # check the coronagraph configuration used to determine IWA
+        cor_mode = curr_frame.ext_hdr['LSAMNAME']
+        # ensures this computation only happens if the coronagraph is in
+        if cor_mode != 'OPEN':
+            iwa = 0 # units of lambda/d
+            if cor_mode == 'NFOV':
+                iwa = 3
+            elif cor_mode == 'WFOV':
+                iwa = 5.9
```
semaphoreP: Based on https://arxiv.org/pdf/2508.08063, we should also check we're conversing spectroscopy mode, which should also have a 3.0 lam/D IWA.

semaphoreP: actually, let's not worry about figuring out how to mask cosmic rays for spectroscopy, but at least mention in the docstrting we only ignore the mask for imaging/pol mode.

corgidrp/l1_to_l2a.py
```diff
@@ -300,13 +318,66 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
+                r_ord = np.sqrt((y - centerxy_ord[1])**2 + (x - centerxy_ord[0])**2)
+                iwa_region_ord = r_ord < iwa_pix
+                r_ext = np.sqrt((y - centerxy_ext[1])**2 + (x - centerxy_ext[0])**2)
+                iwa_region_ext = r_ext < iwa_pix
+                # mask out the defined regions
+                iwa_mask[iwa_region_ord] = 0
+                iwa_mask[iwa_region_ext] = 0
+            
+        # apply mask to the data used for cosmic ray flagging
+        flag_cr_input = crmasked_cube[i:i+1,:,:] * iwa_mask
```
semaphoreP, line 395: ok, annoyingly, I don't think we can just do this, because a bunch of our default recipes have `dataset_copy=False` to save memory, and thus, this would actually set the data moving forward to 0's, which is lossy and not desirable.

I'm not sure if there's a better way than modifying the flag_cosmics code to pass in a 2-D mask to tell it where to ignore detecting cosmic rays in the IWA during the big for loop.

corgidrp/l1_to_l2a.py
```diff
@@ -161,7 +162,8 @@ def prescan_biassub(input_dataset, noise_maps=None, return_full_frame=False,
 def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh=0.95,
                        plat_thresh=0.85, cosm_filter=2, cosm_box=3, cosm_tail=10,
                        mode='image', detector_regions=None, pct_oversat_lim=20,
-                       dataset_copy=True, discard_oversat=False):
+                       dataset_copy=True, discard_oversat=False, centerxy=[512, 512],
```
semaphoreP: can we have a keyword about whether or not to skip the IWA region (default skip)?

semaphoreP (commented):
Some comments. Main thing is that we should avoid corrupting the original data, because this is one of the few step functions where copy_dataset=False by default typically (to save RAM).

It'd also be good for @maxwellmb to look at the pol implementation.

corgidrp/l1_to_l2a.py
semaphoreP: Actually, I think I'm wrong here, and this line creates a copy of the array. However, we may want to avoid doubling the memory footprint of this step function.

everlastingEric: I'm looking at it, because this line lives inside the for loop that goes over each individual 2D frame in the datacube, shouldn't it only create one extra 2D frame that gets written over each time the loop updates instead of a copy of the entire datacube? I can still move the masking to inside `flag_cosmics` to be safe but I wanted to first double-check if this really makes that big of a difference in terms of memory footprint.

semaphoreP: ah you're right. it does live insight the for loop. I somehow missed that i:i+1 indexing. Ok, this should be fine then.

everlastingEric:
addressed the comments, tests should be running again

corgidrp/l1_to_l2a.py
```diff
@@ -300,14 +317,77 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
+        if cor_mode != 'OPEN' and skip_coronagraph_iwa:
+            # determine where the coronagraph center is
+            if eacq_row is not None and eacq_col is not None and eacq_row != 0 and eacq_col != 0:
+                # use EACQ_ROW and EACQ_COL as the coronagraph center if these headers are set properly
+                centerxy = [eacq_col, eacq_row]
+            else:
+                # defaults to (512, 512) if the headers are not initialized properly
+                centerxy = [512, 512]
+            # initialize the iwa
+            iwa = 0 # units of lambda/d
```
maxwellmb: Can we make iwa a kwarg that defaults to None? If it's set then it overwrites the defaults in this function.

maxwellmb (changes requested):
the pol stuff looks fine to me. I have a slight request about how the iwa is defined though.

corgidrp/l1_to_l2a.py
everlastingEric: ok is the idea that the iwa could be an adjustable parameter that the user can fine tune? The only thing I want to check is that if the iwa defaults to none it overrides the other keyword I had that tells the step function whether to exclude the iwa or not, which is set to true by default.

maxwellmb: > ok is the idea that the iwa could be an adjustable parameter that the user can fine tune? 

Yes. 

Ok, that sounds like a fine thing to override, but I'd suggest adding a warning if the other keyword isn't set.

everlastingEric: ok I have updated this. The unit tests are failing for the same unrelated reason that is causing the tests in my other PR to fail (calspec connection issues)

corgidrp/l1_to_l2a.py
```diff
@@ -300,14 +320,75 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
-        m2[i,:,:] = flag_cosmics(cube=crmasked_cube[i:i+1,:,:],
-                        fwc=sat_fwcs[i]/sat_thresh, #sat_fwcs are already multiplied by sat_thresh, so undo that since this function multiplies sat_thresh as well
+
+        # first determine coronagraph iwa to exclude from cosmic ray finding (if applicable)
+        # by default iwa_mask transmits everything, only in applicable conditions will the iwa be zeroed out
+        iwa_mask = np.ones_like(curr_frame.data)
+        # check the coronagraph configuration used to determine IWA
+        cor_mode = curr_frame.ext_hdr['LSAMNAME']
+        # ensures this computation only happens if the coronagraph is in and we explicitly want to skip masking
+        if cor_mode != 'OPEN' and skip_coronagraph_iwa and (coronagraph_iwa_radius is not None):
```
maxwellmb: if coronagraph_iwa_radius is None, we want the the code to default to the values you had for NFOV and WFOV before. It looks like that is is gone now and it just doesn't do anything if it is None.

maxwellmb (changes requested):

corgidrp/l1_to_l2a.py
everlastingEric: oh whoops I misunderstood, that's back in now

corgidrp/l1_to_l2a.py
```diff
@@ -300,14 +320,83 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
...
+            # convert the iwa from lambda/d to pixel units
+            filter_band = curr_frame.ext_hdr['CFAMNAME']
+            # set default values for the iwa if this is not set
+            if coronagraph_iwa_radius is None:
+                if cor_mode == 'NFOV':
+                    coronagraph_iwa_radius = 3
+                elif cor_mode == 'WFOV':
+                    coronagraph_iwa_radius = 5.9
+                else:
+                    coronagraph_iwa_radius = 0
```
maxwellmb, line 357: Almost got it, but right now if it's not none you're setting it to 0. If it's not None, we want to just use the value provided.

maxwellmb (changes requested):

corgidrp/l1_to_l2a.py
everlastingEric: unless I'm misunderstanding something, this whole block is nested inside the if statement that checks if the iwa keyword is none first, so if it's not none this line shouldn't fire. I put it in to guard against cases like spec mode where we haven't necessarily defined what to do yet and changing the iwa from none to 0 should just let the code ignore it instead of raising an error.

maxwellmb: You're correct. Sorry. Long day.

maxwellmb (approved):
things look good on my end. I'm not sure if @semaphoreP has anything outstanding.

semaphoreP (approved):
Looks good to me too now.

PR #1000, opened 2026-08-11
Update boresight calibration data splitting and combining

### LINKED ISSUE #997: boresight should use RA/Dec to distinguish between dither positions
**Describe the bug**
It turns out each dither position is a telescope move that changes RA/Dec. FSM X/Y is a possibility, but not the primary way we'll dither the field. We should update the boresight code to adapt to checking RA/Dec in addition to FSM X/Y to split between dither positions. 

**corgidrp version:**
 - develop

everlastingEric:
## Describe your changes
- Added back the option to combine frames based on grouping a specified number of frames together instead of grouping based on headers in boresight calibration.
- Added the option to group frames based on RA/Dec dither, and additional user-specified header keywords.

## Type of change

#997 


corgidrp/astrom.py
```diff
@@ -995,12 +996,14 @@ def boresight_calibration(input_dataset, field_path='JWST_CALFIELD2020.csv', fie
         platescale_guess (float): An initial guess for the platescale value (default: 21.8 [mas/ pixel])
         platescale_tol (float): A tolerance for finding source matches within a fraction of the initial plate scale guess (default: 0.1)
         center_radius (float): Percent of the image to compute plate scale and north angle from, centered around the image center (default: 0.9 -- ie: 90% of the image is used)
-        combine_frames (boolean): Determine whether to median combine all frames of the same target and instrument configuration into a singular frame (default: False)
+        combine_frames (boolean or int): If True or 1, median combine all frames of the same target and instrument configuration into a singular frame.
```
semaphoreP: can we actually revert this naming scheme to the old `frames_to_combine`. We are using this keyword right now for the CAR processing using v5.0 of the DRP, and it'll be mildly annoying to introduce this breaking change in v5.1. The naming is slightly sub-optimal, but allows us to keep backwards compatibility.

semaphoreP (commented):
Thanks for the quick response on this. Upon some reflection, I'd like to go back to the old keyword name, for backwards compatibility purposes.

corgidrp/astrom.py
everlastingEric: ok the name should be changed back now

semaphoreP (approved):

Issue #1008, opened 2026-08-15
Update headers (SATSPOTS, RA_BORE, DEC_BORE, RA_APER, DEC_APER)


adrienmaillard:
**Describe the bug**

One header is of the wrong type and 4 headers are missing in the simulated images. 
Reference for L1 headers: 
https://collaboration.ipac.caltech.edu/spaces/romancoronagraph/pages/161617086/L1+Current+DRP+Development+Version

1) SATSPOTS

I was working on the DIP's categorization.xml and I saw discrepancies between how SATSPOTS header values are described in various pages. In some, it's boolean, in some other it's int. The reference confirms that SATSPOTS should be a boolean.

When I open recently simulated images (July, August) for CAR purposes (produced with mocks.py, right?), I see that SATSPOTS still equals to 0 or 1. https://media.githubusercontent.com/media/roman-corgi/corgidrp/refs/heads/develop/corgidrp/data/header_formats/l1.csv confirms SATSPOTS is still an int in drp. 

2) RA_BORE, DEC_BORE, RA_APER, DEC_APER
These 4 headers have been recently added in the reference. 

**Expected behavior**
SATSPOTS should be a boolean value and read as T/F in the header, not 0/1, in the simulated images.
RA_BORE, DEC_BORE, RA_APER, DEC_APER should be present in the headers of simulated images.

**corgidrp version:**
 - current

nasavbailey:
Also need to look for knock-on effects of the SATSPOTS data type fix.  For example, l3_to_l4.py, maybe others.

semaphoreP:
Updated in PR #1009

PR #1009, opened 2026-08-18
Update L1 Headers

everlastingEric:
## Describe your changes
- Added the primary header keywords `RA_BORE`, `DEC_BORE`, `RA_APER`, and `DEC_APER` keywords, and changed the extension header keyword `SATSPOTS` to be a boolean value in `mocks.create_default_L1_headers()`  through updating the CSV file containing information on the L1 header keywords.
- This change should allow for all newly generated corgisim and mock data to have the correct most up to date header keywords, as all header creation from both corgisim and `corgidrp.mocks` traces back to the `create_default_L1_headers()` function.
- Note: Changing the `SATSPOTS` keyword from type int to bool shouldn't affect any existing pipeline step functions since 0/1 and T/F are interchangeable in python.

## Type of change

#1008 


everlastingEric:
The unit tests are failing, but for reasons that I'm pretty sure are not due to my changes. They all have this error message: 
```
address = ('archive.stsci.edu', 443)
timeout = <object object at 0x7f8a5b5209c0>, source_address = None
all_errors = False
    def create_connection(address, timeout=_GLOBAL_DEFAULT_TIMEOUT,
                          source_address=None, *, all_errors=False):
        """Connect to *address* and return the socket object.
    
        Convenience function.  Connect to *address* (a 2-tuple ``(host,
        port)``) and return the socket object.  Passing the optional
        *timeout* parameter will set the timeout on the socket instance
        before attempting to connect.  If no *timeout* is supplied, the
        global default timeout setting returned by :func:`getdefaulttimeout`
        is used.  If *source_address* is set it must be a tuple of (host, port)
        for the socket to bind as a source address before making the connection.
        A host of '' or port 0 tells the OS to use the default. When a connection
        cannot be created, raises the last error if *all_errors* is False,
        and an ExceptionGroup of all errors if *all_errors* is True.
        """
    
        host, port = address
        exceptions = []
        for res in getaddrinfo(host, port, 0, SOCK_STREAM):
            af, socktype, proto, canonname, sa = res
            sock = None
            try:
                sock = socket(af, socktype, proto)
                if timeout is not _GLOBAL_DEFAULT_TIMEOUT:
                    sock.settimeout(timeout)
                if source_address:
                    sock.bind(source_address)
>               sock.connect(sa)
E               TimeoutError: [Errno 110] Connection timed out
```

semaphoreP:
Eric, I think all the data format doc pages will need updating due to the existence of new headers, and data type changes of old headers.

everlastingEric:
Yeah, I can go through them and make the updates, but unless I generate new data with the updated mocks function (or hack the input data to have these new headers), the data formats test will fail. What do you think is the best way to approach this?

semaphoreP:
Hmm good point. Probably the right way to do this is to indeed add some code in e2e test to inject these header changes to the existing test data. To save you a lot of work, we could modify the `__init__` of the Image() class to initialize these four header keywords and make sure SATSPOTS is be bool. I think that's relatively harmless, although does give us a small chunk of extra code to maintain.

everlastingEric:
ok the data format docs should be updated now

corgidrp/data.py
```diff
@@ -550,6 +550,18 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq
...
 
+        # insert/update header keywords that might not be there in outdated data
+        if not 'RA_BORE' in self.pri_hdr:
+            self.pri_hdr.set('RA_BORE', 0, 'Boresight RA')
+        if not 'DEC_BORE' in self.pri_hdr:
+            self.pri_hdr.set('DEC_BORE', 0, 'Boresight Dec')
+        if not 'RA_APER' in self.pri_hdr:
+            self.pri_hdr.set('RA_APER', 0, 'CGI Aperture RA')
+        if not 'DEC_APER' in self.pri_hdr:
+            self.pri_hdr.set('DEC_APER', 0, 'CGI Aperture Dec')
```
semaphoreP: can we set these to be equal to the RA and Dec header values instead of 0?

corgidrp/data.py
```diff
@@ -550,6 +550,18 @@ def __init__(self, data_or_filepath, pri_hdr=None, ext_hdr=None, err = None, dq
         self.pri_hdr = adjusted_pri_hdr
         self.ext_hdr = adjusted_img_hdr
 
+        # insert/update header keywords that might not be there in outdated data
```
semaphoreP, line 553: make a note that this should be just apply to simulated data with outdated headers.

semaphoreP (commented):
A couple of minor comments, and one e2e test is failing due to the header change:

```
FAILED tests/e2e_tests/trap_pump_cal_e2e.py::test_trap_pump_cal - ValueError: Header keyword mismatches found. See details above.
```
Can you investigate?

everlastingEric:
hmm I just ran the same test and it passes on my end, do you maybe have a longer traceback that I can look at?

semaphoreP:
this was my mistake. I just needed to clear out the output directory before running it.

semaphoreP:
so just a couple of minor comments, and then we can pull this in.

everlastingEric:
sounds good, I caught an issue with a test where a dictionary instead of a header object was passed into the image constructor which caused the .set() calls I added to fail, so I'm just going to go through the tests again to make the correct datatype is passed in for everything.

everlastingEric:
comments are addressed

semaphoreP (approved):
Looks good. Only test failure is unrelated.

PR #1012, opened 2026-08-19
Make create_onsky_flatfield work for star raster flats (CAR-120)

### LINKED ISSUE #1011: create_onsky_flatfield in flat.py does not work for star raster flats - only for Neptune/Uranus flats
corgidrp.flat.create_onsky_flatfield (and combine_flatfield_rasters) was written and tuned only for making planet flats (Neptune/Uranus). When run on a dithered point-source star raster (CAR-120), it either crashes outright or produces a flat without physically reasonable data (median ~10 at the disk center instead of ~1.0, with lots of NaNs). The root cause is a set of assumptions hard coded in for planet flats.

juliamilton:
## Describe your changes

create_onsky_flatfield was built for planet flat fielding. On a dithered point-source star raster it either crashed or produced a a flat without physically reasonable values. This PR generalizes it to use the star raster conops as well while keeping the planet flat fielding behavior.

Issues: 
The flat is computed as each_frame / median(all_frames): this that only works if every cropped source is co-registered. For star raster flat fielding a single exposure is the star rastered across a slice of a circle - see picture 1 below. On the star data three assumptions in the code didn't work:

- Centering - Hot pixels/background were causing the centering to be skewed, mis-cropping each stamp. The median then averages misaligned sources into a washed-out smear, so dividing by it yields nonsense.
- No point-source radius - planet_rad was only set for Neptune/Uranus. A star left it None, causing a crash.
- Hardcoded constants. The fill radius and n_pix were hard coded numbers numbers tuned to planet size.

Pictures 2 and 3 show results of flat fielding algorithm before and after these changes, respectively.

#1011 

image

image

image


sfr348:
I made a branch that incorporates the thing I did which I think will make the FPM-in cases a bit more robust. 
https://github.com/roman-corgi/corgidrp/tree/flat-fix-disk-kernel

dither01_center
dither02_center
dither03_center
dither04_center
dither05_center


I don't have a super strong opinion either way but let me know what you think.

corgidrp/flat.py
```diff
@@ -314,7 +326,22 @@ def create_onsky_flatfield(dataset, planet=None,band=None,up_radius=55,im_size=N
...
-    
+        else:
+            # for a dithered point source, auto set the fill radius to 
+            # (spacing/sqrt(2)). large enough to close the diagonal gaps
+            # between the dither spots, but not larger than the crop
+            src_cents = []
+            for j in range(len(dataset)):
+                sm = median_filter(np.nan_to_num(dataset[j].data), 3)
+                m = (sm >= 0.3 * np.nanmax(sm)).astype(float)
+                c = np.array(centr.centroid_com(m)); c[np.isnan(c)] = 0
```
semaphoreP: I would've thought centroiding just on the median filtered image `sm` would be better than centroiding on this thresholded image `m`, but whatever works? I also noticed in my own experience Claude likes to blame bad pixels on things not working, even when it clearly isn't the case.

semaphoreP: Also, can you place separate commands on separate lines? Multiple commands on a single line makes it easy to miss one.

corgidrp/flat.py
```diff
@@ -314,7 +326,22 @@ def create_onsky_flatfield(dataset, planet=None,band=None,up_radius=55,im_size=N
...
+                src_cents.append(c)
+            src_cents = np.array(src_cents)
+            d2 = ((src_cents[:, None, :] - src_cents[None, :, :])**2).sum(-1)
+            np.fill_diagonal(d2, np.inf)
+            spacing = np.median(np.sqrt(d2.min(axis=1)))
+            planet_rad = min(int(np.ceil(spacing / np.sqrt(2))) + 1, up_radius - 1)
```
semaphoreP: why is this spacing/sqrt(2)? I would have thought spacing is the distance between dithers, so radius is spacing/2.

semaphoreP (commented):
I just have a couple small questions to make sure I understand the changes. Otherwise, the e2e tests pass so it's not breaking existing features, so no big issues as long as it works for you. I'll leave it up to you and Susan to figure out how to merge your two changes.

corgidrp/flat.py
juliamilton: Changed this so instead of averaging, it fits a circle. Code should now calculate how big the lit area should be (instead of hard-coded), and then it slides a circle of that size around and find where it overlaps the most lit pixels. Also kept the brightness-weighted average as a backup for when the circle fit fails.

corgidrp/flat.py
juliamilton: This is gone now - intent is to measure the radius directly from the illuminated area.

juliamilton:
I updated the code so that it uses @sfr348's centroiding method which fits a circle and slides it around to find where it overlaps the most illuminated pixels. A few extra updates to that code include: using an fft version instead, since ndimage.correlate was slow on a full frame, and also not hard-coding planet radius.

corgidrp/flat.py
```diff
@@ -327,45 +437,44 @@ def create_onsky_flatfield(dataset, planet=None,band=None,up_radius=55,im_size=N
...
+            centroid = source_centroid(planet_image)
+        if centroid is None:
+            raise ValueError("Could not locate the source in frame {0} of the "
+                             "dataset.".format(j))
+        centroid = np.asarray(centroid, dtype=float)
         act_cents.append((centroid[1],centroid[0]))
-        xc =int( centroid[0])
-        yc = int(centroid[1])
+        xc = int(round(centroid[0]))
+        yc = int(round(centroid[1]))
```
semaphoreP: something along these lines of code is now causing the flat_field_e2e test to fail.

```
FAILED tests/e2e_tests/flatfield_e2e.py::test_flat_creation_neptune - assert np.float64(0.015053237400837523) < 0.0071
FAILED tests/e2e_tests/flatfield_e2e.py::test_flat_creation_uranus - assert np.float64(0.03610727712538253) < 0.0071
```

I think we were barely under the assert limit before, and now this creeps it up. Maybe we could have an option for specifying which centroid option to use, and for Neptune/Uranus we default to the old one, and for stars, we default to the new one?

semaphoreP (commented):
Looks like this caused the e2e test to fail now, so we should be careful about modifying code that affects the planet cases, unless we think this should definitely be a strict upgrade.

corgidrp/flat.py
juliamilton: Went back to the old method for Neptune/Uranus and the flatfield_e2e test is passing again locally

semaphoreP (approved):

Issue #1010, opened 2026-08-19
median filter mode for cosmic ray flagging


kjl0025:
**Is your feature request related to a problem? Please describe.**
Tail length of cosmic rays varies with EM gain.  Flagging is not very efficient without tweaking per frame, and the risk of over-flagging is high.

**Describe the solution you'd like**
Use a median filter to make a histogram of (frame - filtered frame) to select a threshold above the background level on a per-frame basis to efficiently flag cosmic rays.  

**Additional context**
This methodology is used for CAR119 walker test.

semaphoreP:
Added in PR #1017

Issue #999, opened 2026-08-10
pre-scan subtraction tainted by rare long-tail cosmic rays


kjl0025:
**Describe the bug**
Occasionally, a cosmic ray from a high-energy impact can cause a cosmic tail that spills over into the next row's pre-scan region, which means bias subtraction using the median of the "good columns" for that next row is not reliable.  This can cause the row in the image area to have severely negative counts.  

**To Reproduce**
This arises somewhat rarely when simulating a high-gain high-exposure time frame.

**Expected behavior**
The bias-subtracted image area should not have pixels that are very negative.  

**corgidrp version:**
v5.0

**Additional context**
Noticed this bug in noise map calibration frames.

semaphoreP:
Fixed in PR #1017

PR #1017, opened 2026-08-28
Cosmic ray median filter, prescan cosmic ray streak handling

kjl0025:
## Describe your changes
The way that weighting the least squares fit in calibrate_darks_lsq() has been changed so that only the number of frames available per gain-exptime combination is considered and not what comes from ERR, which may not be fully reliable.  Also, for a given gain-exptime combination (when CR_threshold_check is True), if the cosmic ray threshold recorded in HISTORY during detect_cosmic_rays() is not high enough above the expected mean count level, that combination is not used in the least squares fit for making noise maps to avoid artificial truncation of the noise count distribution.  Also, fixed a minor error in output DQ map for noise maps and made sure pixels that couldn't be fit given the inputs are marked as 0 on the output noise maps.  (The chances of such unfittable pixels are very, very low.)

New knobs to turn for handling cosmic rays.  If median_filter_mode = 0 and cosm_thresh=sat_thresh (which are the defaults), the usual behavior for detect_cosmic_rays().  If 1, a median filter is utilized to make a histogram and select an appropriate threshold for each frame, and cosmic "tail trails" (pixels overspilling from the saturated cosmic ray) are masked according to the input cosm_tail.  If 2, the median filter method is used (which catches virtually all high-flux cosmic rays), and then the usual method is used on the frame's pixels which were not masked by the median filter method (which catches low-flux cosmic rays if cosm_thresh is low enough).  For science frames, the flux levels of relevant images may be similar to that of low-flux cosmic rays, and 1 is recommended for the input in that case.  

Also, cosmic rays the spill over into the next row's prescan results in a large deficit in count value for that row after bias subtraction.  Now, that has been fixed by subtracting instead the median value over the rows for any row standing out enough from the rest.

This work mainly had in view getting CAR119 to meet error budget constraints.

## Type of change

#1010 
#999 


corgidrp/darks.py
```diff
@@ -306,7 +306,7 @@ def build_trad_dark(dataset, detector_params, detector_regions=None, full_frame=
             dq_temp = Image(dataset[j].filepath).dq
             dq_sum += dq_temp.astype(float)
         dq_sum = np.ma.masked_array(dq_sum, dq_sum == 0)
-        output_dq = 2**((np.ma.log(dq_sum)/np.log(2)).astype(int)) - 1
```
semaphoreP: was this -1 a bug before?

semaphoreP:
@kjl0025 can you also describe in the other changes in the PR, most importantly the changes to the calibrate_darks_lsqr fucntion. Seems somewhat nontrivial.

corgidrp/l1_to_l2a.py
```diff
@@ -227,6 +244,20 @@ def detect_cosmic_rays(input_dataset, detector_params, k_gain = None, sat_thresh
             for use in bypassing the coronagraph IWA from cosmic ray masking when observing in pol mode. Defaults to 45. 
         coronagraph_iwa_radius (float): The radius of the coronagraph IWA in units of lambda/d, for use in excluding the coronagraph
             IWA from cosmic ray flagging. Defaults to None.
+        median_filter_mode (int): If 0, the "usual" method (threshold based on cosm_box, cosm_tail, etc) is employed.  If 1, a median filter is utilized to make a histogram and select an
```
semaphoreP, line 247: what's the case where we want to use case 1? it's just a less aggressive version of case 2?

corgidrp/darks.py
```diff
@@ -553,16 +572,45 @@ def calibrate_darks_lsq(dataset, detector_params, weighting=True, detector_regio
...
+            end_ind = clean_hist_str[ind:].find('used.') #finds next instance of this, which immediately follows the number
+            if ind != -1 and end_ind != -1:
+                cosm_thresh_used_dn = float(clean_hist_str[ind+24 : ind+end_ind])
+        if cosm_thresh_used_dn is None: #if not available from HISTORY, use SAT_DN 
+            cosm_thresh_used_dn = datasets[i][0].ext_hdr['SAT_DN']
+        cosmic_thresh_used_e = cosm_thresh_used_dn * datasets[i][0].ext_hdr['KGAINPAR']
+        CR_thresholds_e = np.append(CR_thresholds_e, cosmic_thresh_used_e)
+        threshold = expected_mean + num_stds * expected_std
+        if CR_thresholds_e[i] <= threshold and CR_threshold_check:
+            continue #skips over this exptime-gain combination
```
semaphoreP: should we wrap this whole thing in `if CR_threshold_check:` just to keep the code a bit cleaner? I'm not sure any of this is used anywhere else beyond the threshold check.

corgidrp/darks.py
```diff
@@ -637,13 +685,19 @@ def calibrate_darks_lsq(dataset, detector_params, weighting=True, detector_regio
         # now pick a pixel from rows_normal and cols_normal to use as a reference for the approximated error for the pixels that have 1 unmasked frame, undo the division by sqrt(unmasked_num), and divide by 1
         stat_std[rows_one, cols_one] = stat_std[rows_normal[0], cols_normal[0]] * np.sqrt(unmasked_num.max())/1
         total_err = np.sqrt(mean_err**2 + stat_std**2)
-        pixel_mask = (unmasked_num < len(datasets[i].frames)/2).astype(int)
+        reliable_fraction = 0.75 # not used in output product, so hard-coded number okay (for debugging)
```
semaphoreP, line 686: it looks like this gets used down the line to store some arrays of unreliable_pix_masks. Are those used for debugging too? Do we need to keep them if they are stacks of pixel maps (any RAM concerns?)

corgidrp/darks.py
```diff
@@ -654,16 +708,21 @@ def calibrate_darks_lsq(dataset, detector_params, weighting=True, detector_regio
...
         output_dqs.append(output_dq)
     output_dqs = np.stack(output_dqs)
     unreliable_pix_map = unreliable_pix_map.astype(int)
+    unreliable_pix_masks = np.stack(unreliable_pix_masks)
     mean_stack = np.stack(mean_frames)
     mean_err_stack = np.stack(total_errs)
+    mean_stat_err_stack = np.stack(stat_errs)
+    weights = np.stack(weights).astype(float)
+    #much smaller than the weight due to 1 unmasked frame but not 0, to avoid singular matrix
+    weights[np.where(weights==0)] = (1/len(dataset.frames))/100 #if normalized per sub-stack: (1/len(datasets[i].frames))/100
```
semaphoreP, line 720: just to clarify, this is just a hard coded really small number for the cases where unmasked_num=0?

tests/test_calibrate_darks_lsq.py
```diff
@@ -280,39 +280,21 @@ def test_weighting():
...
-    # But the result is still a number closer to the correct value compared to the case where no appropriate weighting is used:
+    # the result is a number closer to the correct value compared to the case where no appropriate weighting is used:
     assert(np.abs(np.mean(CIC_image_map) - np.mean(wrong_CIC_image_map_dq)) < 
            np.abs(np.mean(CIC_image_map) - np.mean(wrong_CIC_image_map)))
-    # and err should be reduced overall now, despite effectively having fewer stacks for fitting
-    assert(np.nanmean(wrong_noise_maps_dq.CIC_err) < np.nanmean(wrong_noise_maps.CIC_err))
-    # Finally, demonstrate weighting through fewer frames in the erroneous sub-stacks:
+    # and err different now that very few frames determined one of the data points
+    assert(np.nanmean(wrong_noise_maps_dq.CIC_err) != np.nanmean(wrong_noise_maps.CIC_err))
+    # demonstrate weighting through fewer frames in the erroneous sub-stacks:
```
semaphoreP, line 297: I'm a bit confused here, shouldn't the CIC_err get bigger when we have fewer stacks for fitting? If so, should we be checking for that?

tests/test_calibrate_darks_lsq.py
```diff
@@ -330,7 +312,7 @@ def test_weighting():
     # similar logic that applied for masking via DQ above
     assert(np.abs(np.mean(CIC_image_map) - np.mean(smaller_CIC_image_map)) < 
            np.abs(np.mean(CIC_image_map) - np.mean(wrong_CIC_image_map)))
-    assert(np.nanmean(smaller_noise_maps.CIC_err) < np.nanmean(wrong_noise_maps.CIC_err))
+    assert(np.nanmean(smaller_noise_maps.CIC_err) != np.nanmean(wrong_noise_maps.CIC_err))
```
semaphoreP: same question here about why we can't test on the sign of the err changing?

semaphoreP (commented):
this generally looks good given it's not a breaking change and it fixes some CAR issues. I had mainly some questions so I can better understand what's going on.

corgidrp/l1_to_l2a.py
kjl0025: Case 1 is useful for science images (in theory) in which cosmic rays standing out above the background level of (frame - median-filtered frame) are masked very precisely, without masking extra pixels with cosm_box, for example.  Case 2 goes beyond that and is useful for finding low-flux non-saturating cosmic rays, which is important for noisemap/dark calibration but probably detrimental to science images.

corgidrp/darks.py
kjl0025: Yes, I agree. Just pushed that change

corgidrp/darks.py
kjl0025: This is a stack of ~30 frames (just the number of gain-exptime combinations).  But it is simply for debugging, so I commented it out.

corgidrp/darks.py
kjl0025: That's correct.

tests/test_calibrate_darks_lsq.py
kjl0025: You are right; I changed it to > .

tests/test_calibrate_darks_lsq.py
kjl0025: Changed.

kjl0025:
Also, I just realized I wasn't computing the overall DQ output correctly for the RAM-heavy cases in photon counting and darks.py.  I pushed a fix for those instances just now as well.

semaphoreP (approved):

Issue #1014, opened 2026-08-20
ND Filter should output a Abs Flux Cal if faint star data is passed in


semaphoreP:
**Is your feature request related to a problem? Please describe.**
When faint star data is passed into the ND Filter Calibration step, it already produces abs flux calibration products in the code. It should output one. 

**Describe the solution you'd like**
Average all the abs flux calibration product values together and output the averaged abs flux calibration product. The header should note which files and which star names were averaged together. Return it alongside the ND filter calibration as a two Image `Dataset` instead of a single ND Filter Calibration.

semaphoreP:
@everlastingEric if you're doing this, could you also modify the walker.py selection logic to pick between abs flux and ND filter recipes based on the visittype (bright vs faint star) rather than on fsmx/y?

everlastingEric:
@semaphoreP Sure, just to double check on what the distinguishing factor is, do we want abs flux when only a faint star dataset is passed in?

semaphoreP:
Yup, Abs flux for faint star visits, and ND filter for bright star visits. Tagging @nasavbailey for awareness (I think she originally requested this change).

everlastingEric:
@semaphoreP I want to double check one thing since it requires me to make a few minor changes to walker.py. I'm not actually able to create a dataset consisting of abs flux cal and ND filter cal data since these two calibrations have different data shapes. So instead, I'm just returning them as a list and that requires me to add in a few lines to the walker code to support saving data in this format. Is this fine?

semaphoreP:
@everlastingEric ah dang. could we add a knob to disable the creation of the dataset.all_data/all_dq/add_err to allow heterogenous datasets to be made? or is that not the issue?

everlastingEric:
Yeah that would fix the issue. I'm a little wary that this might break other things but if I keep the default behavior to having all_data be created and only disable it for very specific cases like this it should be fine I think.

semaphoreP:
Added in #1018

PR #1018, opened 2026-08-28
Update nd filter cal

everlastingEric:
## Describe your changes
- ND filter calibration now returns a size 2 dataset that also includes the abs flux calibration created from the dim dataset
- walker distinguishes between abs flux and ND filter calibrations using bright/dim datasets instead of fsmx/fsmy dithering.
- added a new argument to the dataset class constructor to disable the creation of all_data/all_err/all_dq to construct a dataset from frames that have mismatched dimensions (ND filter cal and abs flux cal).

## Type of change

#1014

### DIFF
--- a/corgidrp/nd_filter_calibration.py
+++ b/corgidrp/nd_filter_calibration.py
@@ -436,7 +436,8 @@ def create_nd_filter_cal(stars_dataset,
         gaussian_kernel_size (int): Size of gaussian kernel used to make the PSF centroid algorithm more robust against noise.
 
     Returns:
-        sweet_spot_dataset (corgidrp.Data.NDFilterSweetSpotDataset): ND Filter calibration product for the dataset given
+        sweet_spot_and_fluxcal_dataset (list): Length 2 list containing a ND Filter calibration product (corgidrp.Data.NDFilterSweetSpotDataset) 
+            for the dataset given and an absolute flux calibration product (corgidrp.Data.FluxcalFactor) used to create the ND filter calibration.
     """
     if phot_kwargs is None:
         phot_kwargs = {}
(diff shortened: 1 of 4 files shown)


semaphoreP:
@nasavbailey @juliamilton for awareness: we are moving the walker logic to use CGIVST_CAL_ABSFLUX_BRIGHT and CGIVST_CAL_ABSFLUX_FAINT to distinguish between ND Filter and Abs Flux calibrations. If both BRIGHT/FAINT are in the same dataset, we do ND filter, which now also outputs the abs flux calibration. Let me know if there's any issue with this.

corgidrp/nd_filter_calibration.py
```diff
@@ -436,7 +436,8 @@ def create_nd_filter_cal(stars_dataset,
         gaussian_kernel_size (int): Size of gaussian kernel used to make the PSF centroid algorithm more robust against noise.
 
     Returns:
-        sweet_spot_dataset (corgidrp.Data.NDFilterSweetSpotDataset): ND Filter calibration product for the dataset given
+        sweet_spot_and_fluxcal_dataset (Dataset): Length 2 dataset containing a ND Filter calibration product (corgidrp.Data.NDFilterSweetSpotDataset) 
+            for the dataset given and an absolute flux calibration product (corgidrp.Data.FluxcalFactor) used to create the ND filter calibration.
```
semaphoreP: I think we should only return a ND Filter calibration product if we did not create our own abs flux calibration file. If we did, then we should return the dataset of both.

The reasoning is that the output products will get saved by default, and we don't want to save + add to the caldb the abs flux calibration product if it isn't new (would create duplicates)

semaphoreP (commented):
Overall looks good. I just have one comment that I don't think we should always return a dataset.

corgidrp/nd_filter_calibration.py
everlastingEric: ok done

semaphoreP (approved):
Looks good now. Thanks!

Issue #1004, opened 2026-08-13
create_onsky_pol_flatfield crashes when frames contain cosmic rays


juliamilton:
**Summary**

corgidrp.flat.create_onsky_pol_flatfield fails to find the two polarization spots on realistic frames containing cosmic rays, raising the following error:

```
WARNING: NoDetectionsWarning: No sources were found. [photutils.detection.daofinder]
File ".../corgidrp/flat.py", line 460, in create_onsky_pol_flatfield
    xc=np.mean([pol1_x,pol2_x])
UnboundLocalError: cannot access local variable 'pol1_x' where it is not associated with a value
```

This is coming from `flat.py` where `threshold_value = np.max(planet_image)` finds a single-pixel cosmic ray brighter than the real spots. I believe there are additional issues with `DAOStarFinder` here as well.

**To reproduce**
Run the walker on CAR-154 L1 pol flat frames (can use car154_polflatfield_walker_test.py in the car154-polflat branch of corgi-car-tests repo)

kjl0025:
Just for cognizance, I have a DRP branch called cosmic_ray_med_filter which may solve this issue by dealing with the cosmic ray masking itself, assuming the polarization spots are found after masking for cosmic rays.

semaphoreP:
Fixed in PR #1021

Issue #1003, opened 2026-08-13
Polarimetric flat fielding walker path/ recipe assumes a planet (not star) raster


juliamilton:
**Summary**
The walker/ recipe path for polarimetric flat fields (VISTYPE=CGIVST_CAL_FLAT + DPAMNAME  POL0/POL45) cannot complete as written on CAR-154 data. corgidrp.flat.create_onsky_pol_flatfield is written for a planet (uranus/neptune) raster, but CAR-154 uses a star raster, and the recipe supplies no keyword arguments, so planet_rad (which is needed in the function) defaults to None.

**Steps to reproduce**
Run the walker on CAR-154 L1 pol flat frames (can use car154_polflatfield_walker_test.py in the car154-polflat branch of corgi-car-tests repo)

semaphoreP:
Added in PR #1021

PR #1021, opened 2026-08-31
Pol flat fixes to address #1003 and #1004

juliamilton:
## Describe your changes
Median filter the pol flat images and adapt the imaging mode flat fielding spot detection algorithms so that flat fielding works in polarimetry mode with a star raster. 

## Type of change

#1003 and #1004 

### DIFF
--- a/corgidrp/flat.py
+++ b/corgidrp/flat.py
@@ -561,9 +561,7 @@ def create_onsky_pol_flatfield(dataset, planet=None,band=None,up_radius=55,im_si
         elif planet.lower() == 'uranus':
              planet_rad = 65
         else:
-            # A rastered star fills a disk; measure it from the data rather than
-            # assuming a size or a raster pattern. Each pol component is a separate
-            # lit region, so the largest connected disk gives the per-component radius.
+            # a rastered star fills a disk of unknown size; measure it per frame
             radii = [measure_disk_radius(frame.data) for frame in dataset]
             radii = [r for r in radii if r is not None]
             if not radii:
@@ -578,11 +576,8 @@ def create_onsky_pol_flatfield(dataset, planet=None,band=None,up_radius=55,im_si
          elif band[0] == "4":
             rad_mask = 1.75
     if n_pix is None:
-        # Size the final flat to the full illuminated disk (radius planet_rad) that the
-        # rastered source actually covers, the same region regular imaging keeps via its
-        # n_pix=None auto-sizing. The fixed NFOV/WFOV values clipped the flat to a disk
-        # (radius n_pix//2 = 22 for NFOV) far smaller than what the star illuminates,
-        # throwing away most of the good flat data.
+        # size the flat to the full illuminated disk (radius planet_rad); the flat
+        # keeps radius n_pix//2, so use the disk diameter
         n_pix = 2 * planet_rad
     
     # wollaston prism angle, used below to place the combined rasters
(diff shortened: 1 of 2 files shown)


juliamilton:
These fixes seem to produce a better-looking flat field product, but let me know if we should further address the bright pixels:

image

corgidrp/recipe_templates/l2a_to_polflat.json
```diff
@@ -47,6 +47,12 @@
                 "TrapCalibration" : "AUTOMATIC,OPTIONAL"
             }
         },
+        {
+            "name" : "combine_subexposures",
+            "keywords" : {
+                "num_frames_per_group" : 6
```
semaphoreP: just want to confirm, this needs to be hard coded because there's no header keyword to check on.

semaphoreP:
New version (bottom left) is looking better than develop version (top left) and previous version of this PR (top right).

image

juliamilton:
Below is the star raster case. The left image shows the coadded mean of all the dithers, and the right shows the flat field from the star raster case. @sfr348 does the dither pattern look like you'd expect? The flat field seems to kind of match what is going on from the dither pattern, imo. 
image

juliamilton:
Now, this is the mocked planet case below. The left is the coadded mean of all the dithers, and the right is the flat field. There is not as much visible overlap between the dithers so the flat looks a lot better, but again I don't think we should place too much stock in this until we've been able to use the 2D convolution sims. 
image

corgidrp/flat.py
```diff
@@ -633,37 +729,33 @@ def create_onsky_pol_flatfield(dataset, planet=None,band=None,up_radius=55,im_si
...
-     
+    combined_rasters_pol1=combine_flatfield_rasters(resi_images_pol1_dataset,planet=planet,band=band,cent=cent_pol1, im_size=im_size, rad_mask=rad_mask,planet_rad=planet_rad,n_pix=n_pix, n_pad=n_pad,image_center_x=image_center_x,image_center_y=image_center_y)
+    combined_rasters_pol2=combine_flatfield_rasters(resi_images_pol2_dataset,planet=planet,band=band,cent=cent_pol2, im_size=im_size, rad_mask=rad_mask,planet_rad=planet_rad,n_pix=n_pix, n_pad=n_pad,image_center_x=image_center_x,image_center_y=image_center_y)
     
+    # each component is auto-sized (n_pix=None) to its own largest hole-free disk,
+    # so crop/place each one with its own radius rather than a shared n_pix
+    cent_n_pol1=combined_rasters_pol1[2]; n_pix_pol1=combined_rasters_pol1[3]
+    cent_n_pol2=combined_rasters_pol2[2]; n_pix_pol2=combined_rasters_pol2[3]
+    radius_left=n_pix_pol1//2
+    radius_right=n_pix_pol2//2
```
semaphoreP: do you understand why radius = num_pixels / 2? It seems like it's saying radius = Area / 2. Do you understand this?

semaphoreP (commented):
Looks good. I just had one small question about how the algorithm works.

corgidrp/flat.py
juliamilton: I think n_pix_pol is the diameter, following the convention of n_pix in imaging mode

juliamilton: I don't think we need separate radii for the left and right side spots though, they shouldn't differ. So I simplified it.

semaphoreP (approved):

Issue #1013, opened 2026-08-20
SPEC: add possibility to determine the wavelength zero point with model spectra, see CAR160


JuergenSchreiber:
**Is your feature request related to a problem? Please describe.**
in CAR160 tests to dermine the absolute spectral flux in band 3, no narrowband (3D) observations are/can be done (only 3F), so the wavelength zero point has to be determined by another way.

**Describe the solution you'd like**
for absolute flux calculation we use calspec model spectra, so a cross correlation should be possible, more details: @neilzim 

**Additional context**
see corgi-car-tests issue #99

JuergenSchreiber:
L1 simulations of CAR160 are on sharepoint, but I added also some narrowband (3D) simulations, so you should filter them out first.

semaphoreP:
Fixed in PR #1030

PR #1030, opened 2026-09-04
Issue 1013

neilzim:
## Describe your changes

Some spec mode calibration data sets contain prism images of a star collected in the broadband filter only, without associated narrowband frames. The l3_to_l4 determine_wave_zeropoint() function has been expanded to enable wavelength calibration using stellar templates to register the image and estimate the zeropoint. The default zeropoint determination strategy remains the narrowband image (either via DM satellite spot or narrowband calibration star image).

The registration algorithm uses the existing fit_psf_centroid() step function, which has been revised in response to testing for more robust handling of array bounds.

This feature requires a library of 64 template prism images tracked with Git LFS: 8 spectral types x 2 FPAM configs (OPEN_34, ND225) x 2 FSAM configs (open, slit) x 2 prisms

The spec mode flux calibration and ND calibration E2E tests (l1_to_spec_flux_e2e.py, l1_to_nd_filter_spec_e2e.py) have been expanded to include functions to test broadband-only calibration data.

## Type of change

Issue #1013  


corgidrp/data/spectroscopy/standard_star_sptypes.csv
semaphoreP, line 1: Can we do the same thing here as what we do for the stellar_polarization_database.csv: allow users to override these default values in the `~/.corgidrp` directory. So, when this file is read in (2 times in spec.py), we first check if a file exists in `corgidrp.config_folder/standard_star_sptypes.csv` and fall back to this. This allows us to make modifications or add to it without needing to change the repo.

corgidrp/l3_to_l4.py
```diff
@@ -1050,11 +1051,133 @@ def northup(input_dataset,use_wcs=True,rot_center='im_center',new_center=None):
...
 
 
-def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset = None, subtract_no_offset_frames=True, additional_frame_sep_prikeys = None, additional_frame_sep_extkeys = None, xcent_guess = None, ycent_guess = None, bb_nb_dx = None, bb_nb_dy = None, return_all = False):
-    """ 
+# Half-height of the stamp used to register a broadband prism image against a model template.
+# Also used by tests/test_spec.py::test_template_headers_complete
+BROADBAND_PRISM_HALFHEIGHT = 38
+
+# Half-height of the stamp used to register a narrowband prism image against a model template.
+NARROWBAND_PRISM_HALFHEIGHT = 26
```
semaphoreP: should these be optional keyword values in determine_wave_zeropoint, so do you think we'll never need to change them? If we want to keep them as "global settings", I'd add them to the `~/.corgidrp/corgidrp.cfg` file that's created in `__init__.py`. Although if it's only used for one function, I'd rather expose them as optional keywords.

corgidrp/data/spectroscopy/standard_star_sptypes.csv
semaphoreP: and we should have the same behavior for the folder of spectroscopy/templates/*.fits files. If a file exist in .corgidrp/spectroscopy/templates, maybe we use that one and fall back to these. This allows us to update these templates on the fly (which seems like we might need to do).

corgidrp/l3_to_l4.py
semaphoreP: also `NARROWBAND_PRISM_HALFHEIGHT ` seems to be bigger than before (previously I think this defaulted to 10). Is this to better handle the SPEC-ROT?

corgidrp/l3_to_l4.py
```diff
@@ -1165,54 +1315,70 @@ def determine_wave_zeropoint(input_dataset, spec_filter_offset, template_dataset
...
             science_frames.append(frame)
 
         all_science_frames += science_frames
 
+    # Apply model template registration to science frames whose group has no narrowband frames.
+    # The narrowband_present flag is dataset-wide, and is True here because some other group
+    # does have narrowband frames, so the wholesale model template registration path above was not taken.
+    # When narrowband_present is False, the for loop above already applied the broadband template fallback, 
+    # so this loop would find nothing to do.
+    if with_science and narrowband_present and allow_template_fallback:
```
semaphoreP, line 1362: no need to do anything, but I'm just commenting that I think there probably is a way to simplify this logic so there's only one for loop to go through the different subsets (i.e., by merging this with the previous for loop). But this code is fairly straghtforward, so this is not a big concern.

corgidrp/spec.py
```diff
@@ -305,56 +361,233 @@ def _norm_stamp(a):
...
+    filenames = [f for f in candidates if f.endswith(ND_TEMPLATE_SUFFIX) == nd_wanted]
+    if nd_wanted and len(filenames) == 0:
+        # The ND filter transmission is chromatic, so a template without it is an approximation.
+        warnings.warn("no ND model template matching {0} in {1}; falling back to a template without "
+                      "an ND filter".format(pattern, template_dir))
+        filenames = [f for f in candidates if not f.endswith(ND_TEMPLATE_SUFFIX)]
+    if len(filenames) == 0:
+        raise AttributeError("no model template files matching {0} in {1}".format(pattern, template_dir))
+    available = [read_template_sptype_token(f) for f in filenames]
+    matched = match_template_spectral_type(host_sptype, available)
```
semaphoreP: `get_model_template_filename` only gets used by this function and is only a few lines of code. should we just merge these functions together?

corgidrp/spec.py
```diff
@@ -305,56 +361,233 @@ def _norm_stamp(a):
...
+    Args:
+        sptype (str): spectral type of the observed star
+        available_sptypes (list of str): spectral types for which a template exists
+
+    Returns:
+        str: the closest entry of available_sptypes
+    """
+
+    # max spectral type mismatch tolerated between a star and its model template
+    max_sptype_index_mismatch = 5
```
semaphoreP: should this be a free-parameter or you think this is already generous enough?

corgidrp/spec.py
```diff
@@ -305,56 +361,233 @@ def _norm_stamp(a):
...
+        warnings.warn("no ND model template matching {0} in {1}; falling back to a template without "
+                      "an ND filter".format(pattern, template_dir))
+        filenames = [f for f in candidates if not f.endswith(ND_TEMPLATE_SUFFIX)]
+    if len(filenames) == 0:
+        raise AttributeError("no model template files matching {0} in {1}".format(pattern, template_dir))
+    available = [read_template_sptype_token(f) for f in filenames]
+    matched = match_template_spectral_type(host_sptype, available)
+    return filenames[available.index(matched)]
+
+def read_template_sptype_token(filename):
```
semaphoreP: also wondering if this should just be merged into `get_model_template_filename` given how short it is and the fact it is used once.

corgidrp/spec.py
```diff
@@ -14,6 +14,18 @@
...
+# kind ("offset" / "filtersweep" / "model"), the prism, and the CFAM filter.
+SPAM_TOKEN = {"SPEC": "spec-nom", "SPECROT": "spec-rot"}
+SPAM_FOR_PRISM = {"PRISM3": "SPEC", "PRISM2": "SPECROT"}
+SLIT_TOKEN = {"R1C2": "r1c2slit", "R2C2": "r2c2slit", "OPEN": "noslit"}
+NARROWBAND_FILTER = {"3": "3D", "2": "2C"}
+ND_TEMPLATE_SUFFIX = "_nd225.fits"
+
+# Half-heights of the fitting stamps
+FILTERSWEEP_HALFHEIGHT = 24
+BROADBAND_HALFHEIGHT = 30
```
semaphoreP, line 27: maybe these also should be considered as global settings in the corgidrp config file or optional keywords, rather than hard coded here.

tests/e2e_tests/l1_to_nd_filter_spec_e2e.py
semaphoreP, line 1: could we add a new check on the value of the wavelength solution? given I think the simulated OD vs wavelength is just flat, I think it'd be good to also test we got a decent wavelength solution.

semaphoreP (commented):
Thanks for this big update. I have some minor comments and some suggestions on how to make changing the auxiliary data more flexible (the template files, the star spectral type dictionary, and the psuedo-global config settings). Hopefully not too complicated.

tests/e2e_tests/l1_to_nd_filter_spec_e2e.py
neilzim: The wavelength zeropoint for each dither position is now tested at https://github.com/roman-corgi/corgidrp/blob/c41677fc93af41769bc46199c26a6de2477042c5/tests/e2e_tests/l1_to_nd_filter_spec_e2e.py#L597

corgidrp/l3_to_l4.py
neilzim: I added input parameter options for stamp halfheights. For NARROWBAND_PRISM_HALFHEIGHT, the original value, 10, is better.

corgidrp/data/spectroscopy/standard_star_sptypes.csv
neilzim: - In spec.py, [get_star_spectral_type()](https://github.com/roman-corgi/corgidrp/blob/c41677fc93af41769bc46199c26a6de2477042c5/corgidrp/spec.py#L506) now prioritizes a spectral type CSV file defined in ~/.corgidrp/spectroscopy/

- Similarly, a new function [find_template_files()](https://github.com/roman-corgi/corgidrp/blob/c41677fc93af41769bc46199c26a6de2477042c5/corgidrp/spec.py#L29) looks for possible template file overrides in ~/.corgidrp/spectroscopy/

corgidrp/spec.py
neilzim: Eliminated read_template_sptype_token()

corgidrp/spec.py
neilzim: Added an option to adjust this parameter as an input to [get_model_template_filename()](https://github.com/roman-corgi/corgidrp/blob/c41677fc93af41769bc46199c26a6de2477042c5/corgidrp/spec.py#L445)

corgidrp/spec.py
neilzim: Not familiar with this. Do you have an example of where you want these defined?

JuergenSchreiber: You would need to add them to the "__init__.py" functions, but I would prefer optional keywords if possible

semaphoreP: the current change of moving them to optional keywords is good. No more changes needed.

tests/test_spec.py
```diff
@@ -1770,11 +1723,444 @@ def test_convert_spec_to_flux_factor():
...
+
+    # a user copy of the same name replaces the bundled one
+    shutil.copy(bundled[0], user_templates / os.path.basename(bundled[0]))
+    override = steps.find_template_files("spec-nom_unocc_noslit_model_prism3_3f_g0v.fits")
+    assert len(override) == 1
+    assert override[0] == str(user_templates / os.path.basename(bundled[0]))
+
+    # a name found only in the user directory is added to the bundled ones
+    new_type = user_templates / "spec-nom_unocc_noslit_model_prism3_3f_k5v.fits"
+    shutil.copy(bundled[0], new_type)
```
semaphoreP: does this tmp file get deleted after the test runs?

corgidrp/spec.py
```diff
@@ -14,6 +14,39 @@
...
+
+    Args:
+        pattern (str): glob pattern matched against template file names
+
+    Returns:
+        list of str: matching paths, ordered by file name
+    """
+    paths = {}
+    for directory in (os.path.join(os.path.dirname(__file__), "data", "spectroscopy", "templates"),
+                      os.path.join(os.path.dirname(corgidrp.config_filepath), "spectroscopy", "templates")):
```
semaphoreP, line 47: is the fact that the user directory is 2nd in this tuple the sole reason that the user directory takes preference over the default data directory? If so, can we add a comment here about this?

semaphoreP (commented):
This looks good. I just have one quick question before signing off.

corgidrp/spec.py
neilzim: Yes- added a comment.

semaphoreP (approved):

Issue #1042, opened 2026-09-17
l1 to fluxcal pol e2e test is not reproducible and could be checking the values better


semaphoreP:
**Describe the bug**
Depending on machine, the l1 to fluxcal pol e2e test can fail due to line 60

```
new_l2b_filenames = [os.path.join(l2b_outputdir, f) for f in os.listdir(l2b_outputdir) if f.endswith('l2b.fits')] 
```

The `os.listdir()` call changes behavior depending on machine and gives different file orders. The code after this line of code uses the first file to grab the DPAMNAME but DPAMNAME could be POL0 and POL45, and the test only passes if the DPAMNAME = POL45. 

This raises a bigger issue of why the test takes the median of both POL0 and POL45 data to check for correctness. We should probably not be averaging the two sets of frames when checking the abs flux calibration. 

**To Reproduce**
The DAE machine produces a bad file order, causing the test to fail. Alternatively, force the DPAMNAME = POL0, and the test will fail.

**Expected behavior**
Test should pass and not be dependent on file order.

**corgidrp version:**
 - 5.1

semaphoreP:
Fixed in PR #1043

PR #1043, opened 2026-09-18
fixed bug in pol fluxcal and pol fluxcal e2e test

everlastingEric:
## Describe your changes
- Fixed a bug in the pol fluxcal e2e test where the frames with the DPAM at POL0 and POL45 are combined together, they are now evaluated separately. This should also fix the file ordering dependency issue that determines whether the test fails or passes.
- Fixed a similar bug in the `calibrate_pol_fluxcal_aper` step function where frames are combined together without checking the DPAM position first. 

## Type of change

#1042 


tests/e2e_tests/l1_to_fluxcal_pol_e2e.py
```diff
@@ -75,22 +74,28 @@ def test_l1_to_fluxcal_pol_e2e(e2edata_path, e2eoutput_path):
...
+        aper_pos = [(o_x, o_y), (e_x, e_y)]
+        apertures = CircularAperture(aper_pos, r=5)
+        phot = aperture_photometry(frame.data, apertures, method='center')
+        # combine counts from o and e beam, divide by exposure time
+        counts_arr.append((phot['aperture_sum'][0] + phot['aperture_sum'][1]) / frame.ext_hdr["EXPTIME"])
+    
+    # take the mean normalized counts from all images, cross check with actual flux
+    counts_med = np.nanmean(counts_arr)
+    flux_count = flux_fac.fluxcal_fac * counts_med
+    assert flux == pytest.approx(flux_count, rel = 0.075)
```
semaphoreP: did you have to raise the threshold to get the new test to pass?

semaphoreP (commented):
thanks for this fix. just one question about the test itself

tests/e2e_tests/l1_to_fluxcal_pol_e2e.py
everlastingEric: yeah the difference was just a bit over 5% so I had to raise the thresold a little

semaphoreP: did you look into why? is there some bad frames? Because we are basicaslly doing the same operation in the e2e test as the step function right?

everlastingEric: I know there is one bad frame where the photometry just returned NaN. I think (for reference the flux returned by the test is slightly lower than that returned by the actual calibration) the main difference is that the test does photometry on each individual frame and then takes the median while the step function median combines frames first and does photometry on that master frame.  A possibility is that doing photometry on individual frames means more bad pixels are included that could artifically lower the count compared to the median-combined frame. I'm happy to change the test to also median-combine first to see if that helps with the accuracy.

semaphoreP: I think worth an investigation given we are using the same data.. it might just be that the star is faint and we don't have many frames of data, so the measurement is noisy, but it'd be good to check.

everlastingEric: ah ok I think I found the reason why the test might have returned slightly lower flux. By default the flux calibration uses the subpixel method for aperture photometry while for the test I had it set to center so it counted pixels as only entirely in or out, switching the test to also use the subpixel method seemed to account for the missing flux.

semaphoreP: Nice. Is it ready for re-review?

everlastingEric: yep!

semaphoreP (approved):

